PR #32, opened 2016-07-22
Fixing modules after integrating experiment-level 

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-31
https://biosistemika.atlassian.net/browse/SCI-34


app/helpers/permission_helper.rb
```diff
@@ -122,7 +122,7 @@ module PermissionHelper
     ] do |proxy, *args, &block|
       if args[0]
         my_module = args[0]
-        if my_module.active? and my_module.project.active?
+        if my_module.active? and my_module.experiment.project.active?
```
Ducz0r: Since we will allow experiments to be archived, I believe we should also check for `my_module.experiment.active?` here.

app/helpers/permission_helper.rb
```diff
@@ -562,7 +562,7 @@ def can_view_protocol(protocol)
       protocol.added_by == current_user
     elsif protocol.in_module?
       my_module = protocol.my_module
-      my_module.active? and my_module.project.active? and can_view_module(my_module)
+      my_module.active? and my_module.experiment.project.active? and can_view_module(my_module)
```
Ducz0r: Add `my_module.experiment.active?`.

app/helpers/permission_helper.rb
```diff
@@ -619,7 +619,7 @@ def can_unlink_protocol(protocol)
   def can_revert_protocol(protocol)
     if protocol.linked?
       my_module = protocol.my_module
-      my_module.active? and my_module.project.active? and is_user_or_higher_of_project(my_module.project)
+      my_module.active? and my_module.experiment.project.active? and is_user_or_higher_of_project(my_module.experiment.project)
```
Ducz0r: Add `my_module.experiment.active?`.

app/helpers/permission_helper.rb
```diff
@@ -628,7 +628,7 @@ def can_revert_protocol(protocol)
   def can_update_protocol_from_parent(protocol)
     if protocol.linked?
       my_module = protocol.my_module
-      my_module.active? and my_module.project.active? and is_user_or_higher_of_project(my_module.project)
+      my_module.active? and my_module.experiment.project.active? and is_user_or_higher_of_project(my_module.experiment.project)
```
Ducz0r: .. and onwards (`my_module.experiment.active?`), you get the idea :smile:.

app/models/project.rb
```diff
@@ -127,431 +104,22 @@ def user_role(user)
...
-      .distinct
-  end
-
-  def assigned_samples
-    Sample.joins(:my_modules).where(my_modules: {id: my_modules} )
-  end
-
-  def unassigned_samples(assigned_samples)
-    Sample.where(organization_id: organization).where.not(id: assigned_samples)
+  def project_my_modules
```
Ducz0r: Urm. Project has many experiments. Each experiment has many modules. So wouldn't the correct relation be more like:
```ruby
MyModule.where('experiment_id IN (?)', experiments.select(:id))
```

jbargu:
Wouldn't be much easier and readable though to just create a `project` method in `my_module.rb`, something along the lines:

```
def project
  experiment.project
end
```

Ducz0r:
Or that :+1: .

ZmagoD:
@jbargu @Ducz0r I will leave the project_my_modules method in project model for now because the space_taken method relies on that maybe we should make the project_my_modules method private later

PR #33, opened 2016-07-26
Added experiments_controller

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-37

For now I added experiments controller with update/create/archive actions and a new/update form.


Ducz0r:
I wouldn't merge this for now. I don't like adding code that will probably never be used. The thing is, experiments will have various ways to be created/edited/archived/..., but most certainly not via individual pages with a plain' old form. So I'd use this MR to actually start working on some GUI part of the application.

ZmagoD:
Hmm.. it will be good to have the experiments controller and declare routes to continue with the experiments level integration

Ducz0r:
Well... yeah, but just the bare bones controller, not the e.g. view action, and the corresponding HTML. If you remove that, I agree that we can merge it.

ZmagoD:
OK

PR #34, opened 2016-07-26
Update global search for experiment level

jbargu:
https://biosistemika.atlassian.net/browse/SCI-33

### DIFF
--- a/app/models/experiment.rb
+++ b/app/models/experiment.rb
@@ -1,5 +1,5 @@
 class Experiment < ActiveRecord::Base
-  include ArchivableModel
+  include ArchivableModel, SearchableModel
 
   belongs_to :project, inverse_of: :experiments
   belongs_to :created_by, foreign_key: :created_by_id, class_name: 'User'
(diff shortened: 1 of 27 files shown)


app/models/experiment.rb
```diff
@@ -23,6 +23,47 @@ class Experiment < ActiveRecord::Base
...
+    if include_archived
+      new_query =
+        Experiment
+          .where(project: project_ids)
+          .where_attributes_like([:name, :description], a_query)
+    else
+      new_query =
+        Experiment
+          .is_archived(false)
+          .where(project: project_ids)
```
Ducz0r: Nice, I didn't know you can write the `.id IN (?)` condition this elegantly!

app/models/my_module_group.rb
```diff
@@ -30,8 +31,8 @@ def self.search(user, include_archived, query = nil, page = 1)
     new_query = MyModuleGroup
       .distinct
       .joins(:experiment)
```
Ducz0r: I don't think this join is needed here any more?

Ducz0r:
I've reviewed the MR, and it looks good (with the exception of 1 remark I made in the comments). Jolly good work! :+1: 
I sense somebody's been busy using Rubocop lately.

app/models/my_module_group.rb
jbargu: Good catch!

jbargu:
Is it that obvious? :grin: 
Nah, I actually haven't set it up yet. The above lines just looked hideous in terminal, so I decided to fix  them along. You could say the above lines got TERMINAL treatment. :sunglasses:

PR #42, opened 2016-08-01
Deleting items for report

ZmagoD:
When step file,table or checklist are deleted then get removed from the report
https://biosistemika.atlassian.net/browse/SCI-247


jbargu:
This appears to be working. Good job! :+1: Merge away!

PR #47, opened 2016-08-03
Update reports with experiment level

jbargu:
https://biosistemika.atlassian.net/browse/SCI-35

Currently still missing:
- printing doesn't include .css for some reason
- sorting experiment after adding doesn't work, but does work at the start
- migration for updating old reports with experiment level


jbargu:
Heh, this hound is barking at the wrong tree!

ZmagoD:
The code work like expected but I don't know why hound is yelling at the schema file :scream_cat: 
Good job :+1:

ZmagoD:
The only thing I see which should be correct is in the sidebar the "Report for project name" label

PR #48, opened 2016-08-04
[WIP] fixes problems with adding new sample type or new sample group [SCI-286]

mlorb:
https://biosistemika.atlassian.net/browse/SCI-286?filter=-1


ZmagoD:
It looks to do what it is supposed to but the problem here is that before that issue occurred we fixed an other issue that we wanted to load the created types/groups asynchronously so the reload option don't fit in here we have to load this notices using ajax call.

PR #49, opened 2016-08-04
fixes keywords overlapping when editing keywords of protocol [SCI-293]

mlorb:
https://biosistemika.atlassian.net/browse/SCI-293?filter=-1


ZmagoD:
Look like it's working but is better to put this css class someware in the app/assets/... /protocols and target the .bootstrap-tagsinput > .label . This will take effect only on the required label and also if the bootstrap-tagsinput library gets update this all the changes you made there will be lost

ZmagoD:
:+1: merge away!

PR #53, opened 2016-08-09
Project archive page, with experiment

jbargu:
https://biosistemika.atlassian.net/browse/SCI-42

When experiment is archived, it should be read-only. User shouldn't be able to access module screen or edit anything experiment related. Samples/reports page stay as they are (they can be editable), since they're bound to team/project respectively.


ZmagoD:
It looks OK :+1:  Merge away

PR #56, opened 2016-08-11
fixes problem with split table when generating pdf report [SCI-302]	

mlorb:
https://biosistemika.atlassian.net/browse/SCI-302?filter=-1


ZmagoD:
Merge away :+1:

PR #58, opened 2016-08-12
Filter search query for special chars [fixes SCI_74]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-74


jbargu:
I've tested with some numbers and it looks good. Is there anything special I should do to reproduce it?
If not, merge away! :+1:

PR #61, opened 2016-08-16
fixes dissapearance of sample type/group when editing samlpe or adding new sample [SCI-291]

mlorb:
https://biosistemika.atlassian.net/browse/SCI-291?filter=-1


ZmagoD:
It works like expected :+1: Merge away

PR #62, opened 2016-08-16
Zd sci 318

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-318

https://biosistemika.atlassian.net/browse/SCI-317


app/assets/javascripts/protocols/import_export/import.js.erb
```diff
@@ -495,6 +514,11 @@ function importProtocolFromFile(
       },
       error: function(ev) {
         resultCallback(ev.responseJSON);
+
+        var path = new RegExp("modules");
```
jbargu: Since it's `erb` file, I guess you could use rails here, but to each its own. :)

jbargu:
Everything works as expected. :+1: :shipit:

PR #68, opened 2016-08-17
Zd sci 330

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-330


jbargu:
Work as expected. Just press that merge button, don't be shy.

PR #69, opened 2016-08-17
cleans experiment form report if moved [fixes SCI-335]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-335


jbargu:
Let it go, works like a charm! :+1:

PR #72, opened 2016-08-18
fixed experiments archive bug

ZmagoD:
Restores experiment and tasks

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -149,12 +149,15 @@ class Project < ActiveRecord::Base
     if role.blank?
       MyModule.none
     elsif role == 'owner'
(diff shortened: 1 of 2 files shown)


app/models/project.rb
```diff
@@ -149,12 +149,15 @@ def assigned_modules(user)
     if role.blank?
       MyModule.none
     elsif role == 'owner'
-      project_my_modules.where(archived: false)
+      project_my_modules
+        .joins(:experiment)
```
jbargu: You should include `where(archived: false)` for modules as well.

jbargu:
Archived modules are still shown in due dates.

jbargu:
Much better :smile: :+1:

PR #73, opened 2016-08-18
fixes user duplication [fixes SCI-344]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-344


jbargu:
Ez pz, nice one. :ship: :it:

PR #76, opened 2016-08-19
fixes notifications when creating new sample type or group [SCI-286]

mlorb:


app/controllers/sample_types_controller.rb
```diff
@@ -17,14 +17,19 @@ def create
 
     respond_to do |format|
       if @sample_type.save
-        flash[:success] = t(
-          "sample_types.create.success_flash",
-          sample_type: @sample_type.name,
-          organization: @organization.name
-        )
+        # flash[:success] = t(
```
Ducz0r: You can completely remove this commented block of code. This method call just works on regular page reloads, and since we don't support plain HTML rendering here (no `format.html`).

Ducz0r:
After you fix the one comment I gave, this MR seems ready to be merged to me!

PR #80, opened 2016-08-22
delete and edit comments for projects, protocols, results, and workfl…

scibb:
New feature allows editing & deleting comments from projects, protocols, results, and workflow elements.

### DIFF
--- a/app/assets/javascripts/my_modules.js
+++ b/app/assets/javascripts/my_modules.js
@@ -1,4 +1,3 @@
-
 // Bind ajax for editing descriptions
 function bindEditDescriptionAjax() {
   var editDescriptionModal = $("#manage-module-description-modal");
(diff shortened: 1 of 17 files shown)


Ducz0r:
I have couple of remarks (will probably implement them myself, just for reference) - this PR doesn't cover the **Feature 01** of specifications:
* there is no permission checking (anybody can edit comments from anybody),
* wording change (*Published on* -> *Created on*) was not implemented,
* GUI is very very non-user-friendly and shouldn't go into production in such a state,
* wouldn't it be easier to just have 2 routes for all comments (e.g. `DELETE /comments/{id}` and `PUT /comments/{id}`), and also somehow avoid all the code duplication in JavaScript?

Ducz0r:
I've rebased this branch onto latest master, so I'm closing this PR and opening a new one.

PR #83, opened 2016-08-22
Fixes search result link for experiment [fixes SCI-347]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-347


jbargu:
Be-auty-ful! :white_flower: :+1:

PR #84, opened 2016-08-22
Moving workflows/task from canvas

jbargu:
https://biosistemika.atlassian.net/browse/SCI-328

Addtionality fixes https://biosistemika.atlassian.net/browse/SCI-358


ZmagoD:
If I move the workflow to another experiment everything is OK but if I move a single task it get a "random position" on the canvas.

jbargu:
Fixed it, some other method was overriding new position.

PR #87, opened 2016-08-23
hide the dropdown if user has no permissions(technician) [fixes SCI-371]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-371

### DIFF
--- a/app/views/projects/show/_experiment.html.erb
+++ b/app/views/projects/show/_experiment.html.erb
@@ -2,7 +2,7 @@
   <div class="panel panel-default experiment-panel"
        data-id="<%= experiment.id %>">
     <div class="panel-heading">
-      <% if is_user_or_higher_of_project @project %>
+      <% if can_view_experiment_actions experiment %>
         <div class="dropdown pull-right">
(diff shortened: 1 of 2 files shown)


app/views/projects/show/_experiment.html.erb
```diff
@@ -2,21 +2,21 @@
...
-                aria-haspopup="true"
-                aria-expanded="true">
-          <span class="visible-xs-inline"><i class="glyphicon glyphicon-sort"></i></span>
-          <span class="caret"></span>
-        </button>
-        <%= render partial: 'experiments/dropdown_actions.html.erb',
-                            locals: { project: @project, experiment: experiment } %>
-      </div>
-
+      <% if is_user_or_higher_of_project @project %>
```
jbargu: I would change this to actual permissions method, since this is very error prone. Imagine someone someday fixing permissions and would be scratching his head why it doesn't respond.

jbargu:
If you fix the above with the actual permissions it works as it should! :ship: :it:

jbargu:
Way better! Nice one :+1:

PR #95, opened 2016-08-24
adds edit workflow link if no workflow [fixes SCI-374]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-374

### DIFF
--- a/app/assets/javascripts/projects/canvas.js
+++ b/app/assets/javascripts/projects/canvas.js
@@ -2990,6 +2990,15 @@ function initJsPlumb(containerSel, containerChildSel, modulesSel, params) {
   jsPlumb.fire("jsPlumbLoaded", instance);
 }
 
+// Opens edit mode if redirected from empty experiment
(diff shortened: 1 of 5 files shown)


Ducz0r:
I've reviewed this PR. Looks good, I only have 2 remarks (you can probably ignore the last one):
* This PR introduces a bug that is (at least on my machine) not present on `master`. If you press "Add a task to experiment...`, you are presented with Edit canvas screen. But **cancel** button doesn't work (it keeps you in the edit mode). So that should be fixed.
* Throughout most of application, we have a different way of saying text is blank - e.g. "No comments!", "No description", ... So maybe that would be a preferred way to handle this?

Ducz0r:
After you fix the mentioned bug, this seems ready for merge to me :+1:.

ZmagoD:
@Ducz0r I fixed the canvas related bug can you check if it works on your machine?

Ducz0r:
It works now! Looks good for merge to me! :+1:

PR #103, opened 2016-08-25
Edit&delete comments

Ducz0r:
New feature allows editing & deleting comments from projects, modules/tasks, protocol steps and results.


jbargu:
When you remove the last comment, it should say 'No comment' like when no comments.

jbargu:
It would be cool if you could see, if some edited comment and when.

jbargu:
When editing comments on canvas, right after insert a comment, the dropdown will be misplaced. The more comments, the more it'll be misplaced. However if you open the comment window, dropdown is correctly placed.

jbargu:
When editing protocol and then trying to either edit/delete a comment, nothing will happen.

jbargu:
Jolly good, besides the above remarks, works as a charm. Here have a :cake:

Ducz0r:
Nice finds! I fixed everything you found, with the exception of suggestion for last edit timestamp & user. It's not in the specifications. I believe we can implement that one day, but it's not today :smile_cat:.

PR #108, opened 2016-08-29
Fixes step destroing bug and fixes reports [fixes SCI-247]

ZmagoD:


jbargu:
This ....
appears to be working! Good job! :+1:

PR #109, opened 2016-08-29
fixes notification when adding new sample [SCI-286]

mlorb:
https://biosistemika.atlassian.net/browse/SCI-286?filter=-1

### DIFF
--- a/app/assets/javascripts/samples/sample_datatable.js
+++ b/app/assets/javascripts/samples/sample_datatable.js
@@ -398,6 +398,7 @@ function onClickSave() {
         dataType: "json",
         data: data,
         success: function (data) {
+            sampleAlertMsg(data.flash);
             onClickCancel();
         },
         error: function (e, eData, status, xhr) {
--- a/app/assets/javascripts/samples/sample_datatable.js
+++ b/app/assets/javascripts/samples/sample_datatable.js
@@ -336,7 +336,7 @@ function onClickEdit() {
         },
         error: function (e, data, status, xhr) {
             if (e.status == 403) {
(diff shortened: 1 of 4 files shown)


jbargu:
When I try to add sample now, I get the following error:

```ruby
> Render and/or redirect were called multiple times in this action. Please note that you may only call render OR redirect, and at most once per action. Also note that neither redirect nor render terminate execution of the action, so if you want to exit an action after redirecting, you need to do something like "redirect_to(...) and return".

app/controllers/samples_controller.rb, line 35
----------------------------------------------

``` ruby
   30       errors = {
   31         init_fields: [],
   32         custom_fields: []
   33       };
   34   
>  35       respond_to do |format|
   36         if can_create_samples(@organization)
   37           if params[:sample]
   38             # Sample name
   39             if params[:sample][:name]
   40               sample.name = params[:sample][:name]
```

App backtrace
-------------

 - app/controllers/samples_controller.rb:35:in `create'
 - app/controllers/application_controller.rb:85:in `set_time_zone'
```

File: app/assets/javascripts/samples/sample_datatable.js [hunk 1]
```diff
@@ -398,6 +398,7 @@ function onClickSave() {
         dataType: "json",
         data: data,
         success: function (data) {
+            sampleAlertMsg(data.flash);
             onClickCancel();
         },
         error: function (e, eData, status, xhr) {
```
jbargu: I'd also hide notification on sample error.

jbargu:
I'd just show that notification, otherwise works like a charm. Very nice, here have a :cookie: :+1:

jbargu:
Works like a charm. Merge away!! :+1:

PR #119, opened 2016-08-30
fixes missing experiment name in side bar [SCI-396]

mlorb:
https://biosistemika.atlassian.net/browse/SCI-396?filter=-1


jbargu:
A piece of art if I do say it myself. :+1:

PR #120, opened 2016-08-30
When moving a task, it should connect the source and destination

jbargu:
https://biosistemika.atlassian.net/browse/SCI-328


mlorb:
bulletproof :+1:

PR #123, opened 2016-08-31
Fix the default parameters in JS

Ducz0r:
This was causing major issues in IE.


jbargu:
It appears you swapped every occurrence, I tried with the following regex:
`grep -inr '^.*function.*=.*$' .`

Merge away :smile: :+1:

PR #127, opened 2016-09-01
Table from step stays in report, although it is deleted from step.

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-247


jbargu:
Looks good! :+1:

PR #129, opened 2016-09-02
adds tasks, experiment and project category in protocol search results [SCI-310]

mlorb:
https://biosistemika.atlassian.net/browse/SCI-310?filter=-1

### DIFF
--- a/app/helpers/search_helper.rb
+++ b/app/helpers/search_helper.rb
@@ -6,7 +6,7 @@ module SearchHelper
     end
     experiments.uniq
   end
-  
+
   def sub_results(el)
     elements = []
     el.each do |m|
--- a/app/helpers/search_helper.rb
+++ b/app/helpers/search_helper.rb
@@ -6,4 +6,11 @@ module SearchHelper
     end
     experiments.uniq
   end
+  def sub_results(el)
+    elements = []
+    el.each do |m|
+      elements << m
+    end
+    elements.uniq
+  end
 end
--- a/app/helpers/search_helper.rb
+++ b/app/helpers/search_helper.rb
@@ -6,6 +6,7 @@ module SearchHelper
     end
     experiments.uniq
   end
+  
   def sub_results(el)
     elements = []
     el.each do |m|
--- a/app/helpers/search_helper.rb
+++ b/app/helpers/search_helper.rb
@@ -6,12 +6,4 @@ module SearchHelper
     end
     experiments.uniq
   end
-
-  def sub_results(el)
-    elements = []
-    el.each do |m|
-      elements << m
-    end
-    elements.uniq
-  end
 end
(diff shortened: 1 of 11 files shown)


app/helpers/search_helper.rb
```diff
@@ -6,4 +6,12 @@ def experiments_results(tag)
...
     experiments.uniq
   end
+
+  def sub_results(el)
+    elements = []
+    el.each do |m|
+      elements << m
+    end
+    elements.uniq
+  end
```
jbargu: For this method you could just use oneliner:
```ruby
el.to_a.uniq
```
if I'm not mistaken.

app/views/search/results/_protocols.html.erb
```diff
@@ -48,6 +48,47 @@
...
+                       locals: { my_module: mod } %>
+            <% if i != (sub_results(protocol.linked_modules).count - 1) %>
+              ,&nbsp;
+            <% end %>
+          <% end %>
+        <% end %>
+      </span>
+      <br>
+      <span>
+        <% if protocol.linked_experiments %>
```
jbargu: I'd say you'd rather save `linked_modules` for the first time and then pass it to the `linked_experiments`. Now you do 2 queries for this. And 3 for projects below, whereas it could be 1 for every entity.

app/views/search/results/_protocols.html.erb
```diff
@@ -48,6 +48,47 @@
...
     </span>
     <br>
+    <% if protocol.nr_of_linked_children > 0 %>
+      <span>
+        <% if protocol.linked_modules %>
+          <%=t "search.index.modules" %>
+          <% protocol.linked_modules.each_with_index do |mod, i| %>
+            <%= render partial: "search/results/partials/my_module_text.html.erb",
+                       locals: { my_module: mod } %>
+            <% if i != (sub_results(protocol.linked_modules).count - 1) %>
```
jbargu: You could just `distinct` on the original query and thusly render `sub_results` method useless.

app/views/search/results/_protocols.html.erb
```diff
@@ -48,6 +48,47 @@
...
+                       locals: { experiment: exp } %>
+            <% if i != (sub_results(protocol.linked_experiments).count - 1) %>
+              ,&nbsp;
+            <% end %>
+          <% end %>
+        <% end %>
+      </span>
+      <br>
+      <span>
+        <% if protocol.linked_projects %>
```
jbargu: See above.

jbargu:
Looks good! :smile_cat:  Although I'd look into optimizing queries so we don't query database 6times+ for every protocol. 
Btw. I think you should be redirected to project page when you click on Project instead to projects index.

jbargu:
Prutty nice, I'm a fan! :+1:  Whenever, wherever this branch and master are meant to be together. :ship: :it:

PR #130, opened 2016-09-02
project overview - updating (or creating new) experiment details

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-348


jbargu:
Errors are not cleaned up after multiple failed attempts. Fix that and then just bag it and tag it. :+1:

mz3944:
Done some cahnges, should be working as expected now.

PR #132, opened 2016-09-05
Updates to invite users to team modal window

Ducz0r:
This PR does the following:
* Don't show the "everything belonging to the user will be reassigned onto you" alert when removing a Pending user (=user that hasn't confirmed his email yet) from the Team.
* Fix a bug when you first invited a new user, was notified that such a user exists (email), and then wanted to invite this user as an existing user (without closing the modal window).

https://biosistemika.atlassian.net/browse/SCI-228


jbargu:
Does everything that advertises! Good job! :+1: I don't what's the consensus on invitation mail being extended. If everything's OK, then this branch should just make sweet sweet love with master. :heart:

PR #134, opened 2016-09-06
Fixes files deletion when rails worker is running [fixes SCI-411] and…

mz3944:
… fixes picture's style files deletion.

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -141,6 +141,10 @@ class Asset < ActiveRecord::Base
     !(self.file.content_type =~ /^image/).nil?
   end
 
+  def text?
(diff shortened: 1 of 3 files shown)


Ducz0r:
I've tested this with S3, and it doesn't work yet (files still don't get deleted).
On local storage, the whole step editing is bugged so don't know about that :disappointed:.

mz3944:
I'm pretty sure they get deleted, I've checked the repositories on S3 right after deleting the files... Need to talk about this.

Ducz0r:
After re-testing again, this works now. I don't know what was the problem :astonished:. So merge away!

PR #136, opened 2016-09-06
Enabled to delete results in task archive [fixes SCI-370]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-370

### DIFF
--- a/app/controllers/results_controller.rb
+++ b/app/controllers/results_controller.rb
@@ -38,8 +38,6 @@ class ResultsController < ApplicationController
   end
 
   def can_destroy_result_permission
-    unless can_delete_module_result(@result)
-      render_403
-    end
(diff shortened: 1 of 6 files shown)


app/controllers/results_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+                        result: @result.name,
+                        module: @my_module.name)
+    @result.destroy
+    redirect_to archive_my_module_path(@my_module)
+  end
+
+  private
+
+  def load_vars
+    @result = Result.find_by_id(params[:id])
```
jbargu: What if `@result` is empty?

jbargu:
Beside that remark everything looks fabulous as always. :+1:

jbargu:
Looks good! :dancer:

PR #137, opened 2016-09-06
Removed header checking and added office file types grouping exceptio…

mz3944:
…ns when checking file spoofing [fixes SCI-418]


Ducz0r:
_I've removed Hound comments so I could have a clear picture of what the diff looks like._

Anyway, I've reviewed the MR. Those are my comments:
* Uploading the mentioned files now works (`.csv`, `.xml`, `.xltm` extensions), so that's good.
* If you upload a `.csv` or `.xml`, however, the text is not extracted from the file (so such files are not searchable). I'd at least expect from the `.csv` files to be searchable, do you remember if this used to work, or did we never bother to support this?

Not specifically related to this MR, but I would fix it along - currently (both on production, and if you run it locally), one of the tutorial files doesn't get generated because of incorrect MIME types error (this happens in delayed jobs, if you run `rake jobs:work`). This is the faulty file - [ddCq-quantification_diagnostics-trmplate.xls](app/assets/tutorial_files/ddCq-quantification_diagnostics-trmplate.xls). Since this happens to each new user, we have a pile of uncompleted delayed jobs on production, so __fixing this is pretty urgent__. The fix should be easy - I would think that all we need to do is add the following 2 MIME types to `# Spreadsheet` and `# Word processor` sets:
```
application/x-ole-storage
application/vnd.ms-office
```
But you should know better how to do that. Below is the error from the `Delayed::Job`. Also check if the text gets properly extracted.
```
INFO -- : [paperclip] Content Type Spoof: Filename ddCq-quantification_diagnostics-trmplate.xls (application/x-ole-storage from Headers, application/vnd.ms-excel from Extension), content type discovered: application/vnd.ms-office. See documentation to allow this combination.
```

And, finally - while you're at it already :smile: - Grammar Nazi in me is raging; can you rename the file `ddCq-quantification_diagnostics-trmplate.xls` to `ddCq-quantification_diagnostics-template.xls`?

Ducz0r:
Not everything I mentioned in my last comment was fixed, however, I'm merging this so it can go into today's build.

PR #138, opened 2016-09-06
Fixed result file deletion and result name updating [fixes SCI-409]. …

mz3944:
…Fixed step name and checklists names and its items names checking, and items removing [fixes SCI-407].


ZmagoD:
It works like it supposed to :+1:  nice one 
Merge away :ship:

PR #139, opened 2016-09-07
fixed moved experiment thumbnail [fixes SCI-410]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-410

### DIFF
--- a/app/models/experiment.rb
+++ b/app/models/experiment.rb
@@ -503,10 +503,7 @@ class Experiment < ActiveRecord::Base
     end
 
     # Generate workflow image for the experiment in which we moved the task
-    to_move.each_value do |id|
-      experiment = Experiment.find_by_id(id)
-      experiment.delay.generate_workflow_img
-    end
+    generate_workflow_img_for_moved_modules(to_move)
   end
 
   # Move module groups; this method accepts a map where keys
@@ -552,8 +549,19 @@ class Experiment < ActiveRecord::Base
     end
 
     # Generate workflow image for the experiment in which we moved the workflow
+    generate_workflow_img_for_moved_modules(to_move)
+  end
+
+  # Generates workflow img when the workflow or module is moved
+  # to other experiment
+  def generate_workflow_img_for_moved_modules(to_move)
+    experiment_ids = []
     to_move.each_value do |id|
+      experiment_ids << id
+    end
+    experiment_ids.uniq.each do |id|
       experiment = Experiment.find_by_id(id)
+      next unless experiment
       experiment.delay.generate_workflow_img
     end
   end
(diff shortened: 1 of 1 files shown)


jbargu:
It'd be great if image refreshed on moving workflow/module as well. Fix that and you good to go. :rocket:

app/models/experiment.rb
```diff
@@ -500,6 +500,7 @@ def move_modules(to_move)
       end
 
       my_module.save!
+      experiment.delay.generate_workflow_img
```
jbargu: I'd rather put this outside this loop. And call uniq on experiment_id first. Currently it could be possible if you moved 3 modules to the same experiment, 3 jobs will spawn for the same experiment whereas only one would be enough.

app/models/experiment.rb
```diff
@@ -542,6 +543,7 @@ def move_module_groups(to_move)
 
         group.experiment = experiment
         group.save!
+        experiment.delay.generate_workflow_img
```
jbargu: See above.

app/models/experiment.rb
```diff
@@ -499,7 +499,12 @@ def move_modules(to_move)
...
       end
 
-      my_module.save!
+      my_module.save
+    end
+
+    # Generate workflow image for the experiment in which we moved the task
+    to_move.each_value do |id|
+      experiment = Experiment.find_by_id(id)
       experiment.delay.generate_workflow_img
```
jbargu: Please no h8 m8, but it's possible that experiment id doesn't exist, so it needs check for `if present`

app/models/experiment.rb
```diff
@@ -543,9 +548,14 @@ def move_module_groups(to_move)
...
         group.experiment = experiment
         group.save!
-        experiment.delay.generate_workflow_img
       end
     end
+
+    # Generate workflow image for the experiment in which we moved the workflow
+    to_move.each_value do |id|
+      experiment = Experiment.find_by_id(id)
+      experiment.delay.generate_workflow_img
```
jbargu: See above.

app/models/experiment.rb
jbargu: `each_value` doesn't do uniq.

app/models/experiment.rb
```diff
@@ -552,8 +549,19 @@ def move_module_groups(to_move)
...
+    generate_workflow_img_for_moved_modules(to_move)
+  end
+
+  # Generates workflow img when the workflow or module is moved
+  # to other experiment
+  def generate_workflow_img_for_moved_modules(to_move)
+    experiment_ids = []
     to_move.each_value do |id|
+      experiment_ids << id
+    end
```
jbargu: Just as a remark, no hate. `to_move.values` would do the same.

PR #141, opened 2016-09-07
styled dottel line in sidebar [fixes SCI-340]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-340


jbargu:
So much wow, very sexy. :+1:

PR #142, opened 2016-09-08
fixes navbar size when resizing window [SCI-415]

mlorb:
https://biosistemika.atlassian.net/browse/SCI-415?filter=-1


ZmagoD:
it looks good! Merge away

PR #144, opened 2016-09-08
set experiment description to 255

ZmagoD:


mz3944:
Sexieness overload! Merge!

PR #155, opened 2016-09-16
Fixes issue with zoom buttons on canvas [fixes SCI-451]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-451?jql=key%20in%20(SCI-451)


ZmagoD:
Works like expected! You can merge it! :ship:

PR #160, opened 2016-09-21
Fixes restore experiment error [fixes SCI-464]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-464

### DIFF
--- a/app/controllers/experiments_controller.rb
+++ b/app/controllers/experiments_controller.rb
@@ -76,15 +76,23 @@ class ExperimentsController < ApplicationController
       @experiment.touch(:workflowimg_updated_at)
       flash[:success] = t('experiments.update.success_flash',
                           experiment: @experiment.name)
+
       respond_to do |format|
         format.json do
           render json: {}, status: :ok
         end
+        format.html do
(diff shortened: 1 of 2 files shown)


app/controllers/experiments_controller.rb
```diff
@@ -76,16 +76,24 @@ def update
...
       respond_to do |format|
         format.json do
           render json: {}, status: :ok
         end
+        format.html do
+          redirect_to project_path(@experiment.project)
+        end
       end
     else
+      flash[:alert] = t('experiments.archive.restore_flash')
```
mz3944: Here also other update experiment errors can occur, not just when restoring archived experiment. You can make a check if restoration is in progress, and based on that show flash error.

mz3944: Just use 'experiments.update.error_flash' instead.

mz3944: Just use 'experiments.update.error_flash' instead.

ZmagoD:
is ok now

mz3944:
http://www.relatably.com/m/img/elegant-memes/cmtc36.jpg

:branch => :merge

PR #164, opened 2016-09-22
Fixes issue with new experiment modal and with move of experiments [fixes SCI-400]

okriuchykhin:
Fixes issue with new in modal window and move of experiment between projects when such experiment already exists [fixes SCI-400]

https://biosistemika.atlassian.net/browse/SCI-400

### DIFF
--- a/app/models/experiment.rb
+++ b/app/models/experiment.rb
@@ -375,9 +375,9 @@ class Experiment < ActiveRecord::Base
       project.tags << new_tags
       m.tags << new_tags
     end
-
-    save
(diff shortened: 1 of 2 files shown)


app/models/experiment.rb
```diff
@@ -375,9 +375,9 @@ def move_to_project(project)
       project.tags << new_tags
       m.tags << new_tags
     end
-
-    save
-    touch(:workflowimg_updated_at)
+    result = save
+    touch(:workflowimg_updated_at) if result
+    result
```
ZmagoD: Good very good :)

PR #170, opened 2016-09-23
Add aliases for STDEV/VAR and ignore strings as numbers

jbargu:
https://biosistemika.atlassian.net/browse/SCI-485?filter=-1

### DIFF
--- a/vendor/assets/javascripts/formula.js
+++ b/vendor/assets/javascripts/formula.js
@@ -5038,7 +5038,7 @@
     };
 
     Formula.MAX = function () {
-      var range = Formula.FLATTEN(arguments);
+      var range = Formula.NUMBERS(Formula.FLATTEN(arguments));
       var n = range.length;
       var max = (n > 0) ? range[0] : 0;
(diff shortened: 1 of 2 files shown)


mz3944:
Is this correct?:
A1: =STDEV.P(B3,C3)
A2: =STDEV.S(B3,C3)
A3: =VAR.P(B3,C3)
A4: =VAR.S(B3,C3)
screenshot from 2016-09-27

If so:
pic2

jbargu:
Nice catch, but it works fine. If you remember there're two standard deviations, corrected(sample) and uncorrected(population). The only difference between them is with which number you divide the differences (1/(N-1) and 1/N), where N is number of samples. So in our case formula parser uses only one valid variable, which is 2 and forgets about 'asdasd' - we have one sample. Therefore when calculating corrected (*.S), you divide by zero, which results into dividing by zero (NaN). It should work the same way in Excel.

I'll arrange a date for this branch and master to have a little 'bonding' session as proposed ... :point_right: :ok_hand:

PR #171, opened 2016-09-26
fixed date duplication [fixes SCI-469]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-469


mz3944:
When clicking on _More comments_, the following error is thrown and nothing happend: 
_step_result_comments.self-01dd2a3….js?body=1:57 Uncaught TypeError: Cannot read property 'remove' of undefined_

Also, current date is not added, as if it treats new comments as comments made on the last previous date. If no comments were present before, then no date is shown at all.

ZmagoD:
Solved with PR #176

ZmagoD:
still work to do .. ;)

PR #178, opened 2016-09-27
Fixed wrong CSS formatting and added release notes link to help link.

mz3944:


app/views/shared/_navigation.html.erb
```diff
@@ -61,11 +61,12 @@
               <span class="glyphicon glyphicon-info-sign"></span>
             </a>
             <ul class="dropdown-menu">
-              <li><%= link_to t('nav.help.tutorials'), TUTORIALS_URL %></li>
-              <li><%= link_to t('nav.help.faq'), FAQ_URL %></li>
-              <li><%= link_to t('nav.help.support'), SUPPORT_URL %></li>
-              <li><%= link_to t('nav.help.plans'), PLANS_URL %></li>
-              <li><%= link_to t('nav.help.contact'), CONTACT_URL %></li>
+              <li><%= link_to t('nav.help.tutorials'), TUTORIALS_URL, :target => "_blank" %></li>
```
Ducz0r: If Hound doesn't bark, I will! 😆 (it seems it ignores `.html.erb` files)

I suggest we follow the new Rails syntax for hashes, and use the single quotes for Strings that don't need interpolation, e.g.
```ruby
link_to t('nav.help.tutorials'), TUTORIALS_URL, target: '_blank'
```

mz3944: Ah, nice catch indeed! Totally missed this without any code analysers.

PR #181, opened 2016-09-28
Displaying more information for individual samples in Sample table [SCI-321]

mlorb:
https://biosistemika.atlassian.net/browse/SCI-321?filter=-1


app/assets/javascripts/samples/sample_datatable.js
```diff
@@ -74,6 +81,7 @@ table = $("#samples").DataTable({
     fnDrawCallback: function(settings, json) {
         animateSpinner(this, false);
         changeToViewMode();
+        sampleInfoListener();
```
Ducz0r: I imagine that this callback is called everytime the table is redrawn, e.g. when user sorts the table, navigates to a different navigation page, etc. (basically, when the rows are reloaded via AJAX from server).

I believe there's (albeit very improbable) potential for a JavaScript memory leak here, since `$(".sample_info").on("click", <handler>);` event handlers are kept in memory for elements that have been removed. 

So I don't know, I'd check if maybe those handlers can be removed (`$(".sample_info").off("click");`) before re-drawing the table (maybe the `fnPreDrawCallback` can be used, needs to be checked).

app/views/samples/_info_sample_modal.html.erb
```diff
@@ -0,0 +1,53 @@
...
+  <div class="modal-dialog" role="document">
+    <div class="modal-content">
+      <div class="modal-header">
+        <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">&times;</span></button>
+        <h4 class="modal-title"><%= t('samples.modal_info.title', sample: @sample.name) %></h4>
+      </div>
+
+      <div class="modal-body">
+      <% if @sample.my_modules.count > 0 %>
+        <div class="linked-children-datatable">
```
Ducz0r: Nitpicking, but oh well - it'd make sense to me to also show a text here saying "This sample is assigned to <n> tasks.", similar as how it's shown in protocols modal. Just to be clarify a bit what you're seeing in the table below.

config/locales/en.yml
```diff
@@ -879,6 +879,10 @@ en:
     add_new_sample_type: "Add sample type"
     add_new_sample_group: "Add sample group"
     add_new_column: "Add column"
+    modal_info:
+      head_title: "Information for sample '%{sample}'"
+      title: "This sample is assigned to tasks."
```
Ducz0r: This one's on me! I know I'm bugging you, but I meant to say "This sample is assigned to N tasks.". (I used `<n>` in the text, and somehow Markdown just hid that part of the text)

PR #185, opened 2016-09-29
Initial code for system notifications [SCI-444]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-444

### DIFF
--- a/app/helpers/notifications_helper.rb
+++ b/app/helpers/notifications_helper.rb
@@ -1,10 +1,8 @@
 module NotificationsHelper
-
   def create_system_notification(title, message)
     users = User.where.not(confirmed_at: nil)
     users.each do |u|
       UserNotification.create_notification(u, title, message, :system_message)
     end
   end
-
 end
--- a/app/helpers/notifications_helper.rb
+++ b/app/helpers/notifications_helper.rb
@@ -1,0 +1,10 @@
+module NotificationsHelper
+
+  def create_system_notification(title, message)
+    users = User.where.not(confirmed_at: nil)
+    users.each do |u|
+      UserNotification.create_notification(u, title, message, :system_message)
+    end
+  end
+
+end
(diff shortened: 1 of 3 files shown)


app/helpers/notifications_helper.rb
```diff
@@ -0,0 +1,8 @@
+module NotificationsHelper
+  def create_system_notification(title, message)
+    users = User.where.not(confirmed_at: nil)
+    users.each do |u|
```
Ducz0r: When iterating over large collections, it is **very important** to use `.find_each` instead of `each` (`each` loads all queried objects in memory, which could be deadly when iterating over whole tables, while `find_each` loads objects in batches).

app/helpers/notifications_helper.rb
```diff
@@ -0,0 +1,8 @@
...
+      UserNotification.create_notification(u, title, message, :system_message)
```
Ducz0r: Wasn't one of the purposes of `Notification` - `UserNotification` - `User` database structure that for system notification, a **single** `Notification` object would be created, and linked to many `UserNotification`-s? Because right now, for each system notification, many `Notification` objects are created, and I think that's very redundant.

lib/tasks/notifications.rake
```diff
@@ -0,0 +1,10 @@
+namespace :notifications do
+
+  desc 'Creates new system notification for all active users'
+  task :new_release => :environment do
```
Ducz0r: We will need a parameterized Rake task, because system notifications can be anything. Since parameters in Rake tasks are pretty rudimentary, one way to input a more detailed data into the task is to provide a single parameter, path to a `.yaml` file, and that file has the detailed data specified (e.g. title & message of notification). I mean, your call, maybe it's not neccesary.

Ducz0r:
With @okriuchykhin we decided the Rake task will be changed at a later stage (in couple of days :smile:). So this is ready for merge.

PR #188, opened 2016-10-03
Global notifications

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-10


Ducz0r:
Okay, @ZmagoD cleared some things up with the latest commit. Merging, so the testers can start working on it.

PR #191, opened 2016-10-05
Comments date duplication

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-469

### DIFF
--- a/app/assets/javascripts/comments.js
+++ b/app/assets/javascripts/comments.js
@@ -63,6 +63,7 @@ var Comments = (function() {
           moreBtn.trigger('blur');
         }
 
+        debugger;
         // Reposition dropdown comment options
         scrollCommentOptions(listItem
                               .closest('.content-comments')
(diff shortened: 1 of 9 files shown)


Ducz0r:
Good work, seems nice!

I only have a remark - on project comments, the edit options are displayed weirdly out of place (must be something in `comments.js`, I believe there are 2 ways on how is this `<div>` is positioned). If that's fixed (maybe it's not even related), this is ready for merge :+1:.

ZmagoD:
I had also fixed the drop down position

Ducz0r:
That `useParentOffset` parameter is there for comments that are used on canvas (task comments). With your last commit, you broke comment editing for those comments, so I suggest reverting and finding alternative solution.

Ducz0r:
I think you just have to pass the `useParentOffset = true` parameter when initializing comments editing on home page (projects page).

ZmagoD:
that do the job thank you!

PR #192, opened 2016-10-05
Added links to notification project name and task

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-437

Here I added the links to notification project and notification task.
For other features I will wait for the specifications..

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -60,7 +60,20 @@ class Activity < ActiveRecord::Base
       notification_type = :recent_changes
     end
 
-    task_m = "| #{I18n.t('search.index.module')} #{my_module.name}" if my_module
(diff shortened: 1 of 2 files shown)


Ducz0r:
This looks good :+1:.

The only issue I find is that there is a possibility (albeit a very **unlikely** one) that the user won't have permission to view the notification project/task (e.g., notification is generated, afterwards, the user is removed from the project, and doesn't have permission to view project any more). The thing is that the link is hardcoded into the notification, so we cannot really solve this issue, and the user will simply get 403 page when he clicks on the link.

What I want to say is that this **is ready for merge**. We can solve the unlikely issue mentioned in the previous paragraph at a later stage, if we deem it neccesary.

PR #193, opened 2016-10-05
Global notifications screen [SCI-442]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-442

### DIFF
--- a/app/views/user_notifications/_list.html.erb
+++ b/app/views/user_notifications/_list.html.erb
@@ -1,0 +1,27 @@
+<% @notifications.each do |notification| %>
+  <li class="notification <%= 'unseen' unless notification.already_seen(current_user) %>">
+    <div class="row">
+      <div class="col-xs-3 col-md-1">
+        <% if notification.type_of == 'recent_changes' %>
+        <div class="text-center">
+          <%= image_tag avatar_path(notification.generator_user, :icon_small), class: 'avatar' %>
+        </div>
+        <% end %>
+        <% if notification.type_of == 'assignment' %>
+        <div class="text-center">
+          <span class="assignment"><%= fa_icon 'newspaper-o' %></span>
+        </div>
+        <% end %>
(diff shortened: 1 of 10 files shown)


config/initializers/assets.rb
```diff
@@ -59,6 +59,8 @@
 Rails.application.config.assets.precompile += %w(comments.js)
 Rails.application.config.assets.precompile += %w(projects/show.js)
 Rails.application.config.assets.precompile += %w(projects/introdutory_popup.js)
+Rails.application.config.assets.precompile += %w(step_result_comments.js)
```
Ducz0r: This file is actually not present. It seems it was removed in some recent commits, so it doesn't make sense to include it here...

app/views/user_notifications/_list.html.erb
```diff
@@ -0,0 +1,27 @@
...
+        </div>
+        <% end %>
+        <% if notification.type_of == 'system_message' %>
+        <div class="text-center">
+          <span class="system-message"><i class="glyphicon glyphicon-tower" aria-hidden="true"></i></span>
+        </div>
+        <% end %>
+      </div>
+      <div class="col-xs-9 col-md-11">
+        <strong><%= notification.title %></strong> <br>
```
Ducz0r: Some notification titles include HTML entities, most of them containing `<strong>` entities. So I see 2 ways to solve this:
* in either way, I think you should output `notification.title.html_safe` (because otherwise, users see `<strong>` which is a big no-no;
* if you think it looks okay, you can remove the outer `<strong></strong>` around `notification.title`, or not, decide which way you like it better.

okriuchykhin: Done!

config/initializers/assets.rb
okriuchykhin: Done!

Ducz0r:
This looks ready for merge to me :+1:.

PR #198, opened 2016-10-05
Fix various minor things concerning notifications

Ducz0r:
* Add DB index to `checked` field of `user_notifications`;
* Optimize data query for checking/seeing recent notifications in `user_notification.rb`;
* Add link to settings from both notification lists;
* Add "no notifications" text if no notifications are displayed in either of the lists.

### DIFF
--- a/app/models/user_notification.rb
+++ b/app/models/user_notification.rb
@@ -23,9 +23,6 @@ class UserNotification < ActiveRecord::Base
   end
 
   def self.seen_by_user(user)
-    where(user: user).map do |n|
-      n.checked = true
-      n.save
-    end
+    where(user: user).where(checked: false).update_all(checked: true)
   end
 end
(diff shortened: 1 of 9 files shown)


app/models/user_notification.rb
```diff
@@ -23,9 +23,6 @@ def self.unseen_notification_count(user)
   end
 
   def self.seen_by_user(user)
-    where(user: user).map do |n|
-      n.checked = true
-      n.save
-    end
+    where(user: user).where(checked: false).update_all(checked: true)
```
ZmagoD: :+1:

db/schema.rb
```diff
@@ -280,11 +245,8 @@
     t.integer  "last_modified_by_id"
     t.string   "description"
     t.integer  "space_taken",         limit: 8, default: 1048576, null: false
-    t.integer  "billing_account_id"
```
ZmagoD: what is that?

PR #201, opened 2016-10-06
code notification mailer [SCI-443]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-443

### DIFF
--- a/app/mailers/app_mailer.rb
+++ b/app/mailers/app_mailer.rb
@@ -12,4 +12,12 @@ class AppMailer < Devise::Mailer
     headers = { to: @user.email, subject: (I18n.t('mailer.invitation_to_organization.subject')) }.merge(opts)
     mail(headers)
   end
+
+  def notification(user, notification)
+    @user = user
+    @notification = notification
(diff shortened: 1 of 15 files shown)


app/mailers/app_mailer.rb
```diff
@@ -12,4 +12,12 @@ def invitation_to_organization(user, added_by, organization, opts = {})
...
     mail(headers)
   end
-end
\ No newline at end of file
+
+  def notification(user, notification)
+    @user = user
+    @notification = notification
+    headers = { to: @user.email,
+      subject: "#{I18n.t('notifications.email_title')} - #{notification.title}" }
```
Ducz0r: I think we need to escape all HTML entities from the `notification.title`. Because it can contain stuff like `<strong>`, or `<a>`, and the subject line usually won't render that, so user will get weird stuff.

We're also using `title` more like a message at the moment (it can be quite long!). Maybe it would make more sense to display the `title` in the contents of the email, and have the subject just say "You received a sciNote notification" or something like that.

PR #208, opened 2016-10-11
Fix failing tests + add test to notification model

ZmagoD:


test/models/checklist_item_test.rb
```diff
@@ -1,6 +1,9 @@
 require 'test_helper'
 
 class ChecklistItemTest < ActiveSupport::TestCase
+  should validate_presence_of(:text)
+  should should validate_length_of(:text)
```
Ducz0r: Double `should`?

PR #215, opened 2016-10-12
Remove invitation_to_organization email

Ducz0r:
This email is no longer needed, because if user enables assignment notifications, he/she will already receive an email whenever he/she is added to a new organization/team (via notifications email).


ZmagoD:
I think that this is ready for merge! :ship: :jamaica:

PR #217, opened 2016-10-12
Zd sci 512 global team switch

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-512

PR #219, opened 2016-10-17
Zmago d zd sci 512 global team switch

ZmagoD:
resolve PR #217


db/migrate/20161011074804_add_current_organization_to_user.rb
```diff
@@ -0,0 +1,14 @@
+class AddCurrentOrganizationToUser < ActiveRecord::Migration
```
mz3944: You're missing Frozen?
http://img.lum.dolimg.com/v1/images/homepage_hero_frozen_winter_18c81bd7.jpeg?region=0%2C0%2C480%2C240

ZmagoD:
hmm I have no errors in the console?? For the constants we will have to fix them on merging in master since on the modular-tasks they are not set yet

mz3944:
In Firefox, dropdown elements don't get wrapped:

screenshot from 2016-10-17 15 02 06

PR #223, opened 2016-10-18
Tab selector

Ducz0r:
This MR implements the first version of the tab selector functionality. The tab selection should work, however, work is not done. I'm just pushing this branch so we can continue working on other features ASAP.

Listing below is the work left to do:
* Edit canvas doesn't work. Also, additional stuff needs to be done:
  * as soon as a sample is assigned onto `my_module`, it should get a samples tab shown if it isn't shown yet;
  * you can click "see all activities" link when on canvas; this should enable activities tab if it's not shown yet.
  * when cloning tasks/modules, clone the shown tabs as well.
* Refactor the whole tabs code into an extensible, yet more user-friendly way (`Extensions` initializer class with extensible JSON, which tells all about tabs (good for names, URLs, bad for Ruby code), or maybe extensible Ruby classes architecture (each tab is its own class, completely separated from `ActiveRecords`).

I've also wrote a [short guide](https://gist.github.com/Ducz0r/4fe3b42dd852a1fe14b8eba7a56aa460) (for future reference) on what needs to be done to add a new tab. This is a good example that it's way too much work at the moment, and the extensibility needs to be upgraded.

### DIFF
--- a/app/assets/javascripts/secondary_navigation.js
+++ b/app/assets/javascripts/secondary_navigation.js
@@ -28,7 +28,7 @@
         url: ajaxUrl,
         type: 'POST',
         dataType: 'json',
-        data: {tab: tab, },
+        data: {tab: tab},
         success: function(data) {
           // Reload page
           location.reload();
--- a/app/assets/javascripts/secondary_navigation.js
+++ b/app/assets/javascripts/secondary_navigation.js
@@ -1,0 +1,49 @@
+(function() {
+  'use strict';
+
+  function initializeModuleTabSelector() {
+    var parent = $('[data-role=module-tab-selector]');
+    var ajaxUrl = parent.attr('data-url');
+
+    // Disable checkboxes if neccesary
+    parent.find('li.disabled input[type=checkbox]').attr('disabled');
+
+    // Prevent clicks on checkboxes in dropdown
+    parent.find('input[type=checkbox]').on('click', function(ev) {
+      this.checked = !this.checked;
+      ev.preventDefault();
+      return false;
+    });
+
+    // Toggle tab functionality
+    parent.find('li:not(.disabled) [data-tab]').on('click', function() {
+      var $this = $(this);
+      var tab = $this.attr('data-tab');
+
+      // Enable spinner on dropdown
+      animateSpinner(parent);
+
+      // Do the AJAX call onto server
+      $.ajax({
+        url: ajaxUrl,
+        type: 'POST',
+        dataType: 'json',
+        data: {tab: tab, },
+        success: function(data) {
+          // Reload page
+          location.reload();
+        },
+        error: function(data) {
+          // Disable spinner
+          animateSpinner(parent, false);
+        }
+      });
+    });
+  }
+
+  // Only initialize module tab selector if it's actually
+  // present on the page
+  if ($('[data-role=module-tab-selector]').length) {
+    initializeModuleTabSelector();
+  }
+})();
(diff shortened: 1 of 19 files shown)


app/assets/javascripts/secondary_navigation.js
```diff
@@ -0,0 +1,49 @@
...
+
+      // Enable spinner on dropdown
+      animateSpinner();
+
+      // Do the AJAX call onto server
+      $.ajax({
+        url: ajaxUrl,
+        type: 'POST',
+        dataType: 'json',
+        data: {tab: tab, },
```
ZmagoD: code nazi strikes again

app/controllers/my_modules_controller.rb
```diff
@@ -386,6 +401,14 @@ def check_unassign_samples_permissions
     end
   end
 
+  def check_toggle_tab_permissions
+    render_403 if !can_edit_module_tabs(@my_module) || params['tab'].blank?
```
ZmagoD: code nazi strikes again 
params[:tab]

app/helpers/my_module_tabs_helper.rb
```diff
@@ -0,0 +1,43 @@
...
+    return '' unless my_module.shown_tabs.include?(tab)
+
+    title = t("nav2.modules.#{tab}")
+    if opts['href'].blank?
+      opts['href'] = send("#{tab}_my_module_path", my_module)
+    end
+
+    res = <<-eos
+    <li id="#{opts['id']}" class="#{'active' if active}">
+      <a href="#{opts['href']}" title="#{title}">
```
ZmagoD: I think that we should sanitize the title here.
http://api.rubyonrails.org/classes/ActionView/Helpers/SanitizeHelper.html

app/models/my_module.rb
```diff
@@ -355,6 +356,51 @@ def get_new_position
...
+      end
+    rescue StandardError
+      toggled = false
+    end
+
+    toggled
+  end
+
+  def can_uncheck_tab_protocols?
+    protocols.count.zero?
```
ZmagoD: nice one!

Ducz0r:
> 1. I can't disable the protocols tab ( is this expected behavior)

Well currently, there's no way to remove a protocol from the task/module (you can have a blank protocol though). Do you think it would make sense to be able to hide a Protocols tab if the protocol in the task doesn't have any steps?


> 2. The toggle works very slowly or it is just my comp when I click on the checkbox it not respond I have to click multiple times to make a change.

So the checkbox's actually disabled, so you need to click on the text to make the change. I know the toggle is slow (complete page reload), but I don't think people are gonna use this feature so much that this could be a problem. If we want to fasten this, then we should have to do it via JS; of course I can do that, I just don't think it's neccesary.

> 3. please sanitize the variables you are passing throw html_save.

Sure will!

> 4.I get this error sometimes (randomly):
Render and/or redirect were called multiple times in this action. Please note that you may only call render OR redirect, and at most once per action. Also note that neither redirect nor render terminate execution of the action, so if you want to exit an action after redirecting, you need to do something like "redirect_to(...) and return".

Just out of curiosity, which browser are you using to test? Firefox? Because I've never encountered this error on chrome.

ZmagoD:
1. hmm I don't know maybe we should ask M&S :smile: 
2. :+1: 
3 Chrome. I think that we already encounter this problem and fix that with the return:

>  return redirect_to show  ....

When you will fix that I think this PR will be ready to :ship:

PR #231, opened 2016-10-20
Tab selector fixes

Ducz0r:
* When cloning tasks, the shown tabs are also cloned,
* When samples are assigned onto a task, its samples tab is automatically set to shown,
* Now you can also click on a checkbox from tab selector to toggle tab (previously you had to click on the text),
* Show a flash message when navigating to activity or samples tab of the task from canvas, and the tab is not shown.


ZmagoD:
good job! Merge away :ship:

PR #232, opened 2016-10-20
Fixes different color of buttons in notifications settings [SCI-590]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-590


ZmagoD:
Nice one! Merge away :ship:

PR #233, opened 2016-10-20
Move tooltip to the right [fixes SCI-591]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-591

### DIFF
--- a/app/assets/stylesheets/themes/scinote.scss
+++ b/app/assets/stylesheets/themes/scinote.scss
@@ -1368,23 +1368,24 @@ html.turbolinks-progress-bar::before {
 }
 
 .modal-tooltip {
-  cursor: help;
-  position: relative;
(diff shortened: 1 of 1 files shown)


mlorb:
sci_591
Now team name on the button is not ok.

ZmagoD:
Do you have the last commit 9214adc?

mlorb:
No, I missed it. But now maybe it would be better to move tooltip a little bit more right, because I still don't see the whole name of the other team.
sci-591-2

mlorb:
And I discovered another problem with displaying long names of teams in dropdown. Maybe you can handle this one as well? :)
sci-591-x

mlorb:
Ok, let's leave some work for testers ;) Merge away :ship:

PR #235, opened 2016-10-20
Mz task overview sci 577 v1

mz3944:
Task overview page implemented. Still needs some styling and some possible modifications.


app/models/my_module.rb
```diff
@@ -355,9 +359,78 @@ def get_new_position
...
+  def toggle_tab(tab)
+    toggled = true
+
+    begin
+      # This is done in advance so even if enabling tabs
+      # (where can_uncheck doesn't need to be called),
+      # it's called anyway just to check if the tab
+      # parameter is legal (e.g. it's not an evil,
+      # injected value)
+      can_uncheck = send("can_uncheck_tab_#{tab}?")
```
mz3944: @Ducz0r is this intentionally left as is, or can I cahnge it with 'Object#send'?

app/models/my_module.rb
```diff
@@ -355,9 +359,78 @@ def get_new_position
...
+    toggled
+  end
+
+  def check_tab(tab)
+    # This is done in advance so even if enabling tabs
+    # (where can_uncheck doesn't need to be called),
+    # it's called anyway just to check if the tab
+    # parameter is legal (e.g. it's not an evil,
+    # injected value)
+    send("can_uncheck_tab_#{tab}?")
```
mz3944: @Ducz0r Same here... Just wan't to make sure, since you added this :).

app/views/my_modules/overview.html.erb
```diff
@@ -0,0 +1,19 @@
...
+<div id='widget-container'>
+  <% if @my_module.widgets.count == 0 %>
+  <div data-role="no-widgets-text">
+    <em><%= t("my_modules.overview.widgets.none") %></em>
+  </div>
+  <% else %>
+  <% @my_module.widgets.order(:position).each do |widget| %>
+  <%= render partial: 'widgets/widget.html.erb', locals: { widget: widget } %>
+  <% end %>
+  <% end %>
```
Ducz0r: Can you maybe indent the HTML tags, I almost thought there's a redundant `<% end %>` here? (sorry for nitpicking)

app/views/shared/_secondary_navigation.html.erb
```diff
@@ -116,12 +116,16 @@
         </ul>
       <% elsif module_page? %>
         <ul class="nav navbar-nav navbar-right" style="vertical-align: bottom">
-          <li id="overview-nav-tab" class="<%= "active" if module_overview? %>">
-            <a href="<%= my_module_path(@my_module) %>" title="<%= t('nav2.modules.overview') %>">
-              <span class="hidden-sm hidden-md"><%= t('nav2.modules.overview') %></span>
-              <span class="hidden-xs hidden-lg glyphicon glyphicon-blackboard"></span>
-            </a>
-          </li>
+          <% if can_view_module(@my_module) %>
```
Ducz0r: So I suggest that you keep the previous code (the one that's removed). The `my_module_tab_li` function is meant for all tabs BUT overview (because it also saves the tab's name, in this case `overview`, into the `shown_tabs` JSONB column of `my_module`). The thing about overview is that it's ALWAYS shown, so I wouldn't trouble ourselves with making sure that each time you create `my_module`, you need to paste `overview` into its `shown_tabs` column, etc.

app/views/widgets/_widget.html.erb
```diff
@@ -0,0 +1,55 @@
...
+        <span><%= raw t('my_modules.overview.widgets.new.added_on', timestamp:
+          l(widget.created_at, format: :full), user: widget.added_by.full_name) %></span>
+      </a>
+    </div>
+    <div
+      class='panel-collapse collapse'
+      id='widget-panel-<%= widget.id %>'
+      role='tabpanel'>
+      <div class='panel-body'>
+        <%= render partial: "widgets/widget_#{widget.widget_type}.html.erb", locals: {
```
Ducz0r: I see you're getting the grip on *convention over configuration*, nice la! :heart:

config/initializers/constants.rb
```diff
@@ -199,6 +199,9 @@ class Constants
   # Organization name for default admin user
   DEFAULT_PRIVATE_ORG_NAME = 'My projects'.freeze
 
+  # Task widget types. Order is important, new widget types must be appended!
+  WIDGET_TYPES = %w(protocol results activities samples).freeze
```
Ducz0r: That was my question, isn't the point of freezing variables to make them unmodifiable :smile:?

db/migrate/20161017111700_add_tab_selector_support.rb
```diff
@@ -6,7 +6,7 @@ def up
                default: [],
                null: false
 
-    MyModule.update_all(shown_tabs: %w(protocols results activities samples))
+    MyModule.update_all(shown_tabs: %w(overview) + Constants::WIDGET_TYPES)
```
Ducz0r: I don't think this change is neccesary. If in future, you add stuff to `Constants::WIDGET_TYPES`, this migration will not work as intended any more. I believe hardcoded values are better here. And, as mentioned previously, I'd remove the `overview` tab here as it's omnipresent (yeah, I just called it that :yum:).

db/migrate/20161018134050_create_widgets.rb
```diff
@@ -0,0 +1,23 @@
+class CreateWidgets < ActiveRecord::Migration
+  def change
+    create_table :widgets do |t|
+      t.integer :widget_type, null: false
+      t.integer :position, null: false
+      t.text :properties
```
Ducz0r: Didn't we talk about using `jsonb` here?
e.g.
```ruby
  t.jsonb :properties, null: false, default: '{}'
```

mz3944: Yes, but that's PostgreSQL specific. It is still interpreted as JSON, as specified in widget model, and by doing it this way, it works for all DB-s.

app/models/widget.rb
```diff
@@ -0,0 +1,16 @@
+class Widget < ActiveRecord::Base
+  enum widget_type: Constants::WIDGET_TYPES
```
mz3944: Don't we need just one `enum`, which holds all of the widget types, just as I did it? If so, then yes, if not, why not? :)

Ducz0r: Yes, but how will an engine, which is a pluggable Rails application, inject its own widget Strings into `Constants::WIDGET_TYPES`? If you freeze this variable, any calls (e.g. `Constants::WIDGET_TYPES << :my_new_widget`) will fail.

app/views/my_modules/overview.html.erb
mz3944: I'm actually using the _atom-beautify_, as specified in the wiki codestyle guide. It formats the code so that HTML elements are indented correctly, but Ruby code stays on the same indentation level as the previous HTML tag. To make application (_*.html.erb_) code consistent, we should probably do it this way. We should also probably make a pull request which formats all of the _*.html.erb_ code with the _atom-beautify_, as it can be done in a jiffy.

db/migrate/20161018134050_create_widgets.rb
Ducz0r: Well we're already married to PostgreSQL anyway, so that's why I don't see why not use its advanced features.

app/models/widget.rb
mz3944: Oh my bad, I remember now, actually meant to do it without the freeze, but the _RuboCop_ mislead me again... I spend some time researching whether to use text or `enum`, and come to the conclusion that using `text` (or `jsonb` perhaps) would give us no additional benefits (or in other words, problems that could occur with `enum` could very well occur with `text`).

mz3944: Oh my bad, I remember now, actually meant to do it without the freeze, but the _RuboCop_ mislead me again... I spend some time researching whether to use text or `enum`, and come to the conclusion that using `text` (or `jsonb` perhaps) would give us no additional benefits (or in other words, problems that could occur with `enum` could very well occur with `text`).

db/migrate/20161017111700_add_tab_selector_support.rb
```diff
@@ -3,10 +3,10 @@ def up
     add_column :my_modules,
                :shown_tabs,
                :jsonb,
-               default: [],
-               null: false
+               null: false,
+               default: {}
 
-    MyModule.update_all(shown_tabs: %w(protocols results activities samples))
```
mz3944: Shame...

app/models/my_module_widget.rb
```diff
@@ -0,0 +1,17 @@
+class MyModuleWidget < ActiveRecord::Base
+  enum widget_type: Constants::MY_MODULE_WIDGET_TYPES
+
+  validates :widget_type, presence: true
+  validates :position,
+            presence: true,
+            numericality: { greater_than_or_equal_to: 0 }
+  validates :properties, exclusion: { in: [nil] }
+  validates :added_by_id, presence: true
+  validates :my_module_id, presence: true, uniqueness: { scope: :position }
```
mz3944: The effect is the same, but will change it just to be clear.

PR #244, opened 2016-10-24
sets current organization id as a current user first organization if is the case that the current_organization is not present  [fixes SCI-608]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-608


mz3944:
I got nothing on you babe! :ok_woman: Merge the damn thing!

PR #245, opened 2016-10-25
Fix a String

Ducz0r:
Closes SCI-611


Ducz0r:
Merging, it's a one-liner fix...

PR #247, opened 2016-10-25
Add check in application controller to always set current org if nil

Ducz0r:
This should hopefully fix some awkward situations where `current_organization_id` can be nil (e.g. when inviting a new user to the team).


ZmagoD:
Looks nice! Merge away :ship:

PR #248, opened 2016-10-25
Add results widget view [SCI-579]

mlorb:
https://biosistemika.atlassian.net/browse/SCI-579?filter=-1

There is still a bug with displaying tables... I am going to deal with it, after all the widgets will be merged.


app/views/widgets/my_module/_widget_results.html.erb
```diff
@@ -1,2 +1,66 @@
...
+<%= javascript_include_tag "numeral" %>
+<%= javascript_include_tag "numeric" %>
+<%= javascript_include_tag "md5" %>
+<%= javascript_include_tag "jstat" %>
+<%= javascript_include_tag "formula" %>
+<%= javascript_include_tag "parser" %>
+<%= javascript_include_tag "ruleJS" %>
+<%= javascript_include_tag "handsontable.formula" %>
+<%= javascript_include_tag "big.min" %>
+<%= stylesheet_link_tag "handsontable.formula" %>
```
Ducz0r: Do we need all those JS files here? Just seems a lot of JS. I think the only neccesary JS is `handsontable.full.min` (and maybe the formula libs, I don't know about them).

I would suggest that you create a single JS file only used here (e.g. `app/assets/javascripts/widgets/my_module/widget_results.js`), and only include the bare-minimum code needed for this widget (I think only the `initHandsOnTables` from `results.js.erb` needs to be called).

PR #250, opened 2016-10-25
Widgets are removed/added when tabs are removed/added.

mz3944:


app/models/my_module.rb
```diff
@@ -373,12 +373,22 @@ def toggle_tab(tab)
 
       if shown_tabs.include?(tab)
         if can_uncheck
+          my_module_widgets.where(widget_type: MyModuleWidget.widget_types[tab])
```
Ducz0r: I would add comment here just so we know why is this so and that it's a dirty "hack" that might be removed in future.

app/models/my_module.rb
```diff
@@ -373,12 +373,22 @@ def toggle_tab(tab)
...
+                           .first.destroy
+
           shown_tabs.delete(tab)
           save
         else
           toggled = false
         end
       else
+        MyModuleWidget.new(
```
Ducz0r: Also write a comment here I guess...

app/models/my_module_widget.rb
```diff
@@ -14,4 +14,16 @@ class MyModuleWidget < ActiveRecord::Base
              class_name: 'User',
              inverse_of: :last_modified_my_module_widgets
   belongs_to :my_module, inverse_of: :my_module_widgets
+
+  def destroy
+    delete
+
+    # Update position on other module widgets of this widget's module
```
Ducz0r: I would put this code into `after_destroy` callback function if it can be done, and remove the `destroy` function (keep its base class definition).

mz3944: Nice call.

PR #251, opened 2016-10-25
Add protocols widget view and comments for all widgets [SCI-578, SCI-583]

mlorb:
https://biosistemika.atlassian.net/browse/SCI-583?filter=-1
https://biosistemika.atlassian.net/browse/SCI-578?filter=-1


app/assets/stylesheets/extend/bootstrap.scss
```diff
@@ -58,6 +58,19 @@
...
+
+.btn-info:hover {
+  color: #fff !important;
+  background-color: #2793ce !important;
+  border-color: #2384b8 !important;
+}
+
 @media (max-width: 1000px) {
   .navbar-header {
     float: none;
```
Ducz0r: Are those color changes in `bootstrap.scss` really needed?
1.) We never use color constants directly in `.scss` files, but always define them elsewhere.
2.) We should define our own style in `scinote.scss`, not this file.

app/controllers/my_modules_controller.rb
```diff
@@ -341,6 +341,7 @@ def load_vars
     if @my_module
       @experiment = @my_module.experiment
       @project = @my_module.experiment.project if @experiment
+      @protocol = @my_module.protocol
```
Ducz0r: I suggest you remove this line, and handle it differently in the protocol widget (see comment there).

app/views/widgets/my_module/_widget_protocol.html.erb
```diff
@@ -1,2 +1,180 @@
 <div class="widget-content">
+  <div data-role="protocol-status-bar" style="display: inline;">
```
Ducz0r: I suggest you simply specify `protocol` here, and then use it (instead of `@protocol`) elsewhere in this file:
```
<% protocol = @my_module.protocol %>
```

app/views/widgets/my_module/_widget_protocol.html.erb
```diff
@@ -1,2 +1,180 @@
...
+		      <% end %>
+		    </a>
+		    <a type="button" class="btn btn-warning dropdown-toggle" disabled="disabled">
+		      <span class="caret"></span>
+		    </a>
+		  <% end %>
+		</div>
+	  </div>
+	</div>
+	<%= javascript_include_tag("my_modules/protocols/protocol_status_bar") %>
```
Ducz0r: Is this needed?

PR #253, opened 2016-10-26
Change edit avatar in profile [SCI-610]

mlorb:
https://biosistemika.atlassian.net/browse/SCI-610

### DIFF
--- a/app/assets/stylesheets/themes/scinote.scss
+++ b/app/assets/stylesheets/themes/scinote.scss
@@ -1453,3 +1453,28 @@ html.turbolinks-progress-bar::before {
     display: inline;
   }
 }
+
+//edit avatar
+#avatar-container {
+  width: 100px;
+  height: 100px;
+  position: relative;
+}
+
+#avatar-image,
+#avatar-edit {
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/themes/scinote.scss
```diff
@@ -1453,3 +1453,28 @@ html.turbolinks-progress-bar::before {
...
+
+//edit avatar
+.avatar-container {
+  height: 100px;
+  position: relative;
+  width: 100px;
+}
+
+.avatar-image,
+.avatar-edit {
```
mz3944: I think height and width are redundant here.

mz3944:
Also, If image doesn't have enough height or is horizontally stretched, use some contrasted background colour for avatar, as it doesn't fit the whole avatar area.

mz3944:
Clicking on any part of the avatar should trigger avatar upload section, not just when clicking on edit button. If nothing else, mobile users will be happier, and there less chance of user getting confused.

mz3944:
Nice work! You're good to go sis!

PR #259, opened 2016-11-02
Added additional activities [SCI-605]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-605


mlorb:
load protocol from file is not working anymore

mlorb:
For sample un/assign activity, you get in the activity log the list of all selected samples. There should be written only those samples that are being un/assigned (their previous state should be checked).

okriuchykhin:
Fixed samples assign/unassign logic

mlorb:
The problem is still present, if you assign already assigned sample (the same with unassigned ones).

mlorb:
great, works as expected... merge away :+1:

PR #260, opened 2016-11-02
Add notification for removing user from team and removing user from project [SCI-614]

mlorb:
https://biosistemika.atlassian.net/browse/SCI-614

### DIFF
--- a/app/controllers/user_projects_controller.rb
+++ b/app/controllers/user_projects_controller.rb
@@ -191,6 +191,7 @@ class UserProjectsController < ApplicationController
           unassigned_by_user: current_user.full_name
         )
       )
+      generate_notification(current_user, @up.user, @project)
 
       respond_to do |format|
         format.html {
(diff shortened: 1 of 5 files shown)


app/controllers/user_projects_controller.rb
```diff
@@ -275,6 +276,24 @@ def check_delete_permisisons
     end
   end
 
+  def generate_notification(user, target_user, project)
```
ZmagoD: I think that this method is duplicated in 2 places now here and in settings_controller. 
Can you refactor the method that will work in both cases and put it in a helper file?

app/controllers/users/settings_controller.rb
```diff
@@ -579,20 +582,27 @@ def update_user_organization_params
     )
   end
 
-  def generate_notification(user, target_user, role, org)
```
ZmagoD: here! see above

app/helpers/notifications_helper.rb
```diff
@@ -17,4 +17,40 @@ def create_system_notification(title, message)
   def send_email_notification(user, notification)
     AppMailer.delay.notification(user, notification)
   end
+
+  def generate_notification(user, target_user, org, role, project)
+    if org
+      if role
```
ZmagoD: fix that and then is OK

PR #262, opened 2016-11-02
Enables notifications to be extended

ZmagoD:
Part of https://biosistemika.atlassian.net/browse/SCI-607

### DIFF
--- a/app/models/notification.rb
+++ b/app/models/notification.rb
@@ -3,7 +3,7 @@ class Notification < ActiveRecord::Base
   has_many :users, through: :user_notifications
   belongs_to :generator_user, class_name: 'User'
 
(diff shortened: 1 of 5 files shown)


app/models/notification.rb
```diff
@@ -3,7 +3,7 @@ class Notification < ActiveRecord::Base
   has_many :users, through: :user_notifications
   belongs_to :generator_user, class_name: 'User'
 
-  enum type_of: [:assignment, :recent_changes, :system_message]
+  enum type_of: Constants::NOTIFICATIONS_TYPES
```
Ducz0r: Could we move this, and all other extendable arrays etc., into a separate file (e.g. `Extends::`, or anything better if you can come up with)?

PR #280, opened 2016-11-11
Show only protocols from Protocol Management in users statistics [SCI-658]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-658


Ducz0r:
Seems good :+1:.

PR #281, opened 2016-11-11
hide flash after 5 sec [fixes SCI-660]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-660


mlorb:
Wouldn't be better to merge this fix to gamification branch? Because bug was reported from there. Otherwise works as expected. Great job, merge away :+1:  :car:

PR #286, opened 2016-11-14
Save first time tutorial Activity-s without callbacks [SCI-655]

Ducz0r:
This PR adds a new, pretty useful little gem called [sneaky-save](https://github.com/einzige/sneaky-save). The Gem adds a new function onto `ActiveRecord`, called `sneaky_save`, which saves an entity into the database via direct SQL, therefore omitting any validations as well as callbacks (which were a pain for fixing the mentioned issue).

For reference, the other solution, [`skip_callback`](http://apidock.com/rails/ActiveSupport/Callbacks/ClassMethods/skip_callback) method that is part of  `ActiveRecord` interface is pretty bad, because it's not thread-locked, so big no-go (especially because we're also using `delayed_jobs` here).

https://biosistemika.atlassian.net/browse/SCI-655


ZmagoD:
Works nice great job! :ship:

PR #288, opened 2016-11-15
Fix a bug where user could be assigned to same project twice [SCI-675]

Ducz0r:
I've also checked production database, and no project has two same users assigned onto it, so no migration is needed for this.

https://biosistemika.atlassian.net/browse/SCI-675


mlorb:
Works as expected. Great job :+1: :tada: :facepunch:

PR #294, opened 2016-11-15
Import protocol steps' rich text styling when importing protocols [SCI-674]

Ducz0r:
https://biosistemika.atlassian.net/browse/SCI-674

Also shield imported protocols from CORS.
@ZmagoD, can you check this MR and check if I need to feed any more parameters into the `sanitize` method?


ZmagoD:
It looks OK! :ship:

PR #302, opened 2016-11-17
Correctly import protocols with rich text [SCI-691]

Ducz0r:
https://biosistemika.atlassian.net/browse/SCI-691


ZmagoD:
It works like expected. Good job! :ship:

PR #303, opened 2016-11-17
Added additional tutorial step for showing protocol link. Tutorial st…

mz3944:
…eps 1-3 are now completelly handled in JS and not in HTML (preferred way).

https://biosistemika.atlassian.net/browse/SCI-571


ZmagoD:
It looks OK I have only one remark but maybe is not related to your task. When the user is redirected to the new page the introjs overlay disappear and the user can click the buttons on the page an end the tutorial with that. Should we handle that somehow? Good job :ship:

mz3944:
I'll open new issue for that.

PR #312, opened 2016-11-22
Integrate reCaptcha into sciNote [SCI-690]

mlorb:
https://biosistemika.atlassian.net/browse/SCI-690

### DIFF
--- a/app/views/users/registrations/new.html.erb
+++ b/app/views/users/registrations/new.html.erb
@@ -38,6 +38,13 @@
       <span><small><%= t 'users.registrations.new.team_name_help' %></small></span>
     </div>
 
+    <% if Rails.configuration.x.enable_recaptcha %>
+      <div class="form-group">
+        <%= recaptcha_tags %>
+        <%= flash[:recaptcha_error] %>
+      </div>
+    <% end %>
+
     <div class="form-group">
       <%= f.submit "Sign up", class: "btn btn-primary" %>
     </div>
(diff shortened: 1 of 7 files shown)


app/views/users/registrations/new.html.erb
```diff
@@ -38,6 +38,13 @@
       <span><small><%= t 'users.registrations.new.team_name_help' %></small></span>
     </div>
 
+    <% if Rails.configuration.x.enable_recaptcha %>
```
Ducz0r: I would style this at least a bit in case of an error (e.g.):
```erb
<% if Rails.configuration.x.enable_recaptcha %>
      <div class="form-group <%= 'has-error' if flash[:recaptcha_error] %>">
        <%= recaptcha_tags %>
        <% if flash[:recaptcha_error] %>
          <span class="help-block">
            <%= flash[:recaptcha_error] %>
          </span>
        <% end %>
      </div>
    <% end %>
```
Or even better, since this will be used on multiple parts of the application, maybe even refactor it into a helper method (e.g. `recaptcha_input_tag`).

If you'd refactor it, I would make a new helper (e.g. `form_tag_helper.rb` or something like that).

mlorb:
I added helper method. Is it ok now?
In SCI_705 I still have to add switch to toggle email confirmations.

Ducz0r:
Good job :+1:. My last comment would be - can you also default to `'false'` value of `ENABLE_RECAPTCHA` on test environments? After that, I think this is ready for merge!

PR #315, opened 2016-11-23
Enable reCaptcha and auto log in on confirm invitation form [SCI-706]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-706

### DIFF
--- a/app/views/users/invitations/edit.html.erb
+++ b/app/views/users/invitations/edit.html.erb
@@ -31,6 +31,17 @@
     <span><small><%= t 'users.registrations.new.team_name_help' %></small></span>
   </div>
 
+  <% if Rails.configuration.x.enable_recaptcha %>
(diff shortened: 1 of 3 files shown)


app/views/users/invitations/edit.html.erb
```diff
@@ -31,6 +31,17 @@
     <span><small><%= t 'users.registrations.new.team_name_help' %></small></span>
   </div>
 
+  <% if Rails.configuration.x.enable_recaptcha %>
```
Ducz0r: You can use `<%= recaptcha_input_tag %>` that @mlorb implemented (see [form_tag_helper.rb](https://github.com/biosistemika/scinote-web/blob/simplify-signup/app/helpers/form_tag_helper.rb)) here.

okriuchykhin:
Done!

PR #323, opened 2016-11-28
Fix task comments not properly showing in reports

Ducz0r:
Closes [SCI-733](https://biosistemika.atlassian.net/browse/SCI-733).


ZmagoD:
Good job! :ship:

PR #325, opened 2016-11-29
Added form for adding new columns to samples table into dropdown [SCI-712]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-712


okriuchykhin:
Thirst two points are now fixed. 
I think it's better to fix column visibility persistence after merging all code together and to use DB for it.

PR #331, opened 2016-11-30
Added automatic sign in with token authentication, needed when signin…

mz3944:
…g in from different webpage. [closes SCI-689].

### DIFF
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -17,18 +17,27 @@ class Users::SessionsController < Devise::SessionsController
   # end
 
   # Singing in with authentication token (needed when signing in automatically
-  # from another website)
+  # from another website). NOTE: For some reason URL needs to end with '/'.
   def auth_token_create
     user = User.find_by_email(params[:user_email])
-    if user.authentication_token == params[:user_token][0..-2]
+    user_token = params[:user_token]
+    # Remove trailing slash if present
+    user_token.chop! if !user_token.nil? && user_token.end_with?('/')
+
+    if user && user.authentication_token == user_token
       sign_in(:user, user)
+      # This will cause new token to be generated
+      user.update(authentication_token: nil)
+
+      redirect_url = root_path
     else
-      flash[:error] = t('devise.sessions.auth_token_createwrong_credentials')
+      flash[:error] = t('devise.sessions.auth_token_create.wrong_credentials')
+      redirect_url = new_user_session_path
     end
 
     respond_to do |format|
       format.html do
-        redirect_to root_path
+        redirect_to redirect_url
       end
     end
   end
@@ -39,5 +48,4 @@ class Users::SessionsController < Devise::SessionsController
   def configure_sign_in_params
     devise_parameter_sanitizer.for(:sign_in) << :attribute
   end
-
 end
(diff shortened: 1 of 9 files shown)


config/routes.rb
```diff
@@ -289,5 +289,6 @@
   devise_scope :user do
     get 'avatar/:id/:style' => 'users/registrations#avatar', as: 'avatar'
     post 'avatar_signature' => 'users/registrations#signature'
+    get 'auth_token_sign_in' => 'users/sessions#auth_token_create'
```
Ducz0r: I know I'm nitpicking, but the normal `sign_in` is located at `users/sign_in`. I believe this route should also be at `users/auth_token_sign_in`.

app/controllers/users/sessions_controller.rb
```diff
@@ -16,6 +16,23 @@ class Users::SessionsController < Devise::SessionsController
   #   super
   # end
 
+  # Singing in with authentication token (needed when signing in automatically
+  # from another website)
+  def auth_token_create
+    user = User.find_by_email(params[:user_email])
+    if user.authentication_token == params[:user_token][0..-2]
```
Ducz0r: And also, can you only trim the `params[:user_token]` if they end with trailing slash?
e.g.
```ruby
user_token = params[:user_token]
# Remove trailing slash if present
user_token.chop! if user_token.end_with?('/')
```

app/controllers/users/sessions_controller.rb
```diff
@@ -16,6 +16,23 @@ class Users::SessionsController < Devise::SessionsController
...
+      sign_in(:user, user)
```
Ducz0r: I believe that we want this token to be a single-use. Therefore, after user is successfully signed in via this token, we need to refresh it (create a new one). This can simply be done by setting it to `nil` ([docs](https://github.com/gonzalo-bulnes/simple_token_authentication#tokens-generation)). So I propose that after `sign_in(:user, user)` line, add following:

```ruby
sign_in(:user, user)

# This will cause new token to be generated
user.update(authentication_token: nil)
```

mz3944: I also thought that would be a good idea, but then I read it's safe enough... But I guess It's better to be more safe than sorry :).

config/routes.rb
mz3944: Ah, your right, my bad!

mz3944:
Also fixed so that the correct error flash is shown when wrong credentials are provided.

app/controllers/users/sessions_controller.rb
```diff
@@ -17,18 +17,27 @@ class Users::SessionsController < Devise::SessionsController
...
-  # from another website)
+  # from another website). NOTE: For some reason URL needs to end with '/'.
   def auth_token_create
     user = User.find_by_email(params[:user_email])
-    if user.authentication_token == params[:user_token][0..-2]
+    user_token = params[:user_token]
+    # Remove trailing slash if present
+    user_token.chop! if !user_token.nil? && user_token.end_with?('/')
+
+    if !user.nil? && user.authentication_token == user_token
```
Ducz0r: Simply use `if user.try(:authentication_token) == user_token`

mz3944: This would cause error for 'http://my.scinote/users/auth_token_sign_in'.

PR #335, opened 2016-12-05
fixes inport eln error message and spinner [fixes SCI-752]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-752

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -552,7 +552,7 @@ en:
         confirm: "Copy to repository"
       load_from_file_flash: "Successfully loaded the protocol from the file."
       load_from_file_error: "Failed to load the protocol from file."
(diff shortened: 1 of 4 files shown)


config/locales/en.yml
```diff
@@ -552,6 +552,7 @@ en:
         confirm: "Copy to repository"
       load_from_file_flash: "Successfully loaded the protocol from the file."
       load_from_file_error: "Failed to load the protocol from file."
+      load_from_file_size_error: "Failed to load the protocol from file. Limit is 50Mb."
```
Ducz0r: Can you make this String parametrized, and feed into it the `FILE_MAX_SIZE_MB` constant (so, if you change the constant, this String still holds true)?

PR #337, opened 2016-12-07
limits max size of all uploaded files at the one time

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-740

Actually this is a workaround of the problem of uploading multiple files in protocol step so now you can upload a max 50 MB of different files at one time.

### DIFF
--- a/app/assets/javascripts/sitewide/form_validators.js.erb
+++ b/app/assets/javascripts/sitewide/form_validators.js.erb
@@ -108,6 +108,7 @@ function filesValidator(ev, fileInputs, fileTypeEnum, canBeEmpty) {
   if (fileInputs.length) {
     var filesPresentValid = canBeEmpty || filesPresentValidator(ev, fileInputs);
     var filesSizeValid = filesSizeValidator(ev, fileInputs, fileTypeEnum);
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/sitewide/form_validators.js.erb
```diff
@@ -142,11 +143,28 @@ function filesSizeValidator(ev, fileInputs, fileTypeEnum) {
...
+      return "<%= I18n.t('general.file.total_size', size: Constants::FILE_MAX_SIZE_MB) %>".strToErrorFormat();
+    }
+  }
+
   // Check if any file exceeds allowed size limit
   var filesSizeValid = true;
   _.each(fileInputs, function(fileInput) {
     var file = fileInput.files[0];
     var assetError = getFileTooBigError(file);
+    var assetError = checkFilesTotalSize(fileInputs);
```
Ducz0r: So if I understand correctly, this is done for each file? Couldn't you just call this function (`checkFilesTotalSize(fileInputs)`) once outside of the loop, save its results into a variable, and then check that inside the loop?

PR #338, opened 2016-12-08
Office integration

Nbernot:
Office integration ready for review


app/controllers/protocols_controller.rb
```diff
@@ -313,155 +313,193 @@ def unlink
...
+          flash[:success] = t(
+            "my_modules.protocols.load_from_repository_flash",
+          )
+          flash.keep(:success)
+          format.json { render json: {}, status: :ok }
+        end
       else
-        # Everything good, display flash & render 200
-        flash[:success] = t(
-          "my_modules.protocols.load_from_repository_flash",
```
ZmagoD: use single quote

app/controllers/protocols_controller.rb
```diff
@@ -313,155 +313,193 @@ def unlink
...
+          format.json {
+            render json: {
+              message: t("my_modules.protocols.load_from_repository_error")
+            },
+            status: :bad_request
+          }
+        else
+          # Everything good, display flash & render 200
+          flash[:success] = t(
+            "my_modules.protocols.load_from_repository_flash",
```
ZmagoD: use single quote

app/controllers/protocols_controller.rb
```diff
@@ -313,155 +313,193 @@ def unlink
...
-          render json: {
-            message: t("my_modules.protocols.load_from_repository_error")
-          },
-          status: :bad_request
-        }
+        if transaction_error
+          # Bad request error
+          format.json {
+            render json: {
+              message: t("my_modules.protocols.load_from_repository_error")
```
ZmagoD: use single quote

app/controllers/protocols_controller.rb
```diff
@@ -313,155 +313,193 @@ def unlink
...
-          render json: { status: :error }, status: :bad_request
-        }
+        if transaction_error
+          format.json {
+            render json: { status: :error }, status: :bad_request
+          }
+        else
+          # Everything good, display flash & render 200
+          flash[:success] = t(
+            "my_modules.protocols.load_from_file_flash",
```
ZmagoD: use single quote

app/controllers/result_assets_controller.rb
```diff
@@ -136,6 +147,19 @@ def update
     elsif @result.archived_changed?(from: true, to: false)
       render_403
     else
+      if previous_asset.locked?
+        @result.errors.add(:asset_attributes,
+                           I18n.t('result_assets.edit.locked_file_error'))
```
ZmagoD: it should work with just t('result_assets.edit.locked_file_error')

app/utilities/subdomain.rb
```diff
@@ -0,0 +1,23 @@
...
+  		return (request.subdomain.present? && request.subdomain == ENV['USER_SUBDOMAIN'])
+  	else
+    	return true
+	end
+  end
+end
+
+class WopiSubdomain
+  def self.matches?(request)
+    if ENV['WOPI_ENABLED'] == "true"
```
ZmagoD: if true == true ???

app/models/protocol.rb
```diff
@@ -588,6 +590,10 @@ def destroy_contents(current_user)
     self.reload
```
ZmagoD: I think that you don't have to use self here

app/models/user.rb
```diff
@@ -252,12 +253,43 @@ def last_activities(last_activity_id = nil, per_page = 10)
       .uniq
   end
 
+  def self.find_by_valid_wopi_token(token)
+    Rails.logger.warn "WOPI: searching by token #{token}"
+    User
+    .joins("LEFT OUTER JOIN tokens ON user_id = users.id")
```
ZmagoD: I think you can use just joins(:tokens) here

app/models/user.rb
```diff
@@ -252,12 +253,43 @@ def last_activities(last_activity_id = nil, per_page = 10)
...
+  end
+
+  def get_wopi_token
+  	  # WOPI does not have a good way to request a new token, so a new token should be provided each time this is called, while keeping any old tokens
+  	  # as long as they have not yet expired
+
+  	tokens = Token.where("user_id = ?", id).distinct
+
+  	for token in tokens
+  		if (token.ttl < Time.now.to_i)
```
ZmagoD: use double spaces not tabs

app/models/user.rb
```diff
@@ -252,12 +253,43 @@ def last_activities(last_activity_id = nil, per_page = 10)
...
+  			token.delete
+  		end
+  	end
+
+    token_string = Devise.friendly_token(20) + "-"  + id.to_s
```
ZmagoD: single quote

app/utilities/subdomain.rb
```diff
@@ -0,0 +1,23 @@
+class UserSubdomain
+  def self.matches?(request)
+  	if ENV['USER_SUBDOMAIN']
+  		return (request.subdomain.present? && request.subdomain == ENV['USER_SUBDOMAIN'])
```
ZmagoD: don't use tabs

app/views/results/_result_asset.html.erb
```diff
@@ -1,9 +1,36 @@
 <% if can_view_or_download_result_assets(result.my_module) %>
   <%= link_to download_asset_path(result.asset), data: {no_turbolink: true} do %>
     <%= image_tag(preview_asset_path result.asset) if result.asset.is_image? %>
-    <p><%= truncate(result.asset.file_file_name, length: 50) %></p>
+		<% if wopi_file?(result.asset) %>
+			<p style='display: inline-block'>
+				<%= file_extension_icon(result.asset) %>
```
ZmagoD: this is all messed up don't use tabs

db/migrate/20161129111100_add_wopi.rb
```diff
@@ -0,0 +1,53 @@
...
+      t.string :action, null: false
+      t.string :extension, null: false
+      t.string :urlsrc, null: false
+      t.integer :wopi_app_id, null: false
+    end
+
+    create_table :tokens do |t|
+      t.string :token, null: false
+      t.integer :ttl, null: false
+      t.integer :user_id, null: false
```
ZmagoD: t.references :user will do the job here and will generate a foreign key or even better t.belongs_to(:user, foreign_key: true)

app/assets/javascripts/my_modules/protocols.js
```diff
@@ -528,6 +528,8 @@ function initImport() {
         if (data.status === "ok") {
           // Simply reload page
           location.reload();
+        } else if (data.status === 'locked') {
+          alert(I18n.t("my_modules.protocols.load_from_file_error_locked"));
```
ZmagoD: you can use only t('.....')

app/utilities/subdomain.rb
Ducz0r: Well, if `ENV['WOPI_ENABLED'] == 'false'`, then this expression evaluates to false which is good. But if you simply put the statement `if ENV['WOPI_ENABLED']`, even in case that it would be set to `false`, it would evaluate to true, which is not something we want? Am I right?

app/models/user.rb
Ducz0r: `.joins` method makes an inner join, so yeah - if outer join is not needed, use `.joins`, otherwise leave it as it is.

app/models/user.rb
Ducz0r: Plus it might look better this way: `token_string = "#{Devise.friendly_token(20)}-#{id}"`.

app/assets/javascripts/my_modules/protocols.js
Ducz0r: From JS as well?

ZmagoD: I didn't notice that is a js file :smile:

app/utilities/subdomain.rb
Ducz0r: Another option is to use:
```ruby
YAML.load(var) # -> true/false if it's one of the below
```
It accepts the following strings:
```
y|Y|yes|Yes|YES|n|N|no|No|NO
|true|True|TRUE|false|False|FALSE
|on|On|ON|off|Off|OFF
```

ZmagoD: svašta :)

PR #341, opened 2016-12-09
Add edit & delete column functionality to columns dropdown

Ducz0r:
Closes [SCI-711](https://biosistemika.atlassian.net/browse/SCI-711).

### DIFF
--- a/app/assets/javascripts/samples/sample_datatable.js
+++ b/app/assets/javascripts/samples/sample_datatable.js
@@ -808,6 +808,7 @@ function changeToEditMode() {
 (function(table) {
   'use strict';
 
+  var dropdown = $('#samples-columns-dropdown');
   var dropdownList = $('#samples-columns-list');
   var editMode = false;
 
@@ -912,7 +913,11 @@ function changeToEditMode() {
         var editClass = (editable) ? '' : 'disabled';
         var delClass = (deletable) ? '' : 'disabled';
         var html =
-          '<li data-position="' + colIndex + '" class="' + visLi + '">' +
+          '<li ' +
+          'data-position="' + colIndex + '" ' +
+          'data-id="' + $(el).attr('id') + '" ' +
+          'class="' + visLi + '"' +
+          '>' +
           '<i class="grippy"></i> ' +
           '<span class="text">' + el.innerText + '</span> ' +
           '<input type="text" class="text-edit form-control" style="display: none;" />' +
@@ -1004,7 +1009,6 @@ function changeToEditMode() {
 
       cancelEditMode();
 
-
       var self = $(this);
       var li = self.closest('li');
       var text = li.find('.text');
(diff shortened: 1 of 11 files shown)


Ducz0r:
So, a lot of those issues you encountered are actually not related to this task, and will be solved by @okriuchykhin (see [SCI-779](https://biosistemika.atlassian.net/browse/SCI-779)) - they're also present on `improved-sample-table` branch.

I guess the one issue with *this* PR is the eye icon problems. Will fix.

app/assets/javascripts/samples/sample_datatable.js
```diff
@@ -971,12 +999,200 @@ function changeToEditMode() {
...
+      event.stopPropagation();
+      var self = $(this);
+      var li = self.closest('li');
+      editColumn(li);
+    });
+
+    // On enter click while editing column text
+    dropdownList.on('keydown', 'input.text-edit', function(event) {
+      if (event.keyCode === 13) {
+        event.preventDefault();
```
mz3944: Maybe use `event.stopPropagation();` instead, as in everywhere else? If it doesn't make a difference of course.

app/helpers/permission_helper.rb
```diff
@@ -674,6 +674,16 @@ def can_create_custom_field_in_organization(organization)
     is_normal_user_or_admin_of_organization(organization)
   end
 
+  def can_edit_custom_field(custom_field)
+    custom_field.user == current_user ||
+      is_admin_of_organization(custom_field.organization)
+  end
```
mz3944: In specs it says you have to be Admin only for deleting custom columns. So I assume you don't have to be one for editing.

app/controllers/custom_fields_controller.rb
```diff
@@ -24,20 +34,69 @@ def create
...
+
   private
 
+  def load_vars
+    @custom_field = CustomField.find_by_id(params[:id])
+    @custom_field = CustomField.find_by_id(
+      params[:custom_field_id]
+    ) unless @custom_field
+    render_404 unless @custom_field
+  end
```
mz3944: Maybe the code should be changed so that only one key is used for getting the custom field id. I assume in some places 'id' is used and in other 'custom_field_id' is used, so I suggest that everywhere only 'custom_field_id' is used.

app/assets/javascripts/samples/sample_datatable.js
Ducz0r: I don't know, we use both throughout application quite a lot. Since the functions don't do exactly the same, I'd leave it as it is.

app/helpers/permission_helper.rb
Ducz0r: I checked with Erik, and he agreed we should use the same permissions as with deleting.

PR #342, opened 2016-12-09
Generate png preview images for uploaded TIFF images

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-758

run make docker
run migrations 
to convert old images to the new style format run
Asset.find_each { |a| a.file.reprocess!(:medium) }

work done:
- removed direct upload
- disabled processing of nonimage files
- refactor assets.js and fix loading files that are processed by delay jobs
- added DelayedPaperclip gem
- convert every allowed image type in jpg format


app/models/asset.rb
```diff
@@ -19,18 +21,32 @@ class Asset < ActiveRecord::Base
   # Should be checked for any security leaks
   do_not_validate_attachment_file_type :file
 
-  before_file_post_process :is_image?
+  # before_file_post_process :is_image?
```
okriuchykhin: Commented out string

PR #344, opened 2016-12-09
Fix adding new samples columns [SCI-779]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-779


mlorb:
If you change the order of columns directly on table, in drop-down column order doesn't update.

mlorb:
The ordering works, but visibility changes after refresh.

okriuchykhin:
Fixed those issues, should be fine now

mlorb:
Great job!
Mergeaway :tada:

PR #349, opened 2016-12-13
Fix columns ordering when deleting custom fields [SCI-783]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-783


mlorb:
Now the bug with visibility appears again. The visibility shifts, maybe again the problem with indexes?

mlorb:
Now it works!!! Great job, merge away :tada: :+1:

PR #352, opened 2016-12-13
Hide dropdown when user interacts with samples table [SCI-793]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-793


ZmagoD:
Good job! :ship:

PR #363, opened 2016-12-19
Restyle step table display in reports [SCI-546]

Ducz0r:
Closes [SCI-546](https://biosistemika.atlassian.net/browse/SCI-546).


ZmagoD:
Good job! :ship

PR #365, opened 2016-12-20
Fix buttons disappear when deleting columns in samples [SCI-807]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-807


ZmagoD:
Good job! :ship:

PR #366, opened 2016-12-20
Samples - freeze header row and edit button... [SCI-284]

mlorb:
... and fixing all the bugs that appeared after adding new scrolling functionality:
- bug with adding new column
- bug with columns width in table header when editing or adding new sample
- bug with header not moving, when scrolling horizontally

https://biosistemika.atlassian.net/browse/SCI-284


ZmagoD:
Good job! :ship:

PR #367, opened 2016-12-20
Fix sample info popup [SCI-818]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-818


ZmagoD:
Good job! Merge away :ship:

PR #372, opened 2016-12-21
Fix flash message display in samples table [SCI-816]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-816


okriuchykhin:
Now solution uses module pattern

PR #376, opened 2016-12-23
Fix search on hidden columns in samples table [SCI-813]

Ducz0r:
Closes [SCI-813](https://biosistemika.atlassian.net/browse/SCI-813).

### DIFF
--- a/app/assets/javascripts/samples/sample_datatable.js.erb
+++ b/app/assets/javascripts/samples/sample_datatable.js.erb
@@ -1044,6 +1044,12 @@ function changeToEditMode() {
         column.visible(true);
         table.setColumnSearchable(column.index(), true);
       }
+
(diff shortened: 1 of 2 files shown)


mlorb:
Sorry, but I can't make it work... the search still displays rows with hidden sample items.

Ducz0r:
Works for me... @mz3944, can you review this?

mz3944:
Same her, doesn't work :/.

Ducz0r:
@mz3944, would it be acceptable if I just clear the searched text if you toggle columns visibility? Re-initializing table on each visibility toggle is costly, and has some weird consequences (shortly, it doesn't work as it should).

Ducz0r:
Nevermind, found an elegant solution.

PR #377, opened 2016-12-23
Fix 2 bugs with comments editing [SCI-820]

Ducz0r:
Fixed [SCI-820](https://biosistemika.atlassian.net/browse/SCI-820).

The introduction of this bug happened when we added the `auto_link` gem. Anyway, I've noticed that we're not rendering URLs for edited comments, so I also fixed that.


ZmagoD:
Good job! :ship:

PR #378, opened 2016-12-27
Add tooltips to samples column dropdown [SCI-802]

Ducz0r:
Closes [SCI-802](https://biosistemika.atlassian.net/browse/SCI-802).


mlorb:
Looks great, just one small comment. Wouldn't be better to use "Save" instead of "Update", from the consistency point of view? We use "Save", when we update sample types and groups.
Otherwise, ready for merge :+1:  :tada:

Ducz0r:
Probably. It's a general inconsistency throughout the software anyway...

PR #403, opened 2017-01-12
Year difference problems with notifications [SCI-876]

Ducz0r:
Fix an issue that occured with incorrect sorting of notifications, as well as all 4 types of comments (project, task, step, result), if you had activities/comments from 2 different years (e.g. 2016 and 2017).

Closes [SCI-876](https://biosistemika.atlassian.net/browse/SCI-876).

### DIFF
--- a/app/views/activities/_list.html.erb
+++ b/app/views/activities/_list.html.erb
@@ -1,7 +1,7 @@
 <%
   current_day = days_since_1970(DateTime.current)
 %>
-<% if !hide_today and activities.count > 0 and days_since_1970(activities[0].created_at) == current_day %>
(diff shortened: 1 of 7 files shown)


app/views/activities/_list.html.erb
```diff
@@ -1,18 +1,18 @@
 <%
-  current_day = DateTime.current.strftime('%j').to_i
+  current_day = days_since_1970(DateTime.current)
 %>
-<% if !hide_today and activities.count > 0 and activities[0].created_at.strftime('%j').to_i == current_day %>
+<% if !hide_today and activities.count > 0 and days_since_1970(activities[0].created_at) == current_day %>
```
ZmagoD: can use && instead of and?

PR #405, opened 2017-01-13
Visualization of User smart annotation after saving the field

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-832

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,4 +1,7 @@
 module ApplicationHelper
+  include ActionView::Helpers::AssetTagHelper
+  include ActionView::Helpers::UrlHelper
+
   def module_page?
     controller_name == 'my_modules'
   end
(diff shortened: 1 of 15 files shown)


ZmagoD:
probably also fixes https://biosistemika.atlassian.net/browse/SCI-891 bug please double check

app/helpers/application_helper.rb
```diff
@@ -116,9 +119,36 @@ def smart_annotation_parser(text)
...
-      if user
-        "<span>#{image_tag avatar_path(user, :icon_small)} " \
-        "#{user.full_name}</span>"
+      organization ||= current_organization
+      if user && organization
+        user_org = user
+                   .user_organizations
+                   .where('user_organizations.organization_id = ?',
+                          organization).first
+        user_description = '<div class=\'pull-left\'>'
```
Ducz0r: Just a styling suggestion, up to you - wouldn't it make sense to use `%Q{ ... }`, or `<<EOF ... EOF` here since it's a big block of escaped text?

PR #407, opened 2017-01-13
adds rake task for repocess file thumbnails [fixes SCI-831]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-831

### DIFF
--- a/lib/tasks/paperclip.rake
+++ b/lib/tasks/paperclip.rake
@@ -22,10 +22,26 @@ namespace :paperclip do
     end
   end
 
-  desc "List all paperclip files"
+  desc 'List all paperclip files'
   task files: :environment do
     print_model_files Asset, :file
     print_model_files User, :avatar
   end
 
+  desc 'Reprocess the Assets attachents styles'
+  task reprocess: :environment do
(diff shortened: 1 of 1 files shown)


lib/tasks/paperclip.rake
```diff
@@ -22,10 +22,26 @@ namespace :paperclip do
...
+      next unless asset.is_image?
+      begin
+        asset.file.reprocess! :medium, :large
+      rescue
+        error = true
+        $stderr.puts "exception while processing #{asset.file_file_name} " \
+                     "ID: #{asset.id}:"
+      end
+    end
+    $stderr.puts 'All gone well! ' \
```
Ducz0r: Hound is love, Hound is life :dog:.

ZmagoD:
@Ducz0r I think that reprocess removes the old thumbnails by default. I found a gem that seems interesting for us https://github.com/janfoeh/paperclip-optimizer

Ducz0r:
I don't think so, the old images remained where they were, at least when using `filesystem` storage. Don't know what happens on S3, but they probably stay there as well.

PR #412, opened 2017-01-16
Add sample info modal popup to sample smart annotations [SCI-818]

Ducz0r:
This MR also loads `datatables.js` globally (as it's required by
the modal), so I removed it everywhere in code, and it shouldn't
be used anymore (e.g. `//= require datatables`).

Closes [SCI-818](https://biosistemika.atlassian.net/browse/SCI-818).


Ducz0r:
Woops, referenced a wrong issue.

PR #414, opened 2017-01-16
Do not show archived projects, experiments and tasks in smart annotations [SCI-887]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-887

### DIFF
--- a/app/models/experiment.rb
+++ b/app/models/experiment.rb
@@ -70,11 +70,7 @@ class Experiment < ActiveRecord::Base
         .where('experiments.project_id IN (?)', projects_ids)
         .where_attributes_like([:name], a_query)
         .limit(Constants::ATWHO_SEARCH_LIMIT)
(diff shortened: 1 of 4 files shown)


app/models/experiment.rb
```diff
@@ -70,7 +70,11 @@ def self.search(
         .where('experiments.project_id IN (?)', projects_ids)
         .where_attributes_like([:name], a_query)
         .limit(Constants::ATWHO_SEARCH_LIMIT)
-      return new_query
+      if include_archived
```
Ducz0r: This seems like a job for ternary operator:
```ruby
return include_archived ? new_query : new_query.is_archived(false)
```

okriuchykhin: Yeah, but in such case it will not fit into line length limit))

okriuchykhin:
Done!

PR #424, opened 2017-01-18
Add “no results found” message in smart annotation [SCI-886]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-886


mz3944:
This PR was transfered to #428, because @okriuchykhin was out of the office, and the branch was not rebased, so I did it instead.

PR #437, opened 2017-01-23
Added smart annotations when adding new sample in samples table [SCI-913]

mz3944:
Fixes [SCI-913](https://biosistemika.atlassian.net/browse/SCI-913).


mz3944:
Take it!

PR #447, opened 2017-01-25
Fix table titles [SCI-941]

mlorb:
https://biosistemika.atlassian.net/browse/SCI-941


okriuchykhin:
All works! Merge away!

PR #454, opened 2017-01-26
Re-display images & checkboxes in PDF reports [SCI-953]

Ducz0r:
Closes [SCI-953](https://biosistemika.atlassian.net/browse/SCI-953).

### DIFF
--- a/app/views/reports/report.pdf.erb
+++ b/app/views/reports/report.pdf.erb
@@ -9,7 +9,8 @@
   </head>
   <body class="print-report-body">
     <div class="print-report">
-      <%= sanitize_input(@html) %>
+      <% # Also whitelist <img> and <input type="checkbox"> tags %>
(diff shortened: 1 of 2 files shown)


app/views/reports/report.pdf.erb
```diff
@@ -9,7 +9,8 @@
   </head>
   <body class="print-report-body">
     <div class="print-report">
-      <%= sanitize_input(@html) %>
+      <% # Also whitelist <img> and <input type="checkbox"> tags %>
+      <%= sanitize_input(@html, ['img', 'input'], ['type', 'disabled', 'checked']) %>
```
mlorb: Is there a reason, why you didn't whitelist the tags and attributes directly in constants.rb file?

Ducz0r: I think we don't allow `` tags in rich text fields, on which we also use `sanitize_input` method. I guess that's why.

PR #466, opened 2017-01-30
Make sciNote work for users without any teams [SCI_934]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-934

### DIFF
--- a/app/controllers/protocols_controller.rb
+++ b/app/controllers/protocols_controller.rb
@@ -860,10 +860,7 @@ class ProtocolsController < ApplicationController
   def check_view_all_permissions
     load_team_and_type
 
-    if @current_team &&
-       !can_view_organization_protocols(@current_organization)
-      render_403
-    end
+    render_403 unless can_view_team_protocols(@current_team)
   end
 
   def check_view_permissions
--- a/app/controllers/protocols_controller.rb
+++ b/app/controllers/protocols_controller.rb
@@ -860,7 +860,8 @@ class ProtocolsController < ApplicationController
   def check_view_all_permissions
     load_team_and_type
 
-    unless can_view_team_protocols(@current_team)
+    if @current_team &&
+       !can_view_organization_protocols(@current_organization)
       render_403
     end
   end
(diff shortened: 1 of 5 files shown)


app/controllers/protocols_controller.rb
```diff
@@ -860,7 +860,8 @@ def load_organization_and_type
   def check_view_all_permissions
     load_organization_and_type
 
-    unless can_view_organization_protocols(@current_organization)
+    if @current_organization &&
+       !can_view_organization_protocols(@current_organization)
```
Ducz0r: Is this correct? Shouldn't it be:
```ruby
unless is_member_of_any_organization &&
       @current_organization &&
       can_view_organization_protocols(@current_organization)
```

Ducz0r:
By the way, please don't start PRs with `Make sciNote...`, I instantly read `Make sciNote great again` 🤣.

app/controllers/protocols_controller.rb
okriuchykhin: If we check user's organizations in the view and disable protocols button on nav bar, then I think we don't need additional checks here and can keep original version which just throws 403 if user has no rights to view protocols:

unless can_view_organization_protocols(@current_organization)
  render_403
end

okriuchykhin: Or just in single line:

render_403 unless can_view_organization_protocols(@current_organization)

okriuchykhin: Isn't it enough to check presence of @current_organization variable in the view to enable/disable protocols button? Because even if variable present but contains wrong data it will be filtered out later in protocols controller. And by simpler checks in the view we'll reduce load as it is navbar and it's present in almost every view.

okriuchykhin:
I've rebased it with latest master which includes org renaming to teams

Ducz0r:
What about removing `current_team` when `user_team` association is destroyed? Didn't we talk about adding that to this PR as well?

okriuchykhin:
Yes, I made it, but then noticed that current_team_id is removed automatically when user leaves team, so there was no need in additional delete of this parameter.

Ducz0r:
Good to go, :ship: away!

PR #470, opened 2017-02-03
fixes invite user error [fixes SCI-987]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-987

here I have changed the variable name from @user_team to @user_t. Seems that the controller overrides that when the variable leaves check_user_team_permission method but I didn't found why is a very strange bug ....

### DIFF
--- a/app/controllers/users/settings_controller.rb
+++ b/app/controllers/users/settings_controller.rb
@@ -39,16 +39,16 @@ class Users::SettingsController < ApplicationController
   def update_preferences
     respond_to do |format|
       if @user.update(update_preferences_params)
-        flash[:notice] = t("users.settings.preferences.update_flash")
-        format.json {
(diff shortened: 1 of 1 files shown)


mlorb:
Fix also the billing (change user role is not working)...

ZmagoD:
http://gitlab.biosistemika.com/scinote/billing/merge_requests/59

mlorb:
I think the variable @user_team gets overridden in line 67 in settings_controller.rb. So the variable @user_team gets different value in method "team" (line 67), then in method "check_user_team_permission".
So actually you solved the problem and everything works great (at least here in scinote core). I am just wondering, if it would be maybe better to rename the variables in a different way (now you don't know what is the difference between @user_team and @user_t). What do you think about it? :)

ZmagoD:
becomes obsolete with @Ducz0r refactoring

PR #475, opened 2017-02-07
Fix usage of permissions [SCI-937]

mlorb:
https://biosistemika.atlassian.net/browse/SCI-937
Goes together with http://gitlab.biosistemika.com/scinote/organization_management/merge_requests/7


mlorb:
Not relevant any more...

PR #486, opened 2017-02-21
Fix user has no teams view [SCI-1033]

mlorb:
https://biosistemika.atlassian.net/browse/SCI-1033

Other part:
http://gitlab.biosistemika.com/scinote/organization_management/merge_requests/16

### DIFF
--- a/app/views/projects/index.html.erb
+++ b/app/views/projects/index.html.erb
@@ -99,7 +99,7 @@
   <div class="jumbotron">
   <h2><%=t 'projects.index.no_teams.title' %></h2>
   <p><%=t 'projects.index.no_teams.text' %></p>
-  <p>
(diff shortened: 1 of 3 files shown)


Ducz0r:
Sorry to be the devil's advocate here, but this should only be true in the `organization_management` addon, not in this codebase (if you're using sciNote core, you CAN create new teams on your own, therefore this text here makes sense). I suggest to open a PR in `organization_management`, and close this PR 🙁.

mlorb:
Ok, I messed it up. :(
Maybe the first line of text can be changed here on core anyway?

mlorb:
And I also added the needed data-hooks.

Ducz0r:
Very nice, :ship: away!

PR #506, opened 2017-03-03
Fix padding of state buttons [SCI-1073]

Ducz0r:
Hopefully now closes [SCI-1073](https://biosistemika.atlassian.net/browse/SCI-1073).


Ducz0r:
Nice find! Can you check it now?

Ducz0r:
Hmm, works for me 😮

PR #511, opened 2017-03-08
Refactoring of the comments using STI [SCI-1090]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-1090

### DIFF
--- a/app/controllers/my_module_comments_controller.rb
+++ b/app/controllers/my_module_comments_controller.rb
@@ -43,12 +43,14 @@ class MyModuleCommentsController < ApplicationController
   end
 
   def create
-    @comment = Comment.new(
+    @comment = TaskComment.new(
       message: comment_params[:message],
-      user: current_user)
+      user: current_user,
+      my_module: @my_module
+    )
 
     respond_to do |format|
-      if (@comment.valid? && @my_module.comments << @comment)
+      if @comment.valid? && @comment.save
         # Generate activity
         Activity.create(
           type_of: :add_comment_to_module,
@@ -176,12 +178,12 @@ class MyModuleCommentsController < ApplicationController
   end
 
   def check_edit_permissions
-    @comment = Comment.find_by_id(params[:id])
+    @comment = TaskComment.find_by_id(params[:id])
     render_403 unless @comment.present? && can_edit_module_comment(@comment)
   end
 
   def check_destroy_permissions
-    @comment = Comment.find_by_id(params[:id])
+    @comment = TaskComment.find_by_id(params[:id])
     render_403 unless @comment.present? && can_delete_module_comment(@comment)
   end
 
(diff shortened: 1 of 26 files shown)


app/controllers/my_module_comments_controller.rb
```diff
@@ -43,12 +43,14 @@ def index
...
+    @comment = TaskComment.new(
       message: comment_params[:message],
-      user: current_user)
+      user: current_user,
+      my_module: @my_module
+    )
 
     respond_to do |format|
-      if (@comment.valid? && @my_module.comments << @comment)
+      if @comment.valid? && @comment.save
```
Ducz0r: I think we can replace this with only `if @comment.save`.

app/controllers/project_comments_controller.rb
```diff
@@ -42,13 +42,14 @@ def index
...
       message: comment_params[:message],
-      user: current_user)
+      user: current_user,
+      project: @project
+    )
 
     respond_to do |format|
-
-      if (@comment.valid? && @project.comments << @comment)
+      if @comment.valid? && @comment.save
```
Ducz0r: I think we can replace this with only `if @comment.save`.

app/controllers/result_comments_controller.rb
```diff
@@ -40,12 +40,14 @@ def index
...
+    @comment = ResultComment.new(
       message: comment_params[:message],
-      user: current_user)
+      user: current_user,
+      result: @result
+    )
 
     respond_to do |format|
-      if (@comment.valid? && @result.comments << @comment)
+      if @comment.valid? && @comment.save
```
Ducz0r: I think we can replace this with only `if @comment.save`.

app/controllers/step_comments_controller.rb
```diff
@@ -39,13 +39,14 @@ def index
...
       message: comment_params[:message],
-      user: current_user)
+      user: current_user,
+      step: @step
+    )
 
     respond_to do |format|
-      if (@comment.valid? && @step.comments << @comment)
-
+      if @comment.valid? && @comment.save
```
Ducz0r: I think we can replace this with only `if @comment.save`.

app/models/step.rb
```diff
@@ -77,9 +75,8 @@ def self.search(user, include_archived, query = nil, page = 1)
   def destroy(current_user)
     @current_user = current_user
 
-    # Store IDs of comments, assets & tables so they
+    # Store IDs of assets & tables so they
     # can be destroyed in after_destroy
-    @c_ids = self.comments.collect { |c| c.id }
```
Ducz0r: Just checking, this code (and the deletion afterwards) is no longer needed?

okriuchykhin: Yeah, it gets deleted automatically with 'dependent: :destroy'

okriuchykhin:
Unneeded validation calls removed

PR #514, opened 2017-03-15
Add error messages to sample group/type edit/delete actions [SCI-1089]

mlorb:
Also fixes the error message when editing custom column on locked task (look the comment in jira task).
Goes together with: http://gitlab.biosistemika.com/scinote/electronic_signatures/merge_requests/18
Closes: https://biosistemika.atlassian.net/browse/SCI-1089

### DIFF
--- a/app/assets/javascripts/samples/sample_datatable.js.erb
+++ b/app/assets/javascripts/samples/sample_datatable.js.erb
@@ -967,6 +967,7 @@ function changeToEditMode() {
             '<th class="custom-field" id="' + data.id + '" ' +
             'data-editable data-deletable ' +
             'data-edit-url="' + data.edit_url + '" ' +
+            'data-update-url="' + data.update_url + '" ' +
             'data-destroy-html-url="' + data.destroy_html_url + '"' +
             '>' + generateColumnNameTooltip(data.name) + '</th>');
 
@@ -1050,6 +1051,7 @@ function changeToEditMode() {
           'data-position="' + colIndex + '" ' +
           'data-id="' + $(el).attr('id') + '" ' +
           'data-edit-url=' + $(el).attr('data-edit-url') + ' ' +
+          'data-update-url=' + $(el).attr('data-update-url') + ' ' +
           'data-destroy-html-url=' + $(el).attr('data-destroy-html-url') + ' ' +
           'class="' + visLi + '"' +
           '>' +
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/samples/sample_datatable.js.erb
```diff
@@ -1193,34 +1193,56 @@ function changeToEditMode() {
 
       cancelEditMode();
 
-      var self = $(this);
-      var li = self.closest('li');
-      var text = li.find('.text');
-      if ($(text).find('.modal-tooltiptext').length > 0) {
-        text = $(text).find('.modal-tooltiptext');
+      var li = $(this).closest('li');
+      var url = li.attr('data-edit-url') + "/electronic_sign_edit";
```
ZmagoD: please don't hard code urls. Here you have erb file so you can pass <%= link_to(whatever) %> in there

app/assets/javascripts/samples/sample_datatable.js.erb
```diff
@@ -1193,34 +1193,56 @@ function changeToEditMode() {
...
+
+      function ajaxCallEvent(){
+        $.ajax({
+          url: url,
+          success: function() {
+            var text = li.find('.text');
+            if ($(text).find('.modal-tooltiptext').length > 0) {
+              text = $(text).find('.modal-tooltiptext');
+            }
+            var textEdit = li.find('.text-edit');
```
ZmagoD: pro trick :smile:  In functions in JS you declare your variables at the top of the function you can do like this 
`var text, textEdit, controls, controlsEdit,.....;` 
and then assign some values to it. Read this: https://www.w3schools.com/js/js_hoisting.asp

app/controllers/custom_fields_controller.rb
```diff
@@ -1,7 +1,7 @@
 class CustomFieldsController < ApplicationController
   include InputSanitizeHelper
 
-  before_action :load_vars, only: [:update, :destroy, :destroy_html]
+  before_action :load_vars, except: :create
```
ZmagoD: reminder: make a double check if this class get decorated by some addon because if it is the variables will be loaded in this methods to and maybe we will face some strange behavior in the app like with team_user

mlorb: I checked. :+1:

PR #524, opened 2017-03-22
Add team_id field to assets and tables [SCI-1046]

okriuchykhin:
Adds team_id field to assets and tables, fills it for existing assets. Also adds code to set team for asset and table in the controllers. It is needed for auditing.


okriuchykhin:
Yeah, this team_id is now used only for auditing, but I will add them just to be consistent and in case somebody decide to work with tutorial projects

PR #525, opened 2017-03-24
Export large files - utility class

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-765


app/controllers/zip_exports_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  private
+
+  def load_var
+    @zip_export = ZipExport.find_by_id(params[:id])
+    redirect_to(file_expired_url, status: 301) and return unless @zip_export
+  end
+
+  def check_edit_permissions
+    render_403 unless @zip_export.user == current_user
```
mz3944: I know it's very basic, but this is how we did all other permissions. Check some of them out and you'll see, how basic they are, but we still put them in the Permissions helper class.

mz3944:
Ahoy, captain! Raise the sails and we embark!

PR #526, opened 2017-03-27
Add docker files for running in production mode [SCI-1096]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-1096

In order to start SciNote in production mode:

1) Set configuration variables in 'production.env';
 You can use 'openssl rand -hex 64' to generate SECRET_KEY_BASE and "openssl rand -base64 128 | tr -d '\n'" for PAPERCLIP_HASH_SECRET  
2) 'make docker-production'
3) 'make cli-production', then in container: 'rake db:create; rake db:migrate; rake db:seed' 
4) docker-compose -f ./docker-compose.production.yml up


Dockerfile.production
```diff
@@ -0,0 +1,31 @@
...
+ENV APP_HOME /usr/src/app
+RUN mkdir $APP_HOME
+WORKDIR $APP_HOME
+COPY . .
+RUN rm -f $APP_HOME/config/application.yml
+RUN rm -f $APP_HOME/production.env
+
+ENV RAILS_ENV production
+ENV DATABASE_URL=postgresql://postgres@db/scinote_production
+RUN bash -c "PAPERCLIP_HASH_SECRET=$(openssl rand -base64 128 | tr -d '\n') SECRET_KEY_BASE=$(openssl rand -hex 64) rake assets:precompile"
```
Ducz0r: If I understand this correctly, if we modify this Dockerfile before this line (e.g. add a `apt-get` dependency), that will require the users to re-run `make docker-production`. However, in that case, the 2 variables will be regenerated right? And user will lose his previous hash. I think that's unacceptable, or should at least be mentioned somewhere.

Or am I getting at this all wrong?

In any case, we could maybe add a `make` command to generate those 2 variables? Should be handier than copying the stuff written here/in Wiki.

Makefile
```diff
@@ -40,6 +46,9 @@ worker:
 cli:
```
Ducz0r: Shouldn't we also add `make` commands for other useful macros, e.g. `make console-production`/`make start-production`/`make stop-production`/...? Is there a better way to handle that? Maybe just [an argument](http://stackoverflow.com/questions/2826029/passing-additional-variables-from-command-line-to-make) provided to `Makefile`?

production.env
```diff
@@ -0,0 +1,21 @@
+SECRET_KEY_BASE=""
+PAPERCLIP_HASH_SECRET=""
+
+DATABASE_URL=postgresql://postgres@db/scinote_production
+
```
Ducz0r: So, if user wants these changed for production, he/she just changes this file? Should that really be in Git repository?

Ducz0r: I guess I would rather have a solution where a user needs to run a command to generate a "default" version of this file (which is `.gitignored`), and then update it according to his/her needs. But I don't know if that's easily achievable. 

From maintenance point of view, it would be better than always modifying a file that's tracked in Git.

Dockerfile.production
Ducz0r: Also, for production, I believe `DEFACE_ENABLED=true rake deface:precompile` should be run alongside `rake assets:precompile`.

okriuchykhin: Ok, if you change something in Dockerfile which changes an image, you need rebuild image in any case. As for this two variables, they are dummy and used only to generate assets because rails initialize all application while building assets and requires all these variables, DATABASE_URL is there by the same reason and contains default data.
It will be written in wiki that user needs to set these variables in production.env.

Makefile
okriuchykhin: I think it's better to use 'make' for development and building, and for for production mode use directly docker-compose. But there is no problem to add them to Makefile

production.env
okriuchykhin: I've put it as template, do we need to generate it with make?

okriuchykhin:
Actually there is no need to re-run make docker after building production images if you already built development images with make docker. But you might need to restart db container as it uses different volumes for db files.

Dockerfile.production
Ducz0r: Aah okay, gotcha. Then ignore my 1st comment.

production.env
Ducz0r: We decided **yes**, and also to generate the 2 secrets alongside.

okriuchykhin:
Added additional targets to Makefile and now you can generate initial config production.env with make. Also renamed docker images and containers so they don't interfere with development ones.
So procedure for fresh install now is:
make config-production;
make docker-production;
make database-production;
docker-compose -f ./docker-compose.production.yml up

Ducz0r:
Looks good! Oleksii, can you also update the Wiki once you merge this PR? So that the Wiki would include new/changed `Makefile` commands, as well as maybe a section explaining running server in production mode?

PR #532, opened 2017-03-31
Fixed so that file deletion callbacks are triggered

mz3944:
Fixed so that file deletion callbacks are triggered (and hence audit trails for this is generated). [fixes SCI-1150]

[SCI-1150](https://biosistemika.atlassian.net/browse/SCI-1150)

Also check: http://gitlab.biosistemika.com/scinote/audit_trails/merge_requests/30


mz3944:
Fixed. Check also: http://gitlab.biosistemika.com/scinote/audit_trails/merge_requests/30.

PR #542, opened 2017-04-06
Smart annotations and Notifications

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-825


app/controllers/concerns/steps_actions.rb
```diff
@@ -0,0 +1,155 @@
...
+      new_text: step.description,
+      title: t('notifications.step_description_title',
+               user: current_user.full_name,
+               step: step.name),
+      message: annotation_message(step)
+    )
+  end
+
+  def annotation_message(step)
+    return t('notifications.step_annotation_message_html',
```
ZmagoD: :banana:

app/helpers/application_helper.rb
```diff
@@ -58,6 +58,50 @@ def sample_types_page_expermient?
       !@experiment.nil?
   end
 
+  def smart_annotation_notification(options = {})
+    title = options.fetch(:title) { :title_must_be_present }
```
ZmagoD: if we get the field like title = options[:title] and you call this smart_annotation_notification function without passing the title in the function call the title will a nil value and later when you cast it into a string it will be an empty string. In the case you use fetch the function will blow up and you will get the :title_must_be_present error message. For the old_text in this case I don't use fetch because the param is optional. I only convert ti to an empty string if the param is nil so I can call gsub on it.

Ducz0r:
Okay, the fixes look good. What about the `fetch`, are we keeping it the way it is?

ZmagoD:
Yes it have the same effect if I throw an error or the error is triggered when .gsub is passed to the title/message variable it throw a message which is that symbol.

Ducz0r:
Alright then. Time to :ship:!

PR #557, opened 2017-04-13
Rename cloning text [SCI-1118]

mlorb:
[SCI-1118](https://biosistemika.atlassian.net/browse/SCI-1118)


Ducz0r:
Nice, :ship: away!

PR #564, opened 2017-04-18
Case insensitive email invitation

mz3944:
[SCI-856](https://biosistemika.atlassian.net/browse/SCI-856)

Also related: http://gitlab.biosistemika.com/scinote/organization_management/merge_requests/28


mz3944:
The problem was only when inviting users. When registering, the Devise already ignored case.

PR #572, opened 2017-04-25
make search for users case insensitive [fixes SCI-1192]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-1192


okriuchykhin:
As we already use Postgres specific case insensitive ILIKE in the rest of the search may be it is worth to use it also here to be consistent?

PR #574, opened 2017-04-25
prevent to get nil as a param in checklist_item_annotation method [fi…

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-1194


app/controllers/concerns/steps_actions.rb
```diff
@@ -74,8 +74,9 @@ def update_annotation_notifications(step,
       e.items.each do |ci|
         old_list = old_checklists.select { |i| i.id == e.id }.first
         old_item = old_list.items.select { |i| i.id == ci.id }.first if old_list
-
-        checklist_item_annotation(step, ci, old_item.text)
+        text = old_item.text if old_item
+        text ||= ''
```
mz3944: How about a one-liner?: `text = old_item ? old_item.text : ''`

PR #577, opened 2017-04-25
Change color of turbolinks progress bar [SCI-1029]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-1029


app/assets/stylesheets/themes/scinote.scss
```diff
@@ -1442,10 +1442,6 @@ table.dataTable {
   z-index: 1000000000;
 }
 
-html.turbolinks-progress-bar::before {
```
Ducz0r: Removing this causes the turbolinks progress bar's color to be `#0377ff`, which is not default sciNote theme. So please change it to `$color-theme-primary`. After that, merge away, it should work :).

PR #579, opened 2017-04-26
Tab Teams is highlighted when in Audit trails [fixes SCI-1214]

gazvoda:
https://biosistemika.atlassian.net/browse/SCI-1214?filter=-1

### DIFF
--- a/app/helpers/user_settings_helper.rb
+++ b/app/helpers/user_settings_helper.rb
@@ -13,7 +13,7 @@ module UserSettingsHelper
   end
 
   def on_settings_team_page?
-    controller_name == 'teams' &&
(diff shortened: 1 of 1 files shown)


app/helpers/user_settings_helper.rb
```diff
@@ -13,7 +13,7 @@ def on_settings_account_preferences_page?
   end
 
   def on_settings_team_page?
-    controller_name == 'teams' &&
-      action_name.in?(%w(index new create show))
+	  controller_name.in?(%w(teams audits)) &&
```
mz3944: I would leave out brackets, i.e.`....in? %w(teams audits)`.

PR #588, opened 2017-05-04
Add information how many samples are selected [SCI-191]

mlorb:
[SCI-191](https://biosistemika.atlassian.net/browse/SCI-191)


ZmagoD:
Good job @mlorb :+1: :rocket:

PR #592, opened 2017-05-05
removes TempFile after 7 days [fixes SCI-160]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-160

### DIFF
--- a/app/models/temp_file.rb
+++ b/app/models/temp_file.rb
@@ -4,11 +4,10 @@ class TempFile < ActiveRecord::Base
   has_attached_file :file
   do_not_validate_attachment_file_type :file
 
-  def destroy_obsolete_files
-    file = TempFile.find_by_id(self.id)
-    file.destroy! if file
+  def destroy_obsolete
+    destroy! if self
   end
 
-  handle_asynchronously :destroy_obsolete_files,
+  handle_asynchronously :destroy_obsolete,
                         run_at: proc { 7.days.from_now }
 end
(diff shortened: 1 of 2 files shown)


app/models/temp_file.rb
```diff
@@ -3,4 +3,12 @@ class TempFile < ActiveRecord::Base
 
   has_attached_file :file
   do_not_validate_attachment_file_type :file
+
+  def destroy_obsolete_files
+    file = TempFile.find_by_id(self.id)
+    file.destroy! if file
```
Ducz0r: Can't this be simply
```ruby
def destroy_obsolote
  destroy!
end
handle_asynchronously :destroy_obsolete,
                      run_at: proc { 7.days.from_now }
```
Or you need your queries because it's executed in delayed job?

ZmagoD: If the user complete the import of samples the temp file is deleted. Here I check if the TempFile is still present to prevent the call destroy on nil class because this job is triggered on every upload.

Ducz0r: Okay, then maybe
```ruby
def destroy_obsolote
  destroy! if self
end
handle_asynchronously :destroy_obsolete,
                      run_at: proc { 7.days.from_now }
```

Sounds like nitpicking, sorry, but I'd rename the method, and shorten it this way.

Ducz0r:
Sorry for nagging you about the code changes. You know what they say, 
> There are only two hard things in Computer Science: cache invalidation and naming things.
> - Phil Karlton

Merge away! :+1:

PR #593, opened 2017-05-05
Add advanced search options [SCI-1158]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-1158

### DIFF
--- a/app/assets/javascripts/search.js
+++ b/app/assets/javascripts/search.js
@@ -1,12 +1,16 @@
-$(document.body).ready(function() {
-  $('#search_whole_word').click(function() {
-    if ($(this).prop('checked') === true) {
-      $('#search_whole_phrase').prop('checked', false);
-    }
+(function() {
+  'use strict';
+
+  $(document.body).ready(function() {
+    $('#search_whole_word').click(function() {
+      if ($(this).prop('checked') === true) {
+        $('#search_whole_phrase').prop('checked', false);
+      }
+    });
+    $('#search_whole_phrase').click(function() {
+      if ($(this).prop('checked') === true) {
+        $('#search_whole_word').prop('checked', false);
+      }
+    });
   });
-  $('#search_whole_phrase').click(function() {
-    if ($(this).prop('checked') === true) {
-      $('#search_whole_word').prop('checked', false);
-    }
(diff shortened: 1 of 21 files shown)


app/assets/javascripts/search.js
```diff
@@ -0,0 +1,12 @@
+$(document.body).ready(function() {
```
ZmagoD: Can you put this in an IIFE? 
(function() {
  'use strict';
  ....
  ....
})();

app/controllers/search_controller.rb
```diff
@@ -41,47 +39,60 @@ def new
   private
 
   def load_vars
-    @search_query = params[:q] || ''
+    query = params[:q].strip || ''
```
ZmagoD: you can use 'fetch' to set the default value here if you want:

query = params.fetch(:q) { '' }

app/models/asset.rb
```diff
@@ -152,11 +179,12 @@ def self.search(
...
     Asset
-      .joins("LEFT JOIN asset_text_data ON assets.id = asset_text_data.asset_id")
-      .select("assets.*")
-      .select("ts_headline(data, to_tsquery('" + s_query + "'),
-        'StartSel=<mark>, StopSel=</mark>') headline")
-      .where("assets.id IN (?)",  ids)
+      .joins('LEFT JOIN asset_text_data ON ' \
+             ' assets.id = asset_text_data.asset_id')
+      .select('assets.*')
+      .select("ts_headline(data, to_tsquery('" + s_query +
```
ZmagoD: Can you use ActiveRecord::Sanitization for s_query variable here or sanitize it before you do the trim/map/join stuff above?

app/assets/javascripts/search.js
okriuchykhin: Done

app/models/asset.rb
okriuchykhin: done

app/controllers/search_controller.rb
okriuchykhin: done

ZmagoD:
Good job! :rocket:

PR #604, opened 2017-05-09
Added migration to remove updated_at from Activity [fixes SCI-1145]

gazvoda:
https://biosistemika.atlassian.net/browse/SCI-1145

### DIFF
--- a/app/utilities/first_time_data_generator.rb
+++ b/app/utilities/first_time_data_generator.rb
@@ -137,7 +137,8 @@ module FirstTimeDataGenerator
         user: user.full_name,
         project: project.name
       ),
-      created_at: project.created_at
+      created_at: project.created_at,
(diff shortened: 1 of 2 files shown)


okriuchykhin:
I think it is better to add updated_at parameter to Activity creation code in first_time_data_generator.rb. This error is caused due to use of .sneaky_save method which doesn't set automatically this field. You can just use same data as for created_at.

```
# Activity for creating archived module
    Activity.new(
      type_of: :create_module,
      user: user,
      project: project,
      my_module: archived_module,
      message: I18n.t(
        'activities.create_module',
        user: user.full_name,
        module: archived_module.name
      ),
      created_at: archived_module.created_at,
      updated_at: archived_module.created_at
    ).sneaky_save 
```

PR #606, opened 2017-05-09
Fix error messages display during file upload in steps and results [SCI-1235]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-1235

### DIFF
--- a/app/controllers/steps_controller.rb
+++ b/app/controllers/steps_controller.rb
@@ -97,11 +97,6 @@ class StepsController < ApplicationController
           status: :ok
         end
       else
-        # On error, delete the newly added files from S3, as they were
-        # uploaded on client-side (in case of client-side hacking of
-        # asset's signature response)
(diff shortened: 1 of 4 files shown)


app/controllers/steps_controller.rb
```diff
@@ -97,11 +97,6 @@ def create
           status: :ok
         end
       else
-        # On error, delete the newly added files from S3, as they were
-        # uploaded on client-side (in case of client-side hacking of
-        # asset's signature response)
-        Asset.destroy_all(new_assets)
```
Ducz0r: This was probably needed when we still had `PAPERCLIP_DIRECT_UPLOAD` option, right?

okriuchykhin: Probably, but anyway there is no 'new_assets' in the code, so it will not work in any case.

PR #607, opened 2017-05-10
Add specific classes to task card for analytics [SCI-1243]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-1243


Ducz0r:
So I have a question: does having a `class` solve out Jure's problems? Or would we need to also inject some `id`-s with module/task IDs? Just asking, I assume `class`-es are enough, but want to double-check.

If that is affirmative, merge away!

okriuchykhin:
He said that classes should be enough

PR #611, opened 2017-05-10
Fix root_url method unavailable in application_helper [SCI-1252]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-1252


ZmagoD:
Good job! :rocket:

PR #637, opened 2017-05-18
Add repositories dropdown to task navigation [SCI-1273]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-1273

### DIFF
--- a/app/views/shared/_secondary_navigation.html.erb
+++ b/app/views/shared/_secondary_navigation.html.erb
@@ -188,7 +188,8 @@
               </a>
             </li>
           <% end %>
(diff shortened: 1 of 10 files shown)


app/views/shared/_secondary_navigation.html.erb
```diff
@@ -188,6 +188,27 @@
               </a>
             </li>
           <% end %>
+          <% if can_view_repositories(@my_module.experiment.project.team) %>
```
mz3944: I would add `@my_module.experiment.project.team.repositories.exists?`, so when there are no custom repositories, the dropdown isn't shown, because it's empty in this case.

PR #643, opened 2017-05-26
Handled user removal from the team for user smart annotation [fixes SCI-1305]

mz3944:
[SCI-1305](https://biosistemika.atlassian.net/browse/SCI-1305)


mz3944:
Darnit, again, forgot to commit the locales file... I never learn :/.

PR #651, opened 2017-06-01
Add custom repository table [SCI-1274]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-1274
https://biosistemika.atlassian.net/browse/SCI-1290

### DIFF
--- a/app/datatables/repository_datatable.rb
+++ b/app/datatables/repository_datatable.rb
@@ -1,0 +1,346 @@
+require 'active_record'
+
+class RepositoryDatatable < AjaxDatatablesRails::Base
+  include ActionView::Helpers::TextHelper
+  include SamplesHelper
+  include InputSanitizeHelper
+  include Rails.application.routes.url_helpers
+  include ActionView::Helpers::UrlHelper
+  include ApplicationHelper
+  include ActiveRecord::Sanitization::ClassMethods
+
+  ASSIGNED_SORT_COL = 'assigned'.freeze
+
+  REPOSITORY_TABLE_DEFAULT_STATE = {
+    'time' => 0,
+    'start' => 0,
+    'length' => 5,
+    'order' => [[2, 'desc']],
+    'search' => { 'search' => '',
+                  'smart' => true,
+                  'regex' => false,
+                  'caseInsensitive' => true },
+    'columns' => [],
+    'ColReorder' => [*0..4]
+  }
+  5.times do
+    REPOSITORY_TABLE_DEFAULT_STATE['columns'] << {
+      'visible' => true,
+      'search' => { 'search' => '',
+                    'smart' => true,
+                    'regex' => false,
+                    'caseInsensitive' => true }
+    }
+  end
+  REPOSITORY_TABLE_DEFAULT_STATE.freeze
+
+  def initialize(view,
+                 repository,
+                 my_module = nil,
+                 user = nil)
+    super(view)
+    @repository = repository
+    @team = repository.team
+    @my_module = my_module
+    @user = user
+  end
+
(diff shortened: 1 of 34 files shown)


app/models/repository_rows_my_module.rb
```diff
@@ -0,0 +1,7 @@
+class RepositoryRowsMyModule < ActiveRecord::Base
```
Ducz0r: I think this class needs to be in singularity, and maybe use some additional checks (see how I've done it in my PR) - https://github.com/Ducz0r/scinote-web/blob/lm-sci-1279/app/models/repository_row_my_module.rb

okriuchykhin: I think Rails has special rules for naming join models, which represents many to many association and I think both sides of association should be plural

Ducz0r: I don't know. Maybe we messed up with the rest of the join tables, which are all named e.g. `user_my_modules`, but that translates into ActiveRecord being named `UserMyModule`, which I believe is a good way to name a join model. I mean, I think it's more consistent to keep this table named `repository_row_my_modules`, which will then also mean that the ActiveRecord is named `RepositoryRowMyModule` (with no plural there).

okriuchykhin: https://stackoverflow.com/questions/11590469/rails-naming-convention-for-join-table

app/datatables/repository_datatable.rb
```diff
@@ -0,0 +1,346 @@
...
+                        # repository_cells: :value
+                      ).references(
+                        :repository_columns,
+                        :created_by
+                      )
+                      .where(repository: @repository)
+    if @my_module
+      @assigned_rows = @my_module.repository_rows
+
+      # repository_rows.joins(
```
Ducz0r: Should this be thrown out?

app/datatables/repository_datatable.rb
```diff
@@ -0,0 +1,346 @@
...
+    end
+  end
+
+  def new_sort_column(item)
+    coli = item[:column].to_i - 1
+    model, column = sortable_columns[sortable_displayed_columns[coli].to_i]
+                    .split('.')
+
+    return model if model == ASSIGNED_SORT_COL
+    col = [model.constantize.table_name, column].join('.')
```
Ducz0r: Indeed, is this meant as `coli`? Or should both names be `col`. Is this function used at all?

app/models/my_modules_repository_row.rb
```diff
@@ -0,0 +1,8 @@
+class MyModulesRepositoryRow < ActiveRecord::Base
```
Ducz0r: I still think you should change this to `MyModuleRepositoryRow`, just to be consistent with the rest of the codebase, no other reason.

app/models/repository_table.rb
```diff
@@ -0,0 +1,60 @@
+class RepositoryTable < ActiveRecord::Base
```
Ducz0r: Should we rename this to `RepositoryTableState`? I mean, just to point out specifically that this is a table just to store state data.

okriuchykhin: Renamed. I've just made the same name as we use for samples

app/models/my_modules_repository_row.rb
okriuchykhin: Renamed

app/datatables/repository_datatable.rb
okriuchykhin: Removed

Ducz0r:
Aaalright, start up the engines and merge away! May God have mercy on our souls :pray: :rainbow:.

PR #654, opened 2017-06-05
Fix bug with protocol preview modal not working in protocol repo [SCI-1323]

Ducz0r:
Closes [SCI-1323](https://biosistemika.atlassian.net/browse/SCI-1323).

### DIFF
--- a/app/assets/javascripts/protocols/index.js
+++ b/app/assets/javascripts/protocols/index.js
@@ -209,9 +209,9 @@ function initProtocolPreviewModal() {
           modalTitle.html(data.title);
           modalBody.html(data.html);
           modalFooter.html(data.footer);
-          initHandsOnTable(modalBody);
+          window.initHandsOnTable(modalBody);
           modal.modal("show");
-          initHandsOnTable(modalBody);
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/protocols/steps.js.erb
```diff
@@ -673,4 +673,5 @@
     });
   })
 
+  window.initHandsOnTable = initHandsOnTable;
```
ZmagoD: You can also use global to bind the function to the window object 😄  the window is passed as a param in the IIFE.

app/assets/javascripts/protocols/index.js
```diff
@@ -209,9 +209,9 @@ function initProtocolPreviewModal() {
           modalTitle.html(data.title);
           modalBody.html(data.html);
           modalFooter.html(data.footer);
-          initHandsOnTable(modalBody);
+          window.initHandsOnTable(modalBody);
```
ZmagoD: You can just call initHandsOnTable(modalBody) since the problem was that the method hasn't been exposed globaly.

PR #655, opened 2017-06-05
Fix bug with TinyMCE not working for steps with nil description [SCI-1321]

Ducz0r:
Closes [SCI-1321](https://biosistemika.atlassian.net/browse/SCI-1321).

### DIFF
--- a/app/helpers/tiny_mce_helper.rb
+++ b/app/helpers/tiny_mce_helper.rb
@@ -18,7 +18,7 @@ module TinyMceHelper
   end
 
   def generate_image_tag_from_token(text)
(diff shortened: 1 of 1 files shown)


app/helpers/tiny_mce_helper.rb
```diff
@@ -18,6 +18,7 @@ def parse_tiny_mce_asset_to_token(text, ref = nil)
   end
 
   def generate_image_tag_from_token(text)
+    return nil if text.nil?
```
ZmagoD: Oh you can do better 😄  
`return unless text`

PR #659, opened 2017-06-07
Include repositories in reports [SCI-1279]

Ducz0r:
Also fix some very minor bugs, and remove `billing` stuff from `db/schema.rb`, which shouldn't even be there in the first place.

Closes [SCI-1279](https://biosistemika.atlassian.net/browse/SCI-1279).


config/initializers/extends.rb
```diff
@@ -35,7 +35,8 @@ class Extends
                            result_comments: 12,
                            project_activity: 13, # TODO
                            project_samples: 14, # TODO
-                           experiment: 15 }
+                           experiment: 15,
+                           my_module_repository: 16 }
```
ZmagoD: Please check this how affect the addons

Ducz0r: Thanks, nice catch! Damn, I hate that this is implemented this way... We should make it a "string" enum somehow, this is gonna cause us problems in the long run.

PR #669, opened 2017-06-09
Fix `Gemfile.lock`

paulmenzel:


mz3944:
This doesn't seem to be needed. If anything, when updating `rdoc` the only thing that gets updated is its' version: `rdoc (4.3.0)`.

paulmenzel:
@mz3944, I am using Bundler 1.15.1. What version do you use? Running `bundle install` the file `Gemfile.lock` is changed as written here.

mz3944:
We're using `Bundler 1.14.6`, just as it says at the end of `Gemfile.lock file`. I've updated it to Bundler `1.15.1` but it still didn't produce your changes... It's not needed, so I guess we can close this issue.

PR #685, opened 2017-06-14
Fix bug with audit trail - ajax error [SCI-1373]

mlorb:
[SCI-1373](https://biosistemika.atlassian.net/browse/SCI-1373)


ZmagoD:
Good job! :rocket:

PR #687, opened 2017-06-14
Repository management - filter on custom columns [SCI-1357]

mz3944:
[SCI-1357](https://biosistemika.atlassian.net/browse/SCI-1357)

### DIFF
--- a/app/datatables/repository_datatable.rb
+++ b/app/datatables/repository_datatable.rb
@@ -185,18 +185,20 @@ class RepositoryDatatable < AjaxDatatablesRails::Base
   end
 
   # Overriden to make it work for custom columns, because they are polymorphic
-  def simple_search(records)
-    return records unless params[:search].present? &&
-                          params[:search][:value].present?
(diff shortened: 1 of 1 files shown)


mz3944:
Well, I also have the raw SQL version. If you think it would be better to use this, let me know and I'll replace the `select` with this:

`RepositoryRow.find_by_sql(
      "SELECT DISTINCT repository_rows.*
      FROM repository_rows
      LEFT OUTER JOIN (SELECT repository_cells.*,
        repository_text_values.data AS value FROM repository_cells
      INNER JOIN repository_text_values
      ON repository_text_values.id = repository_cells.value_id) AS values
      ON values.repository_row_id = repository_rows.id
      WHERE repository_rows.repository_id = #{@repository.id} AND (repository_rows.name Like '%#{params[:search][:value]}%' OR value Like '%#{params[:search][:value]}%')"
    )`

Ducz0r:
Well I'd prefer the raw SQL (if it works), but this query doesn't filter by dates right?

PR #695, opened 2017-06-19
Retain the same word case in repositories dropdown [SCI-1353]

mlorb:
[SCI-1353](https://biosistemika.atlassian.net/browse/SCI-1353)

### DIFF
--- a/app/assets/stylesheets/repositories.scss
+++ b/app/assets/stylesheets/repositories.scss
@@ -2,6 +2,7 @@
   height: auto;
   max-height: 400px;
   overflow-x: hidden;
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/repositories.scss
```diff
@@ -2,6 +2,7 @@
   height: auto;
   max-height: 400px;
   overflow-x: hidden;
+  text-transform: initial;
```
ZmagoD: Težka :banana: 
Could we do more nicer here or is this dropdown text transformed with css?

PR #697, opened 2017-06-19
Fix total number of records in task repository [SCI-1381]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-1381

### DIFF
--- a/app/datatables/repository_datatable.rb
+++ b/app/datatables/repository_datatable.rb
@@ -149,6 +149,14 @@ class RepositoryDatatable < AjaxDatatablesRails::Base
                       .joins(:created_by)
                       .where(repository: @repository)
 
+    # Make mappings of custom columns, so we have same id for every column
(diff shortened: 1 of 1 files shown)


app/datatables/repository_datatable.rb
```diff
@@ -158,6 +158,7 @@ def get_raw_records
                                  )
                                  .joins(:created_by)
                                  .where(repository: @repository)
+      return @assigned_rows if params[:assigned] == 'assigned'
```
mz3944: I think the condition should be `@my_module && params[:assigned] == 'assigned'`, just like on line 185, and you can remove line 185, since you do this here.

mz3944: Oh right, it's not needed anymore, since we also show assigned column in repositories...

mz3944:
Megre away!!

PR #723, opened 2017-06-29
Fixes bugs on canvas [fixes SCI-1408]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-1408

### DIFF
--- a/app/models/experiment.rb
+++ b/app/models/experiment.rb
@@ -761,8 +761,6 @@ class Experiment < ApplicationRecord
     workflows = []
     dg.to_undirected.each_connected_component { |w| workflows <<  w }
 
-    # Retrieve maximum allowed module group name
-    max_length = Constants::NAME_MAX_LENGTH
     # For each workflow, generate new names
     new_index = 1
     wf_names = []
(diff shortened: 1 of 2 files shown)


app/models/experiment.rb
```diff
@@ -762,7 +762,7 @@ def update_module_groups(module_groups, current_user)
     dg.to_undirected.each_connected_component { |w| workflows <<  w }
 
     # Retrieve maximum allowed module group name
-    max_length = (MyModuleGroup.validators_on(:name).select { |v| v.class == ActiveModel::Validations::LengthValidator }).first.options[:maximum]
```
mz3944: Like the ingenuity of the original solution. Don't know what it does, don't know how it does it, and definitely don't know why it does it like so, but it looks pretty awesome!

app/models/experiment.rb
```diff
@@ -762,7 +762,7 @@ def update_module_groups(module_groups, current_user)
...
+    max_length = Constants::NAME_MAX_LENGTH
```
mz3944: It would be nicer to remove this line and use `Constants::NAME_MAX_LENGTH` directly instead, just as it is used on line `367`.

mz3944:
Merge away!

PR #724, opened 2017-06-29
Fix Repeat Tutorial Action / Tutorial

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-1410


app/models/my_module.rb
```diff
@@ -285,7 +285,7 @@ def get_downstream_modules
       if !final.include?(my_module)
         final << my_module
       end
-      modules.push(*my_module.my_modules.flatten)
+      modules.push(*my_module.my_modules)
```
mz3944: You've already fixed that in another branch, but OK.

PR #733, opened 2017-07-06
setup testing tools [fixes SCI-1400]

ZmagoD:
https://biosistemika.atlassian.net/secure/RapidBoard.jspa?rapidView=1&view=detail&selectedIssue=SCI-1400

to run feature tests open the cli and type cucumber
for unit type rspec
when you run the tests you can check the app coverage open the file coverage/index.html


spec/models/asset_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    it { should have_db_column :file_updated_at }
+    it { should have_db_column :created_by_id }
+    it { should have_db_column :last_modified_by_id }
+    it { should have_db_column :estimated_size }
+    it { should have_db_column :file_present }
+    it { should have_db_column :lock }
+    it { should have_db_column :lock_ttl }
+    it { should have_db_column :version }
+    it { should have_db_column :file_processing }
+    it { should have_db_column :team_id }
```
programmertoni: I mostly used shoulda matchers for associations and validations.

It is good to have expectations on columns but the down side is that the tests are fragile. If you change column name you need to always change tests.

programmertoni:
It looks Good

PR #752, opened 2017-07-26
fix background worker bug

ZmagoD:


Ducz0r:
Seems to be working, :ship: away!

PR #758, opened 2017-07-28
Td sci 1492

programmertoni:


app/views/layouts/application_react.html.erb
```diff
@@ -0,0 +1,21 @@
+
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta data-hook="head-js">
+    <title>React App</title>
```
ZmagoD: still can you remove this

app/views/layouts/application_react.html.erb
```diff
@@ -0,0 +1,21 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no">
+    <%= stylesheet_link_tag 'application', media: 'all', 'data-turbolinks-track' => true %>
```
ZmagoD: we said that we will use styled_components so probably we don't need all the mess be loaded here :smile 
just load bootstrap css

config/routes.rb
```diff
@@ -8,6 +8,9 @@
 
     root 'projects#index'
 
+    # React Frontend-Framework
+    get '/react/settings', to: 'react/settings#index'
```
ZmagoD: hmm probably we should name this somehow different like jsAPI or clentAPI or whatever

app/controllers/react/settings_controller.rb
```diff
@@ -0,0 +1,7 @@
+module React
```
ZmagoD: Also here I will name this stuff something different than react maybe Client API or jsAPI ....

ZmagoD:
Can you also switch the target branch to decoupling-settings-page

PR #759, opened 2017-07-28
fixes signup spec [fixes SCI-1471]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-1471


ZmagoD:
No it doesn't because we don't have the controller but looks like devise relies that the /users path exist which make no sense :smile:

PR #767, opened 2017-08-08
Add Omniauth to Devise for external authentication [SCI-1509]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-1509

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -210,7 +210,7 @@ class User < ApplicationRecord
         auth.uid
       )
       .references(:user_identities)
-      .first
+      .take
   end
 
   # Search all active users for username & email. Can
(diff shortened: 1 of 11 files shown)


app/models/user.rb
```diff
@@ -200,6 +202,17 @@ def name=(name)
...
 
+  def self.from_omniauth(auth)
+    includes(:user_identities)
+      .where(
+        'user_identities.provider=? AND user_identities.uid=?',
+        auth.provider,
+        auth.uid
+      )
+      .references(:user_identities)
+      .first
```
Ducz0r: Very *nitpicky*, but would `.take` here work? Because it's faster than `.first`.

PR #769, opened 2017-08-10
Refactor user settings into a jsonb field [SCI-1475]

Ducz0r:
This PR changes the user settings, so they're all contained in a single `jsonb` column in `users` table in database.

They're also extensible, which means that addons can simply put additional stuff into it. To provide default setting values for new stuff, the following helper method was implemented:
```ruby
User.class_eval do
  default_settings(
    default_key: true,
    default_hash: {
      default_key_two: 'Value',
      default_key_three: 1250,
      'Key Four' => false
    }
  )
end
```

Also, a `SettingsModel` concern has been created, so if we ever need to reuse this functionality, we should simply generate a migration creating the new `jsonb` field (with name `:settings`), and then specifying `include SettingsModel` on the model.

It should be pretty easy to add additional settings to the user this way, and not always require a database migration for it.


Ducz0r:
Waow, nice find o.0. Removing the file from index...

db/migrate/20170809131000_refactor_user_settings.rb
```diff
@@ -0,0 +1,53 @@
+class RefactorUserSettings < ActiveRecord::Migration[5.1]
+  def up
+    add_column :users, :settings, :jsonb, default: {}, null: false
```
programmertoni: This (jsonb) is really nice :+1:

PR #773, opened 2017-08-21
Add basic datatables & datagrid [SCI-1496]

Ducz0r:
This PR includes the basic functionality of 2 data table libraries we decided to go with for sciNote:
* [react-bootstrap-table](https://github.com/AllenFang/react-bootstrap-table) - refactored into `<DataTable />`,
* [react-data-grid](https://github.com/adazzle/react-data-grid) - refactored into `<DataGrid />`.

The new, composed components can basically take the corresponding original component parameters, as well as some of our own. A simple use case would be the following:
```js
import DataTable from "../../shared/data_table/DataTable";
import DataGrid from "../../shared/data_grid/DataGrid";

const data = [
  {
    id: 1,
    team: 'Team 1',
    role: 'Member',
    members: 5
  },
  {
    id: 2,
    team: 'Team 2',
    role: 'Administrator',
    members: 12
  },
  {
    id: 3,
    team: 'Team 3',
    role: 'Administrator',
    members: 13
  },
  {
    id: 4,
    team: 'Team 4',
    role: 'Member',
    members: 2
  },
  {
    id: 5,
    team: 'Team 5',
    role: 'Member',
    members: 1
  }
];
const columns = [
    {
    id: 1,
    isKey: true,
    textId: 'id',
    name: 'ID',
    position: 0,
    visible: false,
    sortable: true,
    locked: false
  },
  {
    id: 2,
    isKey: false,
    textId: 'team',
    name: 'Team',
    position: 1,
    visible: true,
    sortable: true,
    locked: false
  },
  {
    id: 3,
    isKey: false,
    textId: 'role',
    name: 'Role',
    position: 2,
    visible: true,
    sortable: true,
    locked: false
  },
  {
    id: 4,
    isKey: false,
    textId: 'members',
    name: 'Members',
    position: 3,
    visible: true,
    sortable: true,
    locked: false
  }
];

...

<DataTable data={data} columns={columns} />
<DataGrid data={data} columns={columns} />
```

Closes [SCI-1496](https://biosistemika.atlassian.net/browse/SCI-1496).


app/javascript/packs/styles/main.scss
```diff
@@ -1,8 +1,70 @@
...
+@import '~react-bootstrap-table/dist/react-bootstrap-table.min';
 
 body {
   background-color: $color-concrete;
   color: $color-emperor;
   font-family: "Open Sans",Arial,Helvetica,sans-serif;
   font-size: 13px;
 }
+
+.react-bs-table {
```
ZmagoD: for styling please use styled-components library in here we will put global styles like default font sizes and bootstrap overrides
https://www.styled-components.com/

app/javascript/packs/styles/main.scss
```diff
@@ -1,8 +1,70 @@
...
+  thead {
+    background-color: #909088;
```
ZmagoD: please use constants.scss if you style in components with styled-components use app/constants/colors

package.json
```diff
@@ -76,6 +76,8 @@
     "styled-components": "^2.1.1",
     "webpack": "^3.2.0",
     "webpack-manifest-plugin": "^1.1.2",
-    "webpack-merge": "^4.1.0"
+    "webpack-merge": "^4.1.0",
+    "react-bootstrap-table": "^4.0.0",
+    "react-data-grid": "^2.0.2"
```
ZmagoD: yarn.lock is missing!

app/javascript/packs/shared/data_table/DataTable.jsx
```diff
@@ -0,0 +1,69 @@
...
+
+  cleanProps() {
+    // Remove additional props from the props value
+    const {columns, ...cleanProps} = this.props;
+    return cleanProps;
+  }
+
+
+  displayHeader() {
+    const orderedCols = [...this.props.columns].sort((a, b) => a.position - b.position);
```
ZmagoD: You don't need the spread operator here https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Spread_operator

here is babel repl you can use it to see how the es6 code is converted to es5 https://babeljs.io/repl/

app/javascript/packs/shared/data_table/DataTable.jsx
```diff
@@ -0,0 +1,69 @@
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import { BootstrapTable, TableHeaderColumn } from 'react-bootstrap-table';
+
+class DataTable extends Component {
+  static cleanColumnAttributes(col) {
+    // Remove additional attributes from the columns
+    const {
```
ZmagoD: why just don't extract the cleanCol for col?

const { cleanCol } = col

app/javascript/packs/shared/data_grid/DataGrid.jsx
```diff
@@ -0,0 +1,118 @@
...
+    this.cleanProps = this.cleanProps.bind(this);
+    this.transformColumns = this.transformColumns.bind(this);
+    this.setupDefaultProps = this.setupDefaultProps.bind(this);
+
+    this.transformColumns();
+    this.setupDefaultProps();
+
+    // Store the original rows array, and make a copy that
+    // can be used for modifying eg.filtering, sorting
+    this.state = {
```
Ducz0r: I don't understand first comment, it's already below `super(props)`. The second comment - I've just used the way the `react-data-grid` examples do it. If you have a better suggestion on how to do it, I'm all up for that.

app/javascript/packs/shared/data_grid/DataGrid.jsx
```diff
@@ -0,0 +1,118 @@
+import React, { Component } from "react";
+import PropTypes from "prop-types";
+import ReactDataGrid from 'react-data-grid';
+
+class DataGrid extends Component {
```
programmertoni: Maybe you can extend ReactDataGrid instead of Component

app/javascript/packs/shared/data_grid/DataGrid.jsx
```diff
@@ -0,0 +1,118 @@
...
+    // can be used for modifying eg.filtering, sorting
+    this.state = {
+      originalRows: this.props.data,
+      rows: this.props.data.slice(0)
+    };
+  }
+
+  setupDefaultProps() {
+    // Setup the default props if they're not provided
+    if ('rowGetter' in this.props) {
```
Ducz0r: The idea I had in mind for those 2 components (`DataGrid`, `DataTable`) was that you can provide some of our own properties that are defined on our end (e.g. `columns` - ideally those props should be as similar as possible for both components, so we can use them interchangably as much as possible), however you can pass-through all the parameters for original components (so, all the properties for `BootstrapTable` and `ReactDataGrid`). And there are many props for those 2 original components.

If I understand React/JS correctly, doing it your way (`let { rowGetter, rowsCount, onGridSort } = this.props`) would throw out all other original component props.

Maybe a wider question for you: do you think the approach I'm defending in this comment is even viable? Or should we do it some other way? I was just trying to make those components semi-reusable - but still customizable by using original component props - and allegedly, [component composition](https://facebook.github.io/react/docs/composition-vs-inheritance.html) is the way to go with React.js.

package.json
Ducz0r: Okay!

app/javascript/packs/styles/main.scss
Ducz0r: Roger!

app/javascript/packs/styles/main.scss
Ducz0r: Okay

app/javascript/packs/shared/data_table/DataTable.jsx
Ducz0r: I want to create a hash/map without our custom keys&values (e.g. props). So we can pass this object safely into `<TableHeaderColumn>`, without worrying it will break something. I think this is the simplest way to do it (and, allegedly, cloning `col` and then calling e.g. `delete col.id` is considered slow).

app/javascript/packs/shared/data_grid/DataGrid.jsx
Ducz0r: I was thinking about inheritance, but in React.js it's way [more powerful](https://facebook.github.io/react/docs/composition-vs-inheritance.html) to use composition instead of inheritance.

programmertoni: :+1:

app/javascript/packs/shared/data_grid/DataGrid.jsx
```diff
@@ -0,0 +1,118 @@
...
+    // Remove additional props from the props value
+    const {
+      columns, rowGetter, rowsCount, ...cleanProps
+    } = this.props;
+    return cleanProps;
+  }
+
+  render() {
+    return (
+      <ReactDataGrid
```
programmertoni: It is probably better to return custom <DataGrid> component so that anyone using React chrome extention will know that this is modified <ReactDataGrid>

Ducz0r:
@ZmagoD I've went through your comments, and fixed some stuff you mentioned. I responded feedback on some of them, but you will need to check commit nr. 2 to view them. Can you revisit them please? Also if there's something not worth writing and would be solved easier by talking, just nag me 🙂.

app/javascript/packs/shared/data_grid/index.jsx
```diff
@@ -85,7 +116,7 @@ class DataGrid extends Component {
 
   render() {
     return (
-      <ReactDataGrid
+      <StyledReactDataGrid
```
ZmagoD: :+1:

PR #783, opened 2017-09-01
Adds .feature files for Cucumber specifications

programmertoni:


test/specifications/gamification.feature
```diff
@@ -0,0 +1,72 @@
+# feature/gamification. feature
```
ZmagoD: This feature don't belong here!
move it in the gamification addon 
https://corlewsolutions.com/articles/article-127-how-to-setup-cucumber-in-a-rails-engine

programmertoni: done

ZmagoD:
Looks better ☺ it will be also Ok if you will put in separate dir the task / experiment / user authentication and so on related features

PR #803, opened 2017-09-28
Add global flash notifications/alerts [SCI-1599]

Ducz0r:
The difference from previous (Rails) implementation is that they're now floating, and you can stack multiple of them. By default, they timeout after 5 seconds.

Generation of flash messages is done using Redux 😎. To generate a flash message from within a component of your choice, you need to perform the following in the component:

```diff
+import { connect } from "react-redux";
+import { addAlert } from "<REL_ROOT_PATH>/components/actions/AlertsActions";

-const MyComponent = ({ my_props }) =>
+const MyComponent = ({ my_props, addMyCustomAlert }) =>
   <Wrapper>
     ...
+    <Button onClick={() => addMyCustomAlert()}>Generate alert</Button>
     ...
   </Wrapper>

 MyComponent.propTypes = {
   ...
+  addMyCustomAlert: PropTypes.func.isRequired,
   ...
 }

 MyComponent.defaultProps = {
   ...
+  addMyCustomAlert: undefined,
   ...
 }

+const mapDispatchToProps = dispatch => ({
+  addMyCustomAlert() {
+    dispatch(addAlert("My custom alert text", "error"));
+  }
+});

-export default MyComponent;
+export default connect(undefined, mapDispatchToProps)(MyComponent);
```

To be even more specific, the following 3 actions are available in `AlertsActions` for alerts manipulation:

```javascript
addAlert(message,
         type,
         id = shortid.generate(),
         timeout = 5000));
clearAlert(id);
clearAllAlerts();
```

Oh, and p.s. - I've spent a lot of time getting animations to work using `styled-components` library, but it simply doesn't go! You can make it work for entrance animations, but for more complex animations where `react-transition-group` is required, there's simply no other way than to specify animation CSS code globally in `.scss` file/s.

Closes [SCI-1599](https://biosistemika.atlassian.net/browse/SCI-1599).


app/javascript/src/styles/animations.scss
```diff
@@ -0,0 +1,27 @@
+.alert-animated-enter {
+  opacity: 0.01;
```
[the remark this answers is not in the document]
ZmagoD: fix hound

app/javascript/src/components/AlertsContainer/index.jsx
```diff
@@ -0,0 +1,71 @@
...
+        </TransitionGroup>
+      </Wrapper>
+    );
+  }
+}
+
+AlertsContainer.propTypes = {
+  alerts: PropTypes.arrayOf(
+    PropTypes.shape({
+      message: PropTypes.string.isRequired,
```
ZmagoD: you can extract string, number and func on the import statement

app/javascript/src/components/AlertsContainer/index.jsx
```diff
@@ -0,0 +1,71 @@
...
+      timeout: PropTypes.number,
+      onClose: PropTypes.func
+    }).isRequired
+  ).isRequired,
+  onAlertClose: PropTypes.func.isRequired
+}
+
+const mapStateToProps = ({ alerts }) => ({ alerts });
+
+const mapDispatchToProps = dispatch => ({
```
ZmagoD: you don't need this. You can pass the { clearAlert } directly in the connect function:
`connect(mapStateToProps, { clearAlert } )(AlertsContainer);`
but then you will have to change onAlertClose to clearAlert  in your component

PR #833, opened 2017-10-17
Add new team to redux [SCI-1686]

mlorb:
[SCI-1686](https://biosistemika.atlassian.net/browse/SCI-1686)


mlorb:
The thing is, that I wanted to add @flow to fix the hound, but then I realized that Zmago is doing that right now for all root components.

Ducz0r:
Okay... with flow, `prop-types` become obsolete if I understand this right?

If that is the case, by all means 🛳 away!

PR #845, opened 2017-10-23
Input validation extracting to reusable service [SCI-1670]

Ducz0r:
The best use case for how to handle this new validation can be seen on New Team page. The validators that I wrote (in JS) are merely a single-input-field validators, and can only get you so far. For more complex validation (e.g. whether password & password confirmation match), server-side validation should be used.

The validation is run by providing the same `tag` property to various elements that relate to the same "data field", as well as hooking the relevant validators to the `<InputFormControl>`, and voila - the validation magically works :rainbow:!

Also the `<ValidatedSubmitButton>` is automatically disabled as soon as there is any error on the form.

To call the form itself from within the hosting component, you should use the `ref` mechanism:

```javascript
<ValidatedForm ref={(f) => { this.myForm = f; }}>
```

Additional validators can be added once we need them. They are basically a Javascript function that receives `target` (input element reference) and `messageIds`, and returns an array of errors (which are a JSON object which support localization). This will be better once we onboard flow onto them.

As per usual, this required refactoring of Profile page...

Closes [SCI-1670](https://biosistemika.atlassian.net/browse/SCI-1670).


app/javascript/src/components/validation/components/ValidatedFormGroup.jsx
```diff
@@ -0,0 +1,41 @@
...
+import PropTypes from "prop-types";
+
+class ValidatedFormGroup extends Component {
+  constructor(props) {
+    super(props);
+
+    this.cleanProps = this.cleanProps.bind(this);
+  }
+
+  cleanProps() {
```
ZmagoD: same as above

app/javascript/src/components/validation/index.js
```diff
@@ -0,0 +1,5 @@
+export {default as ValidatedErrorHelpBlock} from "./components/ValidatedErrorHelpBlock";
+export {default as ValidatedForm} from "./components/ValidatedForm";
+export {default as ValidatedFormControl} from "./components/ValidatedFormControl";
+export {default as ValidatedFormGroup} from "./components/ValidatedFormGroup";
+export {default as ValidatedSubmitButton} from "./components/ValidatedSubmitButton";
```
ZmagoD: :+1:

app/javascript/src/scenes/SettingsPage/scenes/teams/new/index.jsx
```diff
@@ -22,13 +20,20 @@ import {
   SETTINGS_TEAMS_ROUTE,
   SETTINGS_TEAM_ROUTE
 } from "../../../../../config/routes";
-
-import {
-  NAME_MIN_LENGTH,
-  NAME_MAX_LENGTH,
-  TEXT_MAX_LENGTH
-} from "../../../../../config/constants/numeric";
+import { TEAMS_NEW_PATH } from "../../../../../config/api_endpoints";
```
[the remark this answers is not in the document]
ZmagoD: remove this

config/locales/en.yml
```diff
@@ -1822,9 +1822,7 @@ en:
       leave_team_error: "An error occured."
       leave_flash: "Successfuly left team %{team}."
     user:
-      blank_password_error: "Password can't be blank!"
-      passwords_dont_match: "Passwords don't match"
-      password_invalid: "Password is invalid!"
-      avatar_too_big: "Avatar file size must be less than 0.2 MB"
```
ZmagoD: check if the test case https://github.com/Ducz0r/scinote-web/blob/998519066c35e6c19e9bd0d3766ca8eeb986f75d/features/settings_page/profile.feature#L27 still work

Ducz0r: Good catch! Totally forgot about that...

Ducz0r:
@ZmagoD I've fixed your requests, but I alas cannot test integration tests without investing into getting them started in the first place. I know I need to set it up eventually, but in order to merge this asap, can you check if the updated feature test (I've merely renamed the expected Strings to match the ones that are actually being returned) works now?

ZmagoD:
To setup the test just run docker-clean and then make docker again to install the dependencies but since we are fixing other errors you may leave the integration tests for now

PR #849, opened 2017-10-25
adds environment variable that disables/enables the creation of new t…

ZmagoD:
…eam when user sign up
https://biosistemika.atlassian.net/browse/SCI-1724

set the environment variable NEW_TEAM_ON_SIGNUP to false, restart the server.

**important** for testing with addons set this variable to true or leave it undefined! Check if nothing brakes

### DIFF
--- a/app/controllers/users/registrations_controller.rb
+++ b/app/controllers/users/registrations_controller.rb
@@ -128,43 +128,51 @@ class Users::RegistrationsController < Devise::RegistrationsController
 
   def create
     render_403 && return unless Rails.configuration.x.enable_user_registration
-
     build_resource(sign_up_params)
     valid_resource = resource.valid?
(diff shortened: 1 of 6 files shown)


app/controllers/users/registrations_controller.rb
```diff
@@ -128,43 +128,51 @@ def new
...
+            resource.current_team_id = @team.id
+            resource.save
+          end
         end
+      else
+        render :new
       end
     else
-      render :new
+      if valid_resource
```
okriuchykhin: You can use elsif here

ZmagoD: It's a bit tricky because we have an independent check above that checks if an environment variable is present and also we render the new action again if the resource is not valid

PR #850, opened 2017-10-26
add scenic gem, teams datatables sql view

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-1732

fixes https://biosistemika.atlassian.net/browse/SCI-1731
todo: fix current_team on front end blocked by: https://github.com/biosistemika/scinote-web/pull/844


app/models/datatables/datatables_team.rb
```diff
@@ -0,0 +1,14 @@
+module Datatables
```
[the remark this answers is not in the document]
ZmagoD: I don't know what I have seen people treat this as an equivalent to active model.

app/javascript/src/components/actions/TeamsActions.js
```diff
@@ -5,7 +5,11 @@ import type {
   Actopm$SetCurrentTeam
 } from "flow-typed";
 import type { Dispatch } from "redux-thunk";
-import { getTeams, changeCurrentTeam } from "../../services/api/teams_api";
+import {
+  getTeams,
+  changeCurrentTeam,
+  fetchCurrentTeam
```
Ducz0r: I'm just curious - should we enforce some sort of common nomenclature here? We're using `getTeams` - so, "get" - to access an array of entries, and now `fetchCurrentTeam()` to get a single element. I'm okay with using "fetch", but we should then try to stick to this for future actions. Other option would be to also use "get" here.

app/models/datatables/datatables_team.rb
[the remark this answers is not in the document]
Ducz0r: Well it's up to us really. I think it's a bit confusing that it's not clear from this file that it is describing a view, rather than a database table. I think it would make sense to have it separated into its own folder, just to avoid confusion (e.g. a junior developer wanting to add a column to the `DatatablesTeam`, and struggling why the database migration doesn't work).

spec/models/datatables/teams_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'rails_helper'
```
Ducz0r: Nitpicking, but... rename this file to `spec/models/datatables/datatables_team_spec.rb`.

spec/support/api/schemas/teams.json
```diff
@@ -0,0 +1,21 @@
+{
```
Ducz0r: I suppose this file should also be renamed to `spec/support/api/schemas/datatables/datatables_teams.json`. I mean, if we want to mirror database model as much as possible.

PR #851, opened 2017-10-27
adds integration tests for about sciNote modal

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-1568


config/routes/confirmations.rb
```diff
@@ -0,0 +1,2 @@
+# scinote configurations routes
```
Ducz0r: Shouldn't this file be called `configurations.rb`?

config/routes.rb
```diff
@@ -18,7 +18,7 @@ def draw(routes_name)
     get '/settings/*all', to: 'client_api/settings#index'
 
     namespace :client_api, defaults: { format: 'json' } do
-      %i(activities teams notifications users).each do |path|
+      %i(activities teams notifications users confirmations).each do |path|
```
Ducz0r: Again, shouldn't we use `configurations` here?

PR #852, opened 2017-10-30
Onboard Flow onto validation components [SCI-1725]

Ducz0r:
This PR adds Flow onto validation components. They're not 100% covered, but mostly vary in the 70-90% range.

I think that for certain things (like context, and certain React shenanigans), Flow is still very immature and cumbersome to use. Also it sucks that you cannot make a reusable function definition (that I could reuse for validators), you can only create class interfaces, which I think is a bit of an overkill right now).

Closes [SCI-1725](https://biosistemika.atlassian.net/browse/SCI-1725).


app/javascript/src/components/validation/components/ValidatedSubmitButton.jsx
```diff
@@ -1,17 +1,19 @@
-import React from "react";
+// @flow
+
+import * as React from "react";
 import { Button } from "react-bootstrap";
 import PropTypes from "prop-types";
 
-const ValidatedSubmitButton = (props, context) =>
+type Props = {
+  children?: React.Node
```
ZmagoD: you can just import { Node } form react here we don't need all React types

ZmagoD:
Also run flow-typed install to update the types :banana:

Ducz0r:
@ZmagoD just double-checking if I imported the updated `flow-types` correctly? So I don't merge something bad 🙂.

package.json
```diff
@@ -54,6 +53,8 @@
     "css-loader": "^0.28.4",
     "extract-text-webpack-plugin": "^3.0.0",
     "file-loader": "^0.11.2",
+    "flow-bin": "^0.56.0",
```
ZmagoD: Why you moved flow-bin and flow-typed from dev dependencies?

.eslintrc.json
```diff
@@ -19,6 +19,8 @@
   "rules": {
     "import/extensions": "off",
     "import/no-unresolved": "off",
+    // Because of flow-typed & flow-bin
+    "import/no-extraneous-dependencies": ["error", { "devDependencies": true }],
```
ZmagoD: :+1:

package.json
Ducz0r: Because ESLint was telling me so :banana:. I've updated the configuration.

PR #856, opened 2017-11-07
trigger an alert to on canvas save if page reload [fixes SCI-1610]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-1610


ZmagoD:
Damn I forgot about that but who uses the reload button in the browser anyway :smile:

PR #858, opened 2017-11-08
Implement Integration test: forgot password.feature

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-1576

**!!!!**
set `MAILER_PORT` in your application.yml to '587'


features/sign_in_sign_up/forgot_password.feature
```diff
@@ -0,0 +1,46 @@
+Feature: Forgot password
```
Ducz0r: As I also mentioned in #857, we can rename folder `features/sign_in_sign_up/` into `featuers/sessions/`, it's more broad I think.

PR #884, opened 2017-11-16
Fixed; compared file size in wrong units [SCI-1810]

Zanz2:
Im not sure if any protocol will be too big now, the new file size is plenty for all my testing protocols.

### DIFF
--- a/app/controllers/protocols_controller.rb
+++ b/app/controllers/protocols_controller.rb
@@ -610,7 +610,7 @@ class ProtocolsController < ApplicationController
       end
       return 0 # return 0 stops the rest of the controller code from executing
     end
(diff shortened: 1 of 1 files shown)


app/controllers/protocols_controller.rb
```diff
@@ -610,7 +610,7 @@ def protocolsio_import_create
       end
       return 0 # return 0 stops the rest of the controller code from executing
     end
-    if file_size / 1000 > Constants::FILE_MAX_SIZE_MB
+    if file_size > Constants::FILE_MAX_SIZE_MB * 1000000
```
mlorb: You can use the function `megabytes` like this ` Constants::FILE_MAX_SIZE_MB.megabytes`

PR #890, opened 2017-11-21
export and import RTE images

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-1651

### DIFF
--- a/app/models/tiny_mce_asset.rb
+++ b/app/models/tiny_mce_asset.rb
@@ -58,7 +58,7 @@ class TinyMceAsset < ActiveRecord::Base
     if image.is_stored_on_s3?
       Kernel.open(presigned_url, 'rb')
     else
-      File.open(file.path, 'rb')
+      File.open(image.path, 'rb')
     end
   end
 
--- a/app/models/tiny_mce_asset.rb
+++ b/app/models/tiny_mce_asset.rb
@@ -54,6 +54,14 @@ class TinyMceAsset < ActiveRecord::Base
     end
   end
 
+  def open
+    if image.is_stored_on_s3?
+      Kernel.open(presigned_url, 'rb')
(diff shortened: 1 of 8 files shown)


app/utilities/protocols_exporter.rb
```diff
@@ -259,7 +259,7 @@ def generate_eln_xsd
     eln_xsd << "<xs:element name=\"fileType\" " \
                "type=\"xs:string\"></xs:element>\n"
     eln_xsd << "</xs:all>\n"
-    eln_xsd << "<xs:attribute name=\"id\" type=\"xs:int\" " \
+    eln_xsd << "<xs:attribute tokenId=\"xs:string\" name=\"id\" type=\"xs:int\" " \
```
Ducz0r: Shouldn't this be like this:
```ruby
    eln_xsd << "<xs:attribute name=\"id\" type=\"xs:int\" " \
               "use=\"required\"></xs:attribute>\n"
    eln_xsd << "<xs:attribute name=\"tokenId\" type=\"xs:string\" " \
               "use=\"required\"></xs:attribute>\n"
```

app/models/tiny_mce_asset.rb
```diff
@@ -54,6 +54,14 @@ def url(style = :large, timeout: Constants::URL_LONG_EXPIRE_TIME)
     end
   end
 
+  def open
+    if image.is_stored_on_s3?
+      Kernel.open(presigned_url, 'rb')
+    else
+      File.open(file.path, 'rb')
```
Ducz0r: Again, I think that here the following should be written (otherwise it crashes for me!):
```ruby
File.open(image.path, 'rb')
```

Ducz0r:
Wohooo, :rocket: away!

PR #894, opened 2017-11-24
Adds task to copy images [fixes SCI-1650]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-1650

### DIFF
--- a/lib/tasks/tiny_mce_asset.rake
+++ b/lib/tasks/tiny_mce_asset.rake
@@ -4,11 +4,11 @@ namespace :tiny_mce_asset do
   task remove_obsolete_images: :environment do
     TinyMceAsset.where('step_id IS ? AND ' \
                        'result_text_id IS ? AND created_at < ?',
-                       nil, nil, 7.days.ago)
+                       nil, nil, 7.days.ago).destroy_all
   end
 
   desc 'Generate new tiny_mce_assets and replace old assets in RTE'
-  task regenerage_images: :environment do
+  task regenerate_images: :environment do
     regex = /\[~tiny_mce_id:([0-9a-zA-Z]+)\]/
     replaced_images = 0
     failed_attempts = 0
@@ -16,9 +16,7 @@ namespace :tiny_mce_asset do
     failed_attempts_ids = []
     puts 'Start processing steps...'
     Step.find_each do |step|
-      next unless step.description
-      images = step.description.match(regex)
-      next unless images
+      next unless step.description && step.description.match(regex)
       team = step.protocol.team
       step.description.gsub!(regex) do |el|
         match = el.match(regex)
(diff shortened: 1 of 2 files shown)


lib/tasks/tiny_mce_asset.rake
```diff
@@ -6,4 +6,75 @@ namespace :tiny_mce_asset do
                        'result_text_id IS ? AND created_at < ?',
                        nil, nil, 7.days.ago)
   end
+
+  desc 'Generate new tiny_mce_assets and replace old assets in RTE'
+  task regenerage_images: :environment do
```
Ducz0r: Typo, it should be `regenerate_images`!

lib/tasks/tiny_mce_asset.rake
```diff
@@ -6,4 +6,75 @@ namespace :tiny_mce_asset do
                        'result_text_id IS ? AND created_at < ?',
                        nil, nil, 7.days.ago)
```
Ducz0r: The Rake task `remove_obsolete_images` doesn't do anything. It's just a `where` statement. I assume a `.destroy_all` is missing at the end of the query.

lib/tasks/tiny_mce_asset.rake
```diff
@@ -6,4 +6,75 @@ namespace :tiny_mce_asset do
...
+        match = el.match(regex)
+        old_img = TinyMceAsset.find_by_id(match[1])
+        new_img = TinyMceAsset.create(image: old_img.image,
+                                      team: team,
+                                      reference: step)
+        if new_img
+          # This image will be removed by `remove_obsolete_images` rake task
+          # until all the steps are not updated we still need this image
+          # in case it appears on some other step
+          old_img.update_attributes(team_id: nil, step_id: nil)
```
Ducz0r: The team reference should be kept, otherwise the destruction of old images doesn't work (due to team space calculations).

lib/tasks/tiny_mce_asset.rake
```diff
@@ -6,4 +6,75 @@ namespace :tiny_mce_asset do
...
+        match = el.match(regex)
+        old_img = TinyMceAsset.find_by_id(match[1])
+        new_img = TinyMceAsset.create(image: old_img.image,
+                                      team: team,
+                                      reference: result_text)
+        if new_img
+          # This image will be removed by `remove_obsolete_images` rake task
+          # until all the steps are not updated we still need this image
+          # in case it appears on some other step
+          old_img.update_attributes(team_id: nil, step_id: nil)
```
Ducz0r: The team reference should be kept, otherwise the destruction of old images doesn't work (due to team space calculations).

Also, you probably want to set the `result_text_id` to `nil`, not `step_id`...

lib/tasks/tiny_mce_asset.rake
okriuchykhin: Shouldn't be here OR instead of AND?

mlorb:
I think you are right @Ducz0r. I agree, we should add new task to jira. Or if @ZmagoD prefers to fix it here by adding one if statement. :stuck_out_tongue_winking_eye:

ZmagoD:
So the rake task is for the images that are added to the step/result and the step/result is not saved (doesn't have the ID) for the removing of images there is a method destroy_removed_tiny_mce_assets that looks like is not working

Ducz0r:
Cool, so you fixed both things in this branch? I'll check it out.

Ducz0r:
Sorry for taking so long. This all works now, :clap:, merge awaaaaay 🌟 🎉!

PR #900, opened 2017-12-01
adds premissions service to FE

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-1855
and 
https://biosistemika.atlassian.net/browse/SCI-1854
There is one failing test because we don't have the new permissions system in place


app/controllers/client_api/permissions_controller.rb
```diff
@@ -2,7 +2,7 @@ module ClientApi
   class PermissionsController < ApplicationController
     before_action :generate_permissions_object, only: :state
```
Ducz0r: Probably also need to change `:state` into `:status` here, right?

And also fix it if it needs to be fixed somewhere else, not sure (`routes.rb`, any front-end call?).

Ducz0r:
Alright, :ship: it away!

PR #901, opened 2017-12-03
Protocols.io code beautification and refactoring [SCI-1929]

Zanz2:
Made ugly code nicer and moved helper code to helper file.
[SCI-1929](https://biosistemika.atlassian.net/browse/SCI-1929)


app/helpers/protocols_io_helper.rb
```diff
@@ -145,4 +145,283 @@ def not_null(attribute)
   def close_open_html_tags(text)
     Nokogiri::HTML::DocumentFragment.parse(text).to_html
   end
+
+  def prepare_for_view(attribute_text, size)
```
mlorb: You could write this method in more compact way:
```
def prepare_for_view(attribute_text, size, table='no_table')
  attribute_text_not_null = not_null(attribute_text)
  attribute_text_not_null = string_html_table_remove(attribute_text_not_null) if table=='table_present'
  pio_eval_len(sanitize_input(attribute_text_not_null), size)
end
```
Then you would call the method like this:
`prepare_for_view(attribute_text, ProtocolsIoHelper::PIO_ELEMENT_RESERVED_LENGTH_SMALL, 'table_present')`

app/helpers/protocols_io_helper.rb
```diff
@@ -145,4 +145,283 @@ def not_null(attribute)
...
+      append =
+        t('protocols.protocols_io_import.comp_append.expected_result') +
+        prepare_for_view(iterating_key, 'small-no-table') +
+        '<br>'
+      return append
+    end
+    ''
+  end
+
+  def pio_stp_8(iterating_key) # protocols io software package parser
```
mlorb: You could move all the if checks inside the `parse_elements_array` loop like this:
```
parse_elements_array.each do |element|
  return ' ' unless iterating_key[element]
  append += fill_attributes(element, iterating_key[element], '8')
end
```
You can do these changes also in other similar methods:` pio_stp_9(iterating_key)`, `pio_stp_15(iterating_key)`, itd.

PR #906, opened 2017-12-13
Merge gilson-iot-rails4 into master

Ducz0r:
This should be thoroughly reviewed.


Gemfile
```diff
@@ -26,7 +30,7 @@ gem 'uglifier', '>= 1.3.0'
 # jQuery & plugins
 gem 'jquery-turbolinks'
 gem 'jquery-rails'
-gem 'jquery-ui-rails'
+gem 'jquery-ui-rails', '~> 5.0'
```
okriuchykhin: I think it's not needed

Gemfile
```diff
@@ -52,7 +56,7 @@ gem 'wicked_pdf'
 gem 'silencer' # Silence certain Rails logs
 gem 'wkhtmltopdf-heroku'
 gem 'remotipart', '~> 1.2' # Async file uploads
-gem 'faker' # Generate fake data
+gem 'faker', '~> 1.7.3' # Generate fake data
```
okriuchykhin: This one is also not needed

Gemfile
```diff
@@ -92,9 +102,13 @@ group :production do
 end
 
 group :test do
-  gem 'minitest-reporters', '~> 1.1'
```
okriuchykhin: We should keep this

Gemfile
```diff
@@ -92,9 +102,13 @@ group :production do
...
-  gem "shoulda-context"
```
okriuchykhin: This one as well

app/models/user.rb
```diff
@@ -197,7 +199,18 @@ def name
   end
 
   def name=(name)
-    full_name = name
```
okriuchykhin: I think we need keep this

app/views/user_notifications/_list.html.erb
```diff
@@ -3,26 +3,9 @@
...
-        </div>
-        <% end %>
-        <% if notification.type_of == 'deliver' %>
-        <div class="text-center">
-          <span class="deliver"><%= fa_icon 'truck' %></span>
-        </div>
-        <% end %>
+        <%= render partial: "user_notifications/icons/#{notification.type_of}",
+                   locals: {notification: notification},
+                   formats: [:html] %>
```
okriuchykhin: Should we use this refactored version of notifications?

Ducz0r: What do you mean by refactored notifications? You're not referring to the Reactified, no?

okriuchykhin: No, just splitting notification types to different files. As far as I can remember it was done for easier inserts of new types, you just need to create partial with same name as notification type.

Gemfile
Ducz0r: The version suffix, `'~ 5.0'`? Or the Gem itself?

Gemfile
Ducz0r: The version suffix, `'~> 1.7.3'`? Or the Gem itself?

app/models/user.rb
Ducz0r: Instead of `self.full_name = name`?

okriuchykhin: Yes, self is redundant here

Gemfile
okriuchykhin: Version

Gemfile
okriuchykhin: Version suffix

PR #911, opened 2017-12-14
Permissions - refactor organization & team level permissions (core) [SCI-1797]

mlorb:
[SCI-1797](https://biosistemika.atlassian.net/browse/SCI-1797)

### DIFF
--- a/app/permissions/team.rb
+++ b/app/permissions/team.rb
@@ -4,6 +4,11 @@ Canaid::Permissions.register_for(Team) do
     user.is_member_of_team?(team)
   end
 
+  # edit team name, edit team description
+  can :update_team do |user, team|
+    user.is_admin_of_team?(team)
+  end
+
   # invite user to team
   can :create_user_team do |user, team|
     user.is_admin_of_team?(team)
--- a/app/permissions/team.rb
+++ b/app/permissions/team.rb
@@ -36,6 +36,11 @@ Canaid::Permissions.register_for(Team) do
     user.is_normal_user_or_admin_of_team?(team)
   end
 
+  # create, copy, edit, destroy repository
+  can :manage_repository do |user, team|
+    user.is_admin_of_team?(team)
+  end
+
   # create, import, edit, delete repository record
   can :manage_repository_row do |user, team|
     user.is_normal_user_or_admin_of_team?(team)
--- a/app/permissions/team.rb
+++ b/app/permissions/team.rb
@@ -40,4 +40,12 @@ Canaid::Permissions.register_for(Protocol) do
       (protocol.in_repository_public? ||
       protocol.in_repository_private? && user == protocol.added_by)
   end
+
+  # edit protocol in repository,
+  # create, edit, delete or reorder step in repository
+  can :update_protocol_in_repository do |user, protocol|
+    user.is_normal_user_or_admin_of_team?(protocol.team) &&
+      user == protocol.added_by &&
+      protocol.in_repository_active?
+  end
 end
--- a/app/permissions/team.rb
+++ b/app/permissions/team.rb
@@ -53,4 +53,10 @@ Canaid::Permissions.register_for(Protocol) do
     user.is_normal_user_or_admin_of_team?(protocol.team) &&
       user == protocol.added_by
   end
+
+  # clone protocol in repository
+  can :clone_protocol do |user, protocol|
+    can_create_protocol?(user, protocol.team) &&
+      can_read_protocol_in_repository?(user, protocol)
+  end
 end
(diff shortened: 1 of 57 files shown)


app/permissions/team.rb
```diff
@@ -0,0 +1,119 @@
...
+    user.is_member_of_team?(team)
+  end
+
+  # edit team name, edit team description
+  can :update_team do |user, team|
+    user.is_admin_of_team?(team)
+  end
+
+  # invite user to team, change user's role, remove user from team
+  can :manage_user_team do |user, team|
```
Ducz0r: How about `can_manage_user_teams?`, or even `can_manage_team_users?`. I like the latter, it's not "correct" per-se if we base names from actual `ActiveRecord` names, but it tells you exactly what the permission is about.

app/permissions/team.rb
```diff
@@ -0,0 +1,119 @@
...
+  end
+
+  # create, import, edit, delete samples
+  can :manage_samples do |user, team|
+    user.is_normal_user_or_admin_of_team?(team)
+  end
+
+  # create custom field
+  # create, update, delete sample type or sample group
+  can :manage_sample_elements do |user, team|
```
Ducz0r: How about `can_manage_sample_columns?`. Damn, Phil Karlton was totally right.

app/permissions/team.rb
```diff
@@ -0,0 +1,119 @@
...
+    user.is_admin_of_team?(team)
+  end
+
+  # invite user to team, change user's role, remove user from team
+  can :manage_user_team do |user, team|
+    user.is_admin_of_team?(team)
+  end
+
+  # create project
+  can :create_project do |user, team|
```
Ducz0r: I don't know, might be better to use plural? e.g. `can_create_projects?`

app/permissions/team.rb
```diff
@@ -0,0 +1,119 @@
...
+    user.is_normal_user_or_admin_of_team?(team)
+  end
+
+  # create protocol in repository, import protocol to repository
+  can :create_protocol do |user, team|
```
Ducz0r: Same, `can_create_protocols?` ?

app/permissions/team.rb
```diff
@@ -0,0 +1,119 @@
...
+  end
+
+  # create custom field
+  # create, update, delete sample type or sample group
+  can :manage_sample_elements do |user, team|
+    user.is_normal_user_or_admin_of_team?(team)
+  end
+
+  # create, copy repository
+  can :manage_repository do |user, team|
```
Ducz0r: Probably better would be to keep `can_create_repositories?`. Manage relates more to editing, updating it right?

app/permissions/team.rb
```diff
@@ -0,0 +1,119 @@
...
+    user.is_admin_of_team?(team)
+  end
+
+  # create, import, edit, delete repository records
+  can :manage_repository_rows do |user, team|
+    user.is_normal_user_or_admin_of_team?(team)
+  end
+
+  # create repository column
+  can :manage_repository_column do |user, team|
```
Ducz0r: Plural, `can_manage_repository_columns?`

app/permissions/team.rb
```diff
@@ -0,0 +1,119 @@
...
+  end
+
+  # toggle protocol visibility (public, private, archive, restore)
+  can :update_protocol_type_in_repository do |user, protocol|
+    user.is_normal_user_or_admin_of_team?(protocol.team) &&
+      user == protocol.added_by
+  end
+
+  # clone protocol in repository
+  can :clone_protocol do |user, protocol|
```
Ducz0r: No `<...>_in_repository?` 🙂?

app/permissions/team.rb
mlorb: Maybe even `can_create_protocols_in_repository?` ?

app/views/repositories/_repository.html.erb
```diff
@@ -34,11 +34,11 @@
            data-toggle="dropdown"
            aria-haspopup="true"
            aria-expanded="true"
-           <%= "disabled='disabled'" unless can_manage_repository?(repository.team) %>>
+           <%= "disabled='disabled'" if !can_update_or_delete_repository?(repository) and !can_create_repositories?(repository.team) %>>
```
Ducz0r: Use `&&` instead of `and`.

app/views/repositories/_repository.html.erb
```diff
@@ -34,11 +34,11 @@
...
         <span class="glyphicon glyphicon-cog"></span>
         <span class="caret"></span>
       </div>
-      <% if can_manage_repository?(repository.team) %>
+      <% if can_update_or_delete_repository?(repository) or can_create_repositories?(repository.team) %>
```
Ducz0r: Same here, use `||` instead of `or`. `and` and `or` should be used for control flow, not boolean operations [[see this]](http://www.virtuouscode.com/2010/08/02/using-and-and-or-in-ruby/). Of course, we previously also used them wrongly, especially in scope of permissions.

PR #914, opened 2017-12-19
Fix repositories [SCI-1873]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-1873

### DIFF
--- a/app/datatables/repository_datatable.rb
+++ b/app/datatables/repository_datatable.rb
@@ -183,7 +183,8 @@ class RepositoryDatatable < CustomDatatable
   # number of samples/all samples it's dependant upon sort_record query
   def fetch_records
     records = get_raw_records
-    records = filter_records(records) if dt_params[:search].present?
+    records = filter_records(records) if dt_params[:search].present? &&
+                                         dt_params[:search][:value].present?
     records = sort_records(records) if order_params.present?
     records = paginate_records(records) unless dt_params[:length].present? &&
                                                dt_params[:length] == '-1'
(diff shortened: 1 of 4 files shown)


app/datatables/repository_datatable.rb
```diff
@@ -230,12 +232,12 @@ def filter_records(repo_rows)
 
   # Override default sort method if needed
   def sort_records(records)
-    if params[:order].present? && params[:order].length == 1
-      if sort_column(params[:order].values[0]) == ASSIGNED_SORT_COL
+    if order_params.present?
```
ZmagoD: as above it always evaluate true so why we need this? Before that was also a :banana: before params[:order].length == 1 
hmm...

app/datatables/repository_datatable.rb
```diff
@@ -246,40 +248,26 @@ def sort_records(records)
...
-        dir = sort_direction(params[:order].values[0])
+        dir = sort_direction(order_params)
 
         records.joins(
-          "LEFT OUTER JOIN my_module_repository_rows ON
-          (repository_rows.id = my_module_repository_rows.repository_row_id
-          AND (my_module_repository_rows.my_module_id = #{@my_module.id} OR
-                            my_module_repository_rows.id IS NULL))"
-        ).order("my_module_repository_rows.id NULLS #{direction}")
+          "LEFT OUTER JOIN (SELECT repository_cells.repository_row_id,
```
ZmagoD: it was something wrong with the query?
Do this affect the datatable on task?

app/datatables/repository_datatable.rb
```diff
@@ -335,10 +323,10 @@ def sort_assigned_records(records, direction)
       ids = unassigned + assigned
     end
 
-    order_by_index = ActiveRecord::Base.send(
+    order_by_index = ActiveRecord::Base.__send__(
```
ZmagoD: I don't think that ActiveRecord::Base overrides Object.send method. Do you have a good reason to use this?

app/datatables/repository_datatable.rb
okriuchykhin: That was an old code, it was changed in master.

app/datatables/repository_datatable.rb
okriuchykhin: Just linter was complaining about it

ZmagoD: fuck the linter this is not python :smile:

app/datatables/repository_datatable.rb
okriuchykhin: Ok, removed that check

ZmagoD:
Good job :ship:

PR #917, opened 2017-12-20
Fix adding users to a task [SCI-1870]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-1870


okriuchykhin:
Couldn't reproduce it

ZmagoD:
hmm maybe there is another unrelated problem with Admin user that caused this issues now when I invite user works fine for me if I invite the admin user his status is pending and I have the relation to the team and project already when I switch to this branch maybe is this the cause of this bug. OK whatever it is good to go :ship:

PR #924, opened 2018-01-04
Nil image guard [fixes SCI-1907]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-1907

### DIFF
--- a/lib/tasks/tiny_mce_asset.rake
+++ b/lib/tasks/tiny_mce_asset.rake
@@ -21,6 +21,9 @@ namespace :tiny_mce_asset do
       step.description.gsub!(regex) do |el|
         match = el.match(regex)
         old_img = TinyMceAsset.find_by_id(match[1])
+        # skip other processing and deletes tiny_mce tag
+        # if image is not in database
+        next unless old_img
         new_img = TinyMceAsset.create(image: old_img.image,
                                       team: team,
                                       reference: step)
@@ -47,6 +50,9 @@ namespace :tiny_mce_asset do
       result_text.text.gsub!(regex) do |el|
         match = el.match(regex)
         old_img = TinyMceAsset.find_by_id(match[1])
+        # skip other processing and deletes tiny_mce tag
+        # if image is not in database
+        next unless old_img
         new_img = TinyMceAsset.create(image: old_img.image,
                                       team: team,
                                       reference: result_text)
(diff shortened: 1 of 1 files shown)


Ducz0r:
@ZmagoD can you make this script more error-proof? E.g. have a parameter for the script of timestamp, or even asset id, or something like that, so in case the script crashes, we can restart it from where it stopped, so we don't always need to re-run through all assets.

lib/tasks/tiny_mce_asset.rake
```diff
@@ -7,20 +7,33 @@ namespace :tiny_mce_asset do
...
+    all_images = TinyMceAsset.where('step_id IS NOT NULL').count
     failed_attempts_ids = []
     puts 'Start processing steps...'
-    Step.find_each do |step|
+    params = { batch_size: 100 }
+    if args.present? && args[:last_id].present?
+      # fetch all steps and sort them asc
+      params[:start] = args[:last_id].to_i
+    end
+    Step.find_each(batch_size: 100) do |step|
```
Ducz0r: You probably meant `Step.find_each(params) do |step|` right 🙂.

lib/tasks/tiny_mce_asset.rake
```diff
@@ -40,13 +53,42 @@ namespace :tiny_mce_asset do
...
+    puts "Failed attempts: #{failed_attempts}"
+    puts "TinyMceAsset ids of failed attempts: #{failed_attempts_ids}"
+    puts '-----------------------------------------'
+  end
+
+  desc 'Generate new tiny_mce_assets and replace old assets in RTE ' \
+       'for results. Assign the last printed id if the script crashes or ' \
+       'id + 1 if there is a problematic asset'
+  task :regenerate_results_images, [:last_id] => :environment do |_, args|
+    regex = /\[~tiny_mce_id:([0-9a-zA-Z]+)\]/
```
Ducz0r: If it's the same regex as for steps, you can maybe just define it as constant on the `namespace` level.

lib/tasks/tiny_mce_asset.rake
```diff
@@ -40,13 +53,42 @@ namespace :tiny_mce_asset do
...
+    replaced_images = 0
+    failed_attempts = 0
+    all_images = TinyMceAsset.where('result_text_id IS NOT NULL').count
```
Ducz0r: `TinyMceAsset.where.not(result_text: nil).count`

PR #927, opened 2018-01-05
fix import protocol preview [fixes SCI-1867]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-1867

Works well for new exports


app/assets/javascripts/protocols/import_export/import.js.erb
```diff
@@ -230,20 +230,34 @@ function importProtocolFromFile(
 
   // display tiny_mce_assets in step description
   function displayTinyMceAssetInDescription(node, folder, stepGuid) {
+    debugger;
```
Ducz0r: We probably want to remove this right 🙂.

Ducz0r:
Aiiight, :shipit:!

PR #929, opened 2018-01-09
Add create_teams permission checks [SCI-1797]

mlorb:
[SCI-1797](https://biosistemika.atlassian.net/browse/SCI-1797)

### DIFF
--- a/app/controllers/users/settings/teams_controller.rb
+++ b/app/controllers/users/settings/teams_controller.rb
@@ -19,6 +19,9 @@ module Users
         :destroy
       ]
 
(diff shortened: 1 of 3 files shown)


app/controllers/users/settings/teams_controller.rb
```diff
@@ -134,6 +137,10 @@ def destroy
 
       private
 
+      def check_create_team_permission
+        can_create_teams?
```
Ducz0r: You probably want to do something if user can't create teams, like render 403 or something?

PR #937, opened 2018-01-16
Enable loading of external React components

oprearocks:
This PR embodies both Rails and React updates to enable the loading of React components that are not tied/developed with the core application.


ZmagoD:
ready to :rocket: :tada: :banana:

PR #946, opened 2018-01-17
Fix addons display in info modal [SCI-1875]

okriuchykhin:


app/javascript/src/components/Navigation/components/AboutScinoteModal.jsx
```diff
@@ -30,7 +30,11 @@ export default (props: Props): Node => {
             <strong>
               <FormattedMessage id="general.addon_versions" />
             </strong>
-            {addons.map((addon: string): Node => <p>{addon}</p>)}
+            {
+              addons.map(
+                (addon: string): Node => (<p key={addon}>{addon}</p>)
```
oprearocks: The reason we pass `key={addon}` is to enable React to properly work with its diffing algorithm. `addon` was chosen as it is not repetitive and is quite unique — we can't have 2 addons called gamification.

PR #950, opened 2018-01-18
Clear selected rows when changing pages on samples table[SCI-1329]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-1329


ZmagoD:
Looks ok! :ship:

PR #952, opened 2018-01-19
adds Delayed::Worker initializer [fixes SCI_1948]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-1948

### DIFF
--- a/config/initializers/delayed_job_config.rb
+++ b/config/initializers/delayed_job_config.rb
@@ -1,0 +1,65 @@
+# DelayedWorkerConfig is a configuration object that parses and sanitize
+# environmental variables. @note Used only for Delayed::Worker settings
+module DelayedWorkerConfig
+  module_function
+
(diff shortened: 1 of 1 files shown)


config/initializers/delayed_job_config.rb
```diff
@@ -0,0 +1,65 @@
...
+  def read_ahead
+    value = ENV['DELAYED_WORKER_READ_AHEAD'].to_i
+    return 10 if value.zero?
+    value
+  end
+
+  # By default all jobs will be queued without a named queue.
+  # A default named queue can be specified by using
+  def default_queue_name
+    value = ENV['DELAYED_WORKER_DEFAULT_QUEUE_NAME']
```
Ducz0r: Does this affect the live jobs that are in DB and have no name? I guess their `:queue` equals to `nil`. It shouldn't really matter right?

ZmagoD: yes, Delayed job handles that. This way I think is more concrete

PR #953, opened 2018-01-22
update user settings controllers to new permission system [fixes SCI-…

ZmagoD:
…1960]
 https://biosistemika.atlassian.net/browse/SCI-1960


app/controllers/users/invitations_controller.rb
```diff
@@ -192,7 +192,7 @@ def check_invite_users_permission
       @role = params['role']
 
       render_403 if @emails && @emails.empty?
-      render_403 if @team && !is_admin_of_team(@team)
+      render_403 if @team && !can_read_team?(@team)
```
mlorb: Here goes `can_manage_team_users?`

app/controllers/users/settings/teams_controller.rb
```diff
@@ -147,7 +147,7 @@ def load_user
 
       def load_team
         @team = Team.find_by_id(params[:id])
-        render_403 unless is_admin_of_team(@team)
+        render_403 unless can_read_team?(@team)
```
mlorb: I would remove this check from `load_team` and add separate method `check_update_team_permission`, where you would check `can_update_team?(@team)`

app/permissions/team.rb
```diff
@@ -117,3 +117,19 @@
     can_create_repository_columns?(user, repository_column.repository.team)
   end
 end
+
```
mlorb: Ne rabaš :)

app/controllers/users/settings/user_teams_controller.rb
```diff
@@ -150,7 +150,7 @@ def load_user_team
         # Don't allow the user to modify UserTeam-s if he's not admin,
         # unless he/she is modifying his/her UserTeam
         if current_user != @user_t.user &&
-           !is_admin_of_team(@user_t.team)
+           !can_read_team?(@user_t.team)
```
mlorb: Here the same, I would make separate methods `check_leave_team_permission` and `check_manage_user_team_permission`.

app/controllers/users/settings/teams_controller.rb
ZmagoD: but there is a banana. Until now we didn't show this controller actions if the user was not an admin

ZmagoD: Should I make a new permission check like can_manage_team?

mlorb: Just use then `can_update_team?(@team)`

app/controllers/users/settings/user_teams_controller.rb
mlorb: And here use `can_manage_team_users?`

app/permissions/team.rb
ZmagoD: done

mlorb:
Dropdown still needs to be fixed.

ZmagoD:
This is a problem with canaid gem that evaluates can_create_teams to false. This issue will be probably fixed with this PR https://github.com/biosistemika/scinote-web/pull/968

ZmagoD:
Try to remove all add-ons from your source tree

PR #957, opened 2018-01-24
Steps ordering fix [SCI-1816]

Zanz2:
If a step is now reordered in protocols io, a method now sorts trough the step GUIDs and previous GUIDs to determine the step order and resorts them.


app/helpers/protocols_io_helper.rb
```diff
@@ -285,6 +285,49 @@ def protocols_io_fill_desc(json_hash)
...
+  def protocols_io_count_steps(step_json)
+    count = 0
+    step_json['steps'].each do |step|
+      count += 1
+    end
+    count
+  end
+
+  def protocols_io_guid_reorder_step_json(unordered_step_json)
+    base_step = protocols_io_return_first_step_guid(unordered_step_json)
```
mlorb: You can find the base step like this:
`unordered_step_json['steps'].find{|step| step['previous_guid'].nil?}`

app/helpers/protocols_io_helper.rb
```diff
@@ -285,6 +285,49 @@ def protocols_io_fill_desc(json_hash)
...
+          step_counter += 1
+          base_step = step
+      end
+    end
+    step_order
+  end
+
+  def correct_order_guid_check(step_array, max)
+    return false if step_array.length != max
+    step_array.each_with_index do |step, index|
```
mlorb: These checks are redundant. Its enough to just check the max length.

app/helpers/protocols_io_helper.rb
```diff
@@ -285,6 +285,49 @@ def protocols_io_fill_desc(json_hash)
...
+
+  def protocols_io_guid_reorder_step_json(unordered_step_json)
+    base_step = protocols_io_return_first_step_guid(unordered_step_json)
+    number_of_steps = protocols_io_count_steps(unordered_step_json)
+    step_order = []
+    step_counter = 0
+    step_order[step_counter] = base_step
+    step_counter += 1
+    while !correct_order_guid_check(step_order,number_of_steps)
+      unordered_step_json['steps'].each do |step|
```
mlorb: You can use here instead of this each loop:
`unordered_step_json['steps'].find{|step| step['previous_guid']==base_step['guid']}`
to find the next step.

PR #958, opened 2018-01-24
Made images now display in various imported protocols.io protocol fields [SCI-1807]

Zanz2:
Images are now displayed in: protocol description, before_start, warnings,guidelines, manuscript_citation, and also in Step description and step expected result

(the html  tag is whitelisted in these components)


Zanz2:
napaka spet

PR #964, opened 2018-01-26
Steps order fix [SCI-1816]

Zanz2:
I made a mistake with the last commit in the pull request before this one, here i fixed the code according to review and added just one small unrelated fix for the title special characters getting espcaped (more on that in the commit comments)


app/helpers/protocols_io_helper.rb
```diff
@@ -285,6 +287,29 @@ def protocols_io_fill_desc(json_hash)
...
   end
 
+  def protocols_io_guid_reorder_step_json(unordered_step_json)
+    base_step = unordered_step_json.find { |step| step['previous_guid'].nil? }
+    number_of_steps = unordered_step_json.size
+    step_order = []
+    step_counter = 0
+    step_order[step_counter] = base_step
+    step_counter += 1
+    until correct_order_guid_check(step_order, number_of_steps)
```
mlorb: You don't need a new method `correct_order_guid_check`. Just do the check here.

mlorb:
Great! :+1: :1st_place_medal: Merge away!

PR #968, opened 2018-01-26
Fix canaid initializer so it only uses required Gems' permissions

Ducz0r:
This should make sure that other addons that are located in
/addons folder, but not actually required by the core,
do not cause trouble.

### DIFF
--- a/config/initializers/canaid.rb
+++ b/config/initializers/canaid.rb
@@ -12,7 +12,7 @@ Canaid.configure do |config|
                 .eager_load_paths
                 .select { |p| p.ends_with?('permissions') }
                 .first
-    continue unless full_path
+    next unless full_path
     res = rx.match(full_path)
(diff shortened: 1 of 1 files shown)


config/initializers/canaid.rb
```diff
@@ -1,4 +1,19 @@
...
+  # Include the included addons' permissions folders
+  rx = %r{^.*(addons/.*/app/permissions)$}
+  list_all_addons.each do |addon|
+    full_path = addon::Engine
+                .instance
+                .config
+                .eager_load_paths
+                .select { |p| p.ends_with?('permissions') }
+                .first
+    continue unless full_path
```
mlorb: Do you mean `next`?

Ducz0r: :banana:, good catch!

PR #981, opened 2018-02-05
fixes bug with step table update [fixes SCI-2026]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-2026


mlorb:
Can you fix also checklists, because there is the same bug... please :sun_with_face:

PR #982, opened 2018-02-06
Remove component loader (so builds pass)

Ducz0r:
This of course means that the addons cannot hijack React components into core as of right now.


config/initializers/load_addons_specs.rb
```diff
@@ -1,46 +1,47 @@
-require 'singleton'
```
ZmagoD: why that? this load the specs and only if rails env is test or development so it could not affect the build on Heroku

Ducz0r: Woops, my bad, :banana:.

PR #984, opened 2018-02-06
Fixed tables getting cut when certain attributes are at max length [SCI-1834]

Zanz2:
Step description, step expected result, protocol description, protocol manuscript citation, protocol guidelines, protocol warning and protocol before start attributes all now dont get cut tables if they exceed the character limit and have a table included (the table now doesnt get cut, even though text will)

### DIFF
--- a/app/helpers/protocols_io_helper.rb
+++ b/app/helpers/protocols_io_helper.rb
@@ -64,7 +64,6 @@ module ProtocolsIoHelper
       )
       tables[table_counter.to_s]['name'] = nil
     end
-    # return string_without_tables, tables
     return tables, string_without_tables
   end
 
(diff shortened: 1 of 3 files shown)


mlorb:
Import protocol preview and imported protocol are differently cutted off. Check where is the problem.

mlorb:
The scrollbar is now separated from the table, that you are scrolling.

mlorb:
The problem with the separated scrollbar is still present, if the table is smaller then the fixed height.

mlorb:
I realized that this bug was introduced with some other fix, not this PR. That's why you don't need to fix it in this PR. Darja will create a new jira task for it.

Zanz2:
removed last commit (width 300px) and am now merging into master

PR #989, opened 2018-02-07
Table missing scrollbar formatting [SCI-1818]

Zanz2:
Tables are now always atleast 5x5 cells when they get imported ( like scinote standard new table), which also fixes the styling errors in protocol preview, the tables were having


mlorb:
I can't reproduce the bug, because this new bug with the separated scrollbar is present here as well. So we have to wait with this PR until the separated scrollbar issue is fixed.

PR #992, opened 2018-02-07
setup models for repository list column [fixes SCI-2015]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-2015

### DIFF
--- a/app/models/repository_list_item.rb
+++ b/app/models/repository_list_item.rb
@@ -1,6 +1,5 @@
 class RepositoryListItem < ApplicationRecord
-  has_many :repository_list_values,
-           foreign_key: 'selected_item_id'
+  has_many :repository_list_values, inverse_of: :repository_list_item
   belongs_to :repository, inverse_of: :repository_list_items
   belongs_to :created_by,
              foreign_key: :created_by_id,
@@ -8,7 +7,7 @@ class RepositoryListItem < ApplicationRecord
   belongs_to :last_modified_by,
              foreign_key: :last_modified_by_id,
              class_name: 'User'
-  validates :name,
+  validates :data,
             presence: true,
             uniqueness: { scope: :repository, case_sensitive: false },
             length: { maximum: Constants::TEXT_MAX_LENGTH }
--- a/app/models/repository_list_item.rb
+++ b/app/models/repository_list_item.rb
@@ -1,6 +1,7 @@
 class RepositoryListItem < ApplicationRecord
   has_many :repository_list_values, inverse_of: :repository_list_item
   belongs_to :repository, inverse_of: :repository_list_items
+  belongs_to :repository_column, inverse_of: :repository_list_items
   belongs_to :created_by,
              foreign_key: :created_by_id,
              class_name: 'User'
--- a/app/models/repository_list_item.rb
+++ b/app/models/repository_list_item.rb
@@ -1,6 +1,15 @@
 class RepositoryListItem < ApplicationRecord
-  belongs_to :repository_list_value
+  has_many :repository_list_values,
+           foreign_key: 'selected_item_id'
+  belongs_to :repository, inverse_of: :repository_list_items
+  belongs_to :created_by,
+             foreign_key: :created_by_id,
+             class_name: 'User'
+  belongs_to :last_modified_by,
+             foreign_key: :last_modified_by_id,
+             class_name: 'User'
   validates :name,
             presence: true,
+            uniqueness: { scope: :repository, case_sensitive: false },
             length: { maximum: Constants::TEXT_MAX_LENGTH }
 end
--- a/app/models/repository_list_item.rb
+++ b/app/models/repository_list_item.rb
@@ -10,6 +10,6 @@ class RepositoryListItem < ApplicationRecord
              class_name: 'User'
   validates :data,
             presence: true,
-            uniqueness: { scope: :repository, case_sensitive: false },
+            uniqueness: { scope: :repository_column, case_sensitive: false },
             length: { maximum: Constants::TEXT_MAX_LENGTH }
 end
--- a/app/models/repository_list_item.rb
+++ b/app/models/repository_list_item.rb
@@ -1,0 +1,6 @@
+class RepositoryListItem < ApplicationRecord
+  belongs_to :repository_list_value
+  validates :name,
+            presence: true,
+            length: { maximum: Constants::TEXT_MAX_LENGTH }
+end
--- a/app/models/repository_list_value.rb
+++ b/app/models/repository_list_value.rb
@@ -13,7 +13,7 @@ class RepositoryListValue < ApplicationRecord
 
   validates :repository_cell, presence: true
 
-  def formated
+  def formatted
     item = repository_list_items.find_by_id(selected_item)
     return '' unless item
     item.name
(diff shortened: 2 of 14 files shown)


app/models/repository_list_value.rb
```diff
@@ -0,0 +1,21 @@
...
+  belongs_to :last_modified_by,
+             foreign_key: :last_modified_by_id,
+             class_name: 'User',
+             optional: true
+  has_one :repository_cell, as: :value, dependent: :destroy, inverse_of: :value
+  accepts_nested_attributes_for :repository_cell
+
+  validates :repository_cell, presence: true
+
+  def formated
```
okriuchykhin: Typo, should be formatted

spec/models/repository_list_value_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+        repository_row: repository_row
+      }
+    end
+
+    it 'returns the name of a selected item' do
+      list_item = create :repository_list_item,
+                         repository_list_value: repository_list_value,
+                         name: 'my item'
+      repository_list_value.update_attribute(:selected_item, list_item.id)
+      expect(repository_list_value.reload.formated).to eq 'my item'
```
okriuchykhin: Typo, should be formatted

spec/models/repository_list_value_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+      repository_list_value_two =
+        create :repository_list_value, repository_cell_attributes: {
+          repository_column: repository_column,
+          repository_row: repository_row_two
+        }
+      list_item = create :repository_list_item,
+                         repository_list_value: repository_list_value_two,
+                         name: 'new item'
+      repository_list_value.update_attribute(:selected_item, list_item.id)
+      expect(repository_list_value.reload.formated).to_not eq 'my item'
```
okriuchykhin: Typo, should be formatted

spec/models/repository_list_value_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+      expect(repository_list_value.formated).to eq ''
```
okriuchykhin: Typo, should be formatted

spec/models/repository_list_value_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    end
+
+    it 'returns an empty string if no item selected' do
+      list_item = create :repository_list_item,
+                         repository_list_value: repository_list_value,
+                         name: 'my item'
+      expect(repository_list_value.reload.formated).to eq ''
```
okriuchykhin: Typo, should be formatted

spec/models/repository_list_value_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+    end
+
+    it 'returns an empty string if item does not exists' do
+      repository_list_value.update_attribute(:selected_item, 9999999999)
+      expect(repository_list_value.reload.formated).to eq ''
```
okriuchykhin: Typo, should be formatted

db/migrate/20180207095200_create_repository_list_values.rb
```diff
@@ -0,0 +1,14 @@
+class CreateRepositoryListValues < ActiveRecord::Migration[5.1]
+  def change
+    create_table :repository_list_values do |t|
+      t.bigint :selected_item
+      t.references :created_by,
+                   index: true,
+                   foreign_key: { to_table: :users }
+      t.references :last_modified_by,
+                   index: true,
+                   oreign_key: { to_table: :users }
```
okriuchykhin: typo?

db/migrate/20180207102347_create_repository_list_items.rb
```diff
@@ -0,0 +1,9 @@
+class CreateRepositoryListItems < ActiveRecord::Migration[5.1]
+  def change
+    create_table :repository_list_items do |t|
+      t.references :repository_list_value, foreign_key: true
+      t.text :name, index: true, using: :gin, null: false
+      t.timestamps
+    end
+  end
+end
```
okriuchykhin: May be it makes sense to put these 2 migrations into single file?

db/migrate/20180207095200_create_repository_list_values.rb
okriuchykhin: also it wasn't added to schema, so it needs to be regenerated

okriuchykhin:
In current implementation list item belongs to one list value and if cell is deleted than list item is also gets deleted? Thus each cell should have it's unique set of values? May be it makes sense to make list items reusable at least inside repository?

app/models/repository_list_value.rb
```diff
@@ -0,0 +1,21 @@
+class RepositoryListValue < ApplicationRecord
+  belongs_to :selected_item,
+             foreign_key: :selected_item_id,
+             class_name: 'RepositoryListItem',
+             optional: true
```
okriuchykhin: I would say that there is no sense in existing of cell and value if no item is selected. Tables will display just empty cell if there is no value on specific row and column. So I would remove custom foreign key and use default

app/models/repository_list_value.rb
```diff
@@ -0,0 +1,21 @@
...
+             foreign_key: :last_modified_by_id,
+             class_name: 'User'
+  has_one :repository_cell, as: :value, dependent: :destroy, inverse_of: :value
+  accepts_nested_attributes_for :repository_cell
+
+  validates :repository_cell, presence: true
+
+  def formatted
+    return '' unless selected_item
+    selected_item.name
```
okriuchykhin: I would rename field name to data, to be consistent with other types implementation

okriuchykhin: And I would delegate this field, something like this:
delegate :data, to: selected_item, prefix: false

app/models/repository_list_item.rb
```diff
@@ -0,0 +1,15 @@
+class RepositoryListItem < ApplicationRecord
+  has_many :repository_list_values,
+           foreign_key: 'selected_item_id'
```
okriuchykhin: May be it's needed to add dependent destroy here?

app/models/repository_list_item.rb
```diff
@@ -0,0 +1,15 @@
...
+  belongs_to :repository, inverse_of: :repository_list_items
+  belongs_to :created_by,
+             foreign_key: :created_by_id,
+             class_name: 'User'
+  belongs_to :last_modified_by,
+             foreign_key: :last_modified_by_id,
+             class_name: 'User'
+  validates :name,
```
okriuchykhin: I would rename name to data

app/models/repository_list_item.rb
ZmagoD: probably not

app/models/repository_list_item.rb
ZmagoD: ok

app/models/repository_list_value.rb
ZmagoD: ok

app/models/repository_list_value.rb
ZmagoD: ok

app/models/repository_list_item.rb
okriuchykhin: But since there is no point in existence of repository_list_value if no item is selected we need delete all cells once one item type is deleted

app/models/repository_list_value.rb
ZmagoD: hmm apparently this does not working I don't know the reason but will not waste more time on that

PR #1000, opened 2018-02-15
Fixed big crash that was caused by invalid or null step hash in protocol [SCI-2043]

Zanz2:


app/helpers/protocols_io_helper.rb
```diff
@@ -146,10 +150,33 @@ def close_open_html_tags(text)
     Nokogiri::HTML::DocumentFragment.parse(text).to_html
   end
 
+  def step_hash_null?(step_json)
```
mlorb: You can refactor `step_hash_null` method like is suggested in [this answer](https://stackoverflow.com/questions/4371716/how-to-avoid-nomethoderror-for-missing-elements-in-nested-hashes-without-repeat).

Zanz2: wow, hvala to pa prvič vidim, zgleda zlo uporabno

app/helpers/protocols_io_helper.rb
```diff
@@ -146,10 +150,16 @@ def close_open_html_tags(text)
...
+    return true if step_json.dig(0, 'components', 0, 'component_type_id').nil?
```
mlorb: `return true` and `false` is redundant.
Use only `step_json.dig(0, 'components', 0, 'component_type_id').nil?`

PR #1009, opened 2018-02-27
add info modal for repository items [fixes SCI-2066]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-2066


app/views/repositories/_repository_row_info_modal.html.erb
```diff
@@ -0,0 +1,76 @@
...
+          <span>
+            <%= t "records.modal_info.added_on" %>
+            <%= l @record.created_at, format: :full %>
+          </span>
+          <br>
+          <span>
+            <%= t "records.modal_info.added_by" %>
+            <%= @record.created_by.full_name %>
+          </span>
+          <% @record.repository_cells.each do |record_custom_field| %>
```
mlorb: Nazi comment :smiling_imp:
I think we are not using words like `custom_field`, `record`,... in repositories anymore, can you switch to the new wording:
- `record_custom_field` should be `repository_cell`
- `record` should be `repository_row`
...

PR #1024, opened 2018-03-09
Fix sample record creation [SCI-2145]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-2145


ZmagoD:
Good job! :ship:

PR #1028, opened 2018-03-09
Add file custom column type to db [SCI-2144]

mlorb:
[SCI-2144](https://biosistemika.atlassian.net/browse/SCI-2144)

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -44,7 +44,7 @@ class Asset < ApplicationRecord
   # Asset validation
   # This could cause some problems if you create empty asset and want to
   # assign it to result
-  validate :step_or_result
+  validate :step_or_result_or_repository_cell
 
   belongs_to :created_by,
              foreign_key: 'created_by_id',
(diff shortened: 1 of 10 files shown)


app/models/repository_asset_value.rb
```diff
@@ -0,0 +1,21 @@
...
+  belongs_to :asset,
+             inverse_of: :repository_asset_value,
+             dependent: :destroy
+  has_one :repository_cell, as: :value, dependent: :destroy, inverse_of: :value
+  accepts_nested_attributes_for :repository_cell
+
+  validates :asset, :repository_cell, presence: true
+
+  def formatted
+    asset.file
```
okriuchykhin: maybe 'asset.file_file_name' to return just a name of the file?

app/models/asset.rb
```diff
@@ -44,7 +44,7 @@ class Asset < ApplicationRecord
   # Asset validation
   # This could cause some problems if you create empty asset and want to
   # assign it to result
-  validate :step_or_result
+  validate :step_or_result_or_repository_cell
```
okriuchykhin: I think it's sufficient to check for repository_value

PR #1029, opened 2018-03-10
protocols io hash key instead of index bug in null step checker [SCI-2166]

Zanz2:
protocols io sometimes have hash instead of index, so step null checker now checks for this exception

### DIFF
--- a/app/helpers/protocols_io_helper.rb
+++ b/app/helpers/protocols_io_helper.rb
@@ -165,7 +165,13 @@ module ProtocolsIoHelper
   end
 
   def step_hash_null?(step_json)
-    step_json.dig(0, 'components', 0, 'component_type_id').nil?
+    return false unless step_json.dig(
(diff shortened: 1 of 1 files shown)


app/helpers/protocols_io_helper.rb
```diff
@@ -165,7 +165,13 @@ def close_open_html_tags(text)
...
 
   def step_hash_null?(step_json)
-    step_json.dig(0, 'components', 0, 'component_type_id').nil?
+    return false unless step_json.dig(
+      0, 'components', 0, 'component_type_id'
+    ).nil?
+    return false unless step_json.dig(
+      0, 'components', '0', 'component_type_id'
+    ).nil?
+    true
```
mlorb: You can make it shorter:
`step_json.dig(0, 'components', 0, 'component_type_id').nil? ||`
  `step_json.dig(0, 'components', '0', 'component_type_id').nil?`

PR #1030, opened 2018-03-11
Add annotations migrations and optimize queries [SCI-2016]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-2016

### DIFF
--- a/app/services/tasks/samples_to_repository_migration_service.rb
+++ b/app/services/tasks/samples_to_repository_migration_service.rb
@@ -6,6 +6,8 @@
 
 module Tasks
   module SamplesToRepositoryMigrationService
+    include ActiveRecord::Sanitization::ClassMethods
+
     def self.prepare_repository(team, copy_num = 0)
       repository = Repository.new(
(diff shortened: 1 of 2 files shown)


app/services/tasks/samples_to_repository_migration_service.rb
```diff
@@ -151,5 +182,121 @@ def self.get_custom_columns(team, repository)
...
+      cell_sql = <<-SQL
+        INSERT INTO repository_cells
+          (repository_row_id, repository_column_id, value_id, value_type,
+           created_at, updated_at)
+        VALUES (#{values.join(', ')})
+      SQL
+      ActiveRecord::Base.connection.execute(cell_sql)
+    end
+
+    def self.update_smart_annotations(team, mappings)
```
ZmagoD: Do we need to also update existing reports that have smart annotations inside?

okriuchykhin: I think reports pull actual data from db

PR #1035, opened 2018-03-13
Fix repository export to support file columns [SCI-2079]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-2079


app/models/repository_asset_value.rb
```diff
@@ -18,4 +18,8 @@ class RepositoryAssetValue < ApplicationRecord
   def formatted
     asset.file_file_name
   end
+
+  def data
```
ZmagoD: Do we need this duplication?

okriuchykhin: we use data in datatables

PR #1045, opened 2018-03-19
fixes import protocol bug for guest user [fixes SCI-2197]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-2197

### DIFF
--- a/app/controllers/protocols_controller.rb
+++ b/app/controllers/protocols_controller.rb
@@ -60,8 +60,13 @@ class ProtocolsController < ApplicationController
     copy_to_repository
     copy_to_repository_modal
   )
-  before_action :check_import_permissions, only: [:import]
-  before_action :check_export_permissions, only: [:export]
+  before_action :check_import_permissions, only: :import
+  before_action :check_export_permissions, only: :export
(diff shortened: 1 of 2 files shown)


ZmagoD:
If I check the protocol.io stuff with `check_import_permissions` hook the protocol.io import doesn't work anymore. I don't want to touch this spaghetti there so I added a new before action hook just for protocol.io stuff.

app/controllers/protocols_controller.rb
```diff
@@ -60,8 +60,13 @@ class ProtocolsController < ApplicationController
...
     copy_to_repository_modal
   )
-  before_action :check_import_permissions, only: [:import]
-  before_action :check_export_permissions, only: [:export]
+  before_action :check_import_permissions, only: :import
+  before_action :check_export_permissions, only: :export
+
+  before_action :check_user_role,
+                only: %i(
+                  import protocolsio_import_create protocolsio_import_save
```
mlorb: Then just remove `import` from here. And maybe rename before_action to `check_protocolsio_import_permissions`

PR #1046, opened 2018-03-20
Refactor project navigation [SCI-2120]

mlorb:


app/assets/javascripts/sidebar.js.erb
```diff
@@ -260,10 +269,14 @@ function initializeSidebarToggle() {
     $(".navbar-secondary").addClass("navbar-without-sidebar");
   }
 
-  $("#toggle-sidebar-link").on("click", function() {
+  $(".toggle-sidebar-link").on("click", function(e) {
+    // debugger;
```
Ducz0r: This should go out eh?

Ducz0r:
I have couple of comments - without reviewing the code:
* Can we throw out of the codebase all the CSS animations for the toggling of "old" navigation bar? Or do we need to wait for reports first?
* It has just been decided yesterday that interaction for toggling visibility of navigation sidebar by clicking on the item (e.g. Projects) in left menu bar is not a good one. So it will be prepared by UX designer in the following days. I guess we can keep the current behavior you implemented, and correct it as soon as we get the new specs.

PR #1051, opened 2018-03-22
Fix step deletion permissions [SCI-2237]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-2237


okriuchykhin:
Not a bug

PR #1055, opened 2018-03-25
Skip asset if file is missing in protocol export [SCI-2250]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-2250


matjazhren:
@ZmagoD User does have a feedback in UI that assets are being processed to protocol. Therefore I believe we do not need to add any additional feedback to user (e.g. like hide/disable the export button until the assets are processing).

PR #1056, opened 2018-03-25
Update turbolinks and fix repository table init [SCI-2251]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-2251

### DIFF
--- a/app/assets/javascripts/repositories/repository_datatable.js
+++ b/app/assets/javascripts/repositories/repository_datatable.js
@@ -4,7 +4,7 @@ var RepositoryDatatable = (function(global) {
   'use strict';
 
   var TABLE_ID = '';
-  var TABLE = {};
+  var TABLE = null;
 
   // Extend datatables API with searchable options
   // (http://stackoverflow.com/questions/39912395/datatables-dynamically-set-columns-searchable)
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/repositories/repository_datatable.js
```diff
@@ -4,7 +4,7 @@ var RepositoryDatatable = (function(global) {
   'use strict';
 
   var TABLE_ID = '';
-  var TABLE = {};
+  var TABLE = null;
```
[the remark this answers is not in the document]
ZmagoD: you don't need to assign null you just leave var TABLE; and then it will initialize the variable and assign undefined to it

app/assets/javascripts/repositories/repository_datatable.js
```diff
@@ -1405,7 +1405,10 @@ var RepositoryDatatable = (function(global) {
   }
 
   function destroy() {
-    TABLE = {};
+    if (TABLE !== null) {
```
ZmagoD: and here you can check if(!_.isUndefined(TABLE))

okriuchykhin: isn't comparison to null more clear? Variable should be always defined here I assume

PR #1057, opened 2018-03-25
Fix sample assign permissions [SCI-2249]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-2249

### DIFF
--- a/app/permissions/experiment.rb
+++ b/app/permissions/experiment.rb
@@ -58,6 +58,7 @@ Canaid::Permissions.register_for(MyModule) do
   # permissions
   %i(manage_module
      manage_users_in_module
+     assign_repository_rows_to_module
      assign_sample_to_module
      complete_module
      create_comments_in_module)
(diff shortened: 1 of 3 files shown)


app/permissions/experiment.rb
```diff
@@ -88,12 +89,18 @@
...
-  # module: assign/unassign sample, assign/unassign repository record
+  # module: assign/unassign repository record
   # NOTE: Use 'module_page? &&' before calling this permission!
   can :assign_repository_rows_to_module do |user, my_module|
     user.is_technician_or_higher_of_project?(my_module.experiment.project)
   end
 
+  # module: assign/unassign sample
+  # NOTE: Use 'module_page? &&' before calling this permission!
+  can :assign_sample_to_module do |user, my_module|
```
ZmagoD: Why the duplication of the same permission? Maybe we should rename this to manage_repository_rows or something ask the permission queen @mlorb :smile:

mlorb: I agree. Samples will be removed anyway. We can use `assign_repository_rows_to_module` permission also for samples. There's no need for duplication of code.

okriuchykhin: we will remove all samples permissions once we remove samples

PR #1058, opened 2018-03-26
Implement search in file repository cells and improve performance [SCI-2111]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-2111


app/controllers/search_controller.rb
```diff
@@ -86,6 +88,10 @@ def load_vars
 
   protected
 
+  def generate_search_id
+    SecureRandom.urlsafe_base64(32)
```
ZmagoD: you can use SecureRandom.uuid

PR #1069, opened 2018-04-02
Remotipart workaround [SCI-2270]

Zanz2:
Found a hackish fix using jquery, to upload files without using the remotipart gem, this submits the form as ajax, trough jquery, but overrides normal behaviour of my form element. 
The manual ajax call can submit a .js request, while without the gem, the form was submitting a .html request.

**I didnt know if i should leave the removed remotipart gem and jquery library included in this PR or not, because i did not use the gem at all while working, since i was told that it will soon be removed
I also wasnt sure in which file to put my inline .js code, so i left it in the file.
I was thinking of leaving the .js code in assets->javascript->protocols->index.js , but wasnt sure, since it looked like it was made for the .eln importer **

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -57,7 +57,7 @@ gem 'roo', '~> 2.7.1' # Spreadsheet parser
 gem 'wicked_pdf', '~> 1.1.0'
 gem 'silencer' # Silence certain Rails logs
 gem 'wkhtmltopdf-heroku'
-gem 'remotipart', '~> 1.2' # Async file uploads
+# gem 'remotipart', '~> 1.2' # Async file uploads
 gem 'faker' # Generate fake data
 gem 'auto_strip_attributes', '~> 2.1' # Removes unnecessary whitespaces from ActiveRecord or ActiveModel attributes
 gem 'deface', '~> 1.0'
--- a/Gemfile
+++ b/Gemfile
@@ -57,7 +57,6 @@ gem 'roo', '~> 2.7.1' # Spreadsheet parser
 gem 'wicked_pdf', '~> 1.1.0'
 gem 'silencer' # Silence certain Rails logs
 gem 'wkhtmltopdf-heroku'
-# gem 'remotipart', '~> 1.2' # Async file uploads
 gem 'faker' # Generate fake data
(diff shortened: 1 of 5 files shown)


Zanz2:
I should add that this has been said to cause issues in Internet explorer versions before 9

app/views/protocols/protocolsio_import_create.js.erb
```diff
@@ -13,21 +13,15 @@ $('#pio_no_file_error_span').addClass('has-error').html('<span class="help-block
...
-      <% end %>
-      $('#modal-import-json-protocol-preview').modal('show');
-      $('.modal').on('hidden.bs.modal', function (e) {
-        if($('.modal').hasClass('in')) {
-          $('body').addClass('modal-open');
-        }
-      });
-
-
+  $('#protocolsio-preview-modal-target').html(
```
ZmagoD: I don't know, this is definitely not OK to have js code in view and handle this in that way. At least you can make an endpoint that will return this modal and then initialize it,...

Zanz2: As far as i was aware this is normal for a js response, i made it this way (in a .js.erb file) because of this post below, that said it was expected behaviour. The view is never shown, it is just a script that executes all the code since they are ajax requests. I can change it, but then i have to know what my format.js request calls.
https://stackoverflow.com/questions/13545148/rails-respond-to-format-js-api?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

Gemfile
```diff
@@ -57,7 +57,7 @@ gem 'roo', '~> 2.7.1' # Spreadsheet parser
 gem 'wicked_pdf', '~> 1.1.0'
 gem 'silencer' # Silence certain Rails logs
 gem 'wkhtmltopdf-heroku'
-gem 'remotipart', '~> 1.2' # Async file uploads
+# gem 'remotipart', '~> 1.2' # Async file uploads
```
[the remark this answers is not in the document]
Zanz2: should i also undo the removal of //= require jquery.remotipart in application.js or keep that?

PR #1073, opened 2018-04-04
add ID column to repository [fixes SCI-2206]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-2206

run `RepositoryTableState.destroy_all`

### DIFF
--- a/app/models/concerns/searchable_model.rb
+++ b/app/models/concerns/searchable_model.rb
@@ -68,12 +68,13 @@ module SearchableModel
         end
       else
         unless attrs.empty?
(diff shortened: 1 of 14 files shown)


app/models/concerns/searchable_model.rb
```diff
@@ -68,14 +68,25 @@ module SearchableModel
         end
       else
         unless attrs.empty?
+          # quick fix to enable searching by repositoy_row id
```
mlorb: Me no like this fix :banana:

ZmagoD: the whole model is a big :banana: that needs to be refactored and tested someday

PR #1086, opened 2018-04-11
enables copying of repository items [fixes SCI-2207]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-2207

### DIFF
--- a/app/services/repository_actions/duplicate_cell.rb
+++ b/app/services/repository_actions/duplicate_cell.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 module RepositoryActions
-  class RepositoryCellResolver
+  class DuplicateCell
     def initialize(cell, new_row, team)
       @cell    = cell
       @new_row = new_row
@@ -9,21 +9,12 @@ module RepositoryActions
     end
 
     def call
-      case @cell.value_type
-      when 'RepositoryListValue'
-        clone_repository_list_value
-      when 'RepositoryTextValue'
-        clone_repository_text_value
-      when 'RepositoryAssetValue'
-        clone_repository_asset_value
-      when 'RepositoryDateValue'
-        clone_repository_date_value
-      end
+      self.send("duplicate_#{@cell.value_type.underscore}")
     end
 
     private
 
-    def clone_repository_list_value
+    def duplicate_repository_list_value
       old_value = @cell.value
       RepositoryListValue.create(
         old_value.attributes.merge(id: nil,
@@ -34,7 +25,7 @@ module RepositoryActions
       )
     end
 
-    def clone_repository_text_value
+    def duplicate_repository_text_value
       old_value = @cell.value
       RepositoryTextValue.create(
         old_value.attributes.merge(id: nil,
@@ -45,7 +36,7 @@ module RepositoryActions
       )
     end
 
-    def clone_repository_asset_value
+    def duplicate_repository_asset_value
       old_value = @cell.value
       new_asset = create_new_asset(old_value.asset)
       RepositoryAssetValue.create(
@@ -60,7 +51,7 @@ module RepositoryActions
       )
     end
 
-    def clone_repository_date_value
+    def duplicate_repository_date_value
       old_value = @cell.value
       RepositoryDateValue.create(
         old_value.attributes.merge(id: nil,
--- a/app/services/repository_actions/duplicate_cell.rb
+++ b/app/services/repository_actions/duplicate_cell.rb
@@ -2,11 +2,11 @@
 
 module RepositoryActions
   class DuplicateCell
-    def initialize(cell, new_row, user)
+    def initialize(cell, new_row, user, team)
       @cell    = cell
       @new_row = new_row
       @user    = user
-      @team    = user.current_team
+      @team    = team
     end
 
     def call
--- a/app/services/repository_actions/duplicate_cell.rb
+++ b/app/services/repository_actions/duplicate_cell.rb
@@ -2,10 +2,11 @@
 
 module RepositoryActions
   class DuplicateCell
-    def initialize(cell, new_row, team)
+    def initialize(cell, new_row, user)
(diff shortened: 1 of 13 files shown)


app/services/repository_actions/repository_cell_resolver.rb
```diff
@@ -0,0 +1,113 @@
...
+module RepositoryActions
+  class RepositoryCellResolver
+    def initialize(cell, new_row, team)
+      @cell    = cell
+      @new_row = new_row
+      @team    = team
+    end
+
+    def call
+      case @cell.value_type
```
Ducz0r: You could probably omit this `case` statement with some ruby metaprogramming (generate String name of the function based on the `@cell.value_type`, and eval it). But I'm not sure if that is good or bad practice 😄.

app/services/repository_actions/repository_cell_resolver.rb
```diff
@@ -0,0 +1,113 @@
...
+      when 'RepositoryDateValue'
+        clone_repository_date_value
+      end
+    end
+
+    private
+
+    def clone_repository_list_value
+      old_value = @cell.value
+      RepositoryListValue.create(
```
Ducz0r: I think that `created_by` and `last_modified_by` should be set to the user who is performing the copying of the row. So maybe the whole `RepositoryCellResolver` needs additional parameter, `user`.

app/services/repository_actions/repository_cell_resolver.rb
```diff
@@ -0,0 +1,113 @@
...
+          repository_cell_attributes: {
+            repository_row: @new_row,
+            repository_column: @cell.repository_column
+          })
+      )
+    end
+
+    def clone_repository_text_value
+      old_value = @cell.value
+      RepositoryTextValue.create(
```
Ducz0r: I think that `created_by` and `last_modified_by` should be set to the user who is performing the copying of the row.

app/services/repository_actions/repository_cell_resolver.rb
```diff
@@ -0,0 +1,113 @@
...
+            repository_row: @new_row,
+            repository_column: @cell.repository_column
+          })
+      )
+    end
+
+    def clone_repository_asset_value
+      old_value = @cell.value
+      new_asset = create_new_asset(old_value.asset)
+      RepositoryAssetValue.create(
```
Ducz0r: I think that `created_by` and `last_modified_by` should be set to the user who is performing the copying of the row.

app/services/repository_actions/repository_cell_resolver.rb
```diff
@@ -0,0 +1,113 @@
...
+      )
+    end
+
+    # reuses the same code we have in copy protocols action
+    def create_new_asset(old_asset)
+      new_asset = Asset.new_empty(
+        old_asset.file_file_name,
+        old_asset.file_file_size
+      )
+      new_asset.created_by = old_asset.created_by
```
Ducz0r: I think that `created_by` and `last_modified_by` should be set to the user who is performing the copying of the row.

app/services/repository_actions/repository_cell_resolver.rb
```diff
@@ -0,0 +1,113 @@
...
+      new_asset.team = old_asset.team
```
Ducz0r: Also, team should probably set to `@team`? I mean it shouldn't matter, but maybe it's more safe this way.

app/services/repository_actions/repository_cell_resolver.rb
ZmagoD: The copied items have all columns (including hidden ones) copied from the original repository item. The only differences
are the following:
 Newly copied item is not assigned to any task, regardless of the original item’s task assignments.
 Newly copied item has a name “<original_item_name> (1)” – the (1) is appended to the end of the name.
 Newly copied item’s “added on” field is set onto the time the item has been copied.
 Newly copied item’s “added by” field is set to the user who performed the copying.
After copying is done, repository table is reloaded and a flash message appears, notifying the user that “<n> item(s)
successfully copied.” (see Figure 10: Copying repository items).

ZmagoD: :arrow_up: from specs

app/services/repository_actions/repository_cell_resolver.rb
ZmagoD: hmm I don't know you can copy only items in the same repository and that repository belongs to a team so it will be a bit odd that repository items will belong to other team

ZmagoD: but other way it should be OK one way or another

app/services/repository_actions/repository_cell_resolver.rb
Ducz0r: I know this is not specified concretely in the specifications, but don't you agree it should be same as the row's `created_by`?

To me, it doesn't make sense if row's `created_by` is one thing, and then its cells - which are totally dependant on this row for existence - don't have the same `created_by` attribute. 

So please fix this nonetheless.

app/services/repository_actions/duplicate_cell.rb
```diff
@@ -0,0 +1,110 @@
+# frozen_string_literal: true
+
+module RepositoryActions
+  class DuplicateCell
+    def initialize(cell, new_row, user)
+      @cell    = cell
+      @new_row = new_row
+      @user    = user
+      @team    = user.current_team
```
Ducz0r: @ZmagoD sorry for being pain in the neck. I still have one comment - I wouldn't count on `user.current_team` to be the correct team into which to clone the data. If we are certain that `@new_row` is already persisted, we should retrieve the `@team` from this new row (`@new_row.repository.team`), not from the user.

I know it might be more DB querying than checking the user's current team, but it is definitely more reusable if we will ever call this service from some other piece of code in the future (e.g. from the API, where user's current team might be anything).

PR #1094, opened 2018-04-13
implement file column type when editing & creating new rows... [SCI-2075]

mlorb:
...in inventories DataTables 
[SCI-2075](https://biosistemika.atlassian.net/browse/SCI-2075)

### DIFF
--- a/app/controllers/repository_rows_controller.rb
+++ b/app/controllers/repository_rows_controller.rb
@@ -34,62 +34,7 @@ class RepositoryRowsController < ApplicationController
       errors[:default_fields] = record.errors.messages unless record.save
       if cell_params
         cell_params.each do |key, value|
-          column = @repository.repository_columns.detect do |c|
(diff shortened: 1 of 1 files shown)


app/controllers/repository_rows_controller.rb
```diff
@@ -51,6 +51,27 @@ def create
...
+                              last_modified_by: current_user,
+                              team: current_team)
+            if asset.save
+              asset.post_process_file(current_team)
+            else
+              errors[:repository_cells] << {
+                "#{column.id}": { data: asset.errors.messages[:file].first }
+              }
+            end
+            cell_value = RepositoryAssetValue.new(
```
ZmagoD: For this one and the others cell_value stuff we can extract in a method that creates a new object of whatever type. We duplicate this in a lot of places. I wonder if repository object design pattern would help us here.

PR #1101, opened 2018-04-18
adds new reports index page [fixes SCI-2124]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-2124


spec/models/views/datatables/datatables_report_spec.rb
```diff
@@ -0,0 +1,80 @@
+require 'rails_helper'
+
```
mlorb: Nice selection of tests :telescope:

PR #1111, opened 2018-04-20
Update the code for repository table states so it (hopefully) causes less errors [SCI-2276]

Ducz0r:
Refactor the code into something more manage-able. Hopefully.

I'd just double-check with the reviewer because I'm not an expert on assets pipeline: if I'm using `default_table_order_as_js_array` and `default_table_columns` methods inside the `.js.erb` file, will this cause the JavaScript file to be always precompiled on each request? The methods should always return the same value - because it's calculated from the `constants.rb`, but I'm not sure if I'm enforcing some kind of performance hump because of this.

Closes [SCI-2276](https://biosistemika.atlassian.net/browse/SCI-2276).


ZmagoD:
Regarding the assets pipeline, as far as I know, it compiles the assets in static files and they are not updated in production mode until you reboot the server.

ZmagoD:
Looks OK :ship:

PR #1112, opened 2018-04-23
propagate assigned repository items in workflow [fixes SCI-2261]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-2261


app/controllers/my_modules_controller.rb
```diff
@@ -450,6 +467,13 @@ def unassign_repository_records
       end
 
       @my_module.repository_rows.destroy(records & @my_module.repository_rows)
+
+      if downstream
```
Ducz0r: Same as with assign: this code should be rewritten because right now, if there is a row that is not assigned to the current task, it will also get skipped from downstream tasks.

By the way, the `&` operator in `records & my_module.repository_rows` simply performs the intersection of both arrays.

app/controllers/my_modules_controller.rb
```diff
@@ -391,18 +395,30 @@ def repository_index
   def assign_repository_records
     if params[:selected_rows].present? && params[:repository_id].present?
       records_names = []
+      downstream = ActiveModel::Type::Boolean.new.cast(params[:downstream])
 
       params[:selected_rows].each do |id|
```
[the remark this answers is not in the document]
ZmagoD: should we open a new task for refactor that?

app/controllers/my_modules_controller.rb
```diff
@@ -455,25 +462,24 @@ def assign_repository_records
...
       if downstream
         @my_module.downstream_modules.each do |my_module|
+          records = RepositoryRow.assigned_on_my_module(params[:selected_rows],
+                                                        my_module)
           my_module.repository_rows.destroy(records & my_module.repository_rows)
         end
       end
 
+      # update last last_modified_by
+      records.update_all(last_modified_by_id: current_user.id)
```
Ducz0r: Hmm, this `records` variable will probably hold the latest `records` collection from the last downstream module. So I doubt it's the right array of records to be used for setting `last_modified_by_id`.

Probably, you need to somehow have a collection/hash/map of all destroyed repository rows (both in initial module/task, as well as in downstream modules/tasks), and update all such records.

PR #1118, opened 2018-04-24
sign_out user after 7 days [fixes SCI-2323]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-2323

Here I used a custom before_action hook that checks the user last sign_in timestamp. We can't use Devise Timeoutable module here because it checks the idle time, not the last login time: http://www.rubydoc.info/github/plataformatec/devise/master/Devise/Models/Timeoutable

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -5,7 +5,7 @@ class User < ApplicationRecord
   acts_as_token_authenticatable
   devise :invitable, :confirmable, :database_authenticatable, :registerable,
          :async, :recoverable, :rememberable, :trackable, :validatable,
-         :rememberable, :timeoutable, :omniauthable,
+         :timeoutable, :omniauthable,
          omniauth_providers: Extends::OMNIAUTH_PROVIDERS,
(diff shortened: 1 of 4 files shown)


okriuchykhin:
Looks like this is built-in feature of Device to achieve required behavior http://www.rubydoc.info/github/plataformatec/devise/master/Devise/Models/Rememberable

We can use it together with Timeoutable module, setting let's say inactivity period for 1 hour if user didn't check 'Remember me' checkbox and limit remember time by 1 week.

app/models/user.rb
```diff
@@ -5,7 +5,8 @@ class User < ApplicationRecord
   acts_as_token_authenticatable
   devise :invitable, :confirmable, :database_authenticatable, :registerable,
          :async, :recoverable, :rememberable, :trackable, :validatable,
-         :omniauthable, omniauth_providers: Extends::OMNIAUTH_PROVIDERS,
+         :rememberable, :timeoutable, :omniauthable,
```
okriuchykhin: rememberable is already at line 7

PR #1124, opened 2018-05-03
fix the missing icons on task card [fixes SCI-2347]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-2347


ZmagoD:
do you think we use this css only for task card?

mlorb:
I think, yes. But even if we don't, it's still good that we use primary color for active states throughout the app.

PR #1127, opened 2018-05-04
Zd sci 2307 v3

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-2307


spec/services/smart_annotations/html_preview_spec.rb
```diff
@@ -52,12 +52,47 @@
...
+
+    describe 'repository name with three words' do
+      it 'returns the first letter of first 3 words capitalized' do
+        trimmed_repository_name = subject.__send__(
+          :trim_repository_name, 'two words of wisdom'
+        )
+        expect(trimmed_repository_name).to eq('Two')
+      end
+
+      it 'does not return 3 letters of the first word' do
```
Ducz0r: This block is totally duplicated from above, it's probably not needed?

PR #1151, opened 2018-05-18
shorten project name [fixes SCI-2344]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-2344


app/controllers/reports_controller.rb
```diff
@@ -485,7 +485,7 @@ def load_visible_projects
                                     .select(:id, :name)
     @visible_projects = projects.collect do |project|
       VisibleProject.new(new_project_reports_path(project),
-                         ellipsize(project.name, 75, 50))
+                         ellipsize(project.name, 50, 40))
```
mlorb: Shouldn't we use constants? ;)

ZmagoD: no

PR #1164, opened 2018-05-24
adds discard gem, move deletion of repositories in a rake task to run…

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-2448

### DIFF
--- a/db/migrate/20180518113526_remove_repository_rows_and_columns_without_repository.rb
+++ b/db/migrate/20180518113526_remove_repository_rows_and_columns_without_repository.rb
@@ -1,7 +1,9 @@
 class RemoveRepositoryRowsAndColumnsWithoutRepository < ActiveRecord::Migration[5.1]
   def up
-    repository_ids = Repository.select(:id)
-    RepositoryRow.where.not(repository_id: repository_ids).delete_all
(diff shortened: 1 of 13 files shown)


db/migrate/20180518113526_remove_repository_rows_and_columns_without_repository.rb
```diff
@@ -1,7 +1,9 @@
 class RemoveRepositoryRowsAndColumnsWithoutRepository < ActiveRecord::Migration[5.1]
```
okriuchykhin: Probably we can get rid of it

lib/tasks/clear_repository_rows_and_columns_without_repository.rake
```diff
@@ -0,0 +1,9 @@
+namespace :clear_repository_rows_and_columns_without_repository do
+  desc 'Removes all repository rows/columns and ' \
+       'referenced entities with dependent: destroy'
+  task run: :environment do
+    repository_ids = Repository.select(:id)
+    RepositoryRow.where.not(repository_id: repository_ids).delete_all
```
okriuchykhin: Probably destroy_all will work better and delete all dependents

PR #1166, opened 2018-05-24
Fix WOPI controller to support repositories [SCI-2342]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-2342


app/controllers/wopi_controller.rb
```diff
@@ -245,15 +245,20 @@ def load_vars
       logger.warn 'Found asset: ' + @asset.id.to_s
       step_assoc = @asset.step
       result_assoc = @asset.result
+      repository_cell_assoc = @asset.repository_cell
       @assoc = step_assoc unless step_assoc.nil?
```
ZmagoD: probably we can use ||= here instead of checking for nil?

PR #1188, opened 2018-06-01
Improve sidebar rendering performance [SCI-2478]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-2478

### DIFF
--- a/app/views/shared/sidebar/_projects.html.erb
+++ b/app/views/shared/sidebar/_projects.html.erb
@@ -1,9 +1,7 @@
 <ul>
   <% if @projects_tree.present? then %>
     <% @projects_tree.each do |project| %>
-      <% if (project_page? ||
-             sample_types_page_project? ||
-             sample_groups_page_project?) && project == @project %>
+      <% if project_page? && project == @project %>
         <li class="active" data-parent="candidate">
           <span class="tree-link line-wrap no-indent">
             <i class="no-arrow"></i>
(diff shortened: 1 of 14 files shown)


config/environments/development.rb
```diff
@@ -75,7 +75,7 @@
   config.assets.raise_runtime_errors = true
 
   # Only log info and higher on development
-  config.log_level = :info
+  config.log_level = :debug
```
Ducz0r: I also use this locally a lot, but I'm not sure if this is needed to be changed?

app/views/shared/sidebar/_projects.html.erb
```diff
@@ -1,35 +1,31 @@
...
-            </span>
-            <%= render partial: 'shared/sidebar/experiments',
-                      locals: { project: project } %>
-          </li>
-        <% end %>
+  <% if @projects_tree.present? then %>
+    <% @projects_tree.each do |project| %>
+      <% if (project_page? ||
+             sample_types_page_project? ||
+             sample_groups_page_project?) && project == @project %>
```
Ducz0r: Probably we can get rid of these 2 checks: `sample_types_page_project?`, `sample_groups_page_project?`, right?

config/environments/development.rb
okriuchykhin: I don't know, it's default value in Rails 5

app/views/shared/sidebar/_projects.html.erb
okriuchykhin: Done

Ducz0r:
Great, now it works perfectly :+1:!

PR #1192, opened 2018-06-05
removes visibility toggle from Name column [fixes SCI-2435]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-2435


Ducz0r:
Yeah, we should run Rake task this with the next release, I'll attach it to the release documentation.

I have another question since you dragged me into this - if I undestand this issue correctly, we might have an invalid state in the database - users, until this is merged/fixed - were able to save repository rows without a name, right? Which is problematic, because we shouldn't allow such repository rows to be saved in DB.

So we probably need to also write another Rake task that would name all repository rows without a name to something (e.g. `Item` or something generic). What do you two think?

ZmagoD:
the repository item name is a string 'undefined' so the state in the database is valid

Ducz0r:
Okay, great :+1:!

PR #1209, opened 2018-06-13
Update team migration script to support "list of values" and "file" repository columns/cells [fixes SCI-2165]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-2165

### DIFF
--- a/app/services/team_importer.rb
+++ b/app/services/team_importer.rb
@@ -190,7 +190,7 @@ class TeamImporter
   def update_annotation(text)
     return false if text.nil?
     updated = false
-    %w(prj exp tsk).each do |name|
+    %w(prj exp tsk rep_item).each do |name|
       text.scan(/~#{name}~\w+\]/).each do |text_match|
         orig_id_encoded = text_match.match(/~#{name}~(\w+)\]/)[1]
         orig_id = orig_id_encoded.base62_decode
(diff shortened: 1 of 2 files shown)


ZmagoD:
Smart annotations are handled in the team_import script.
Regarding S3_BUCKET probably we should use ENV['S3_BUCKET'] there may be @okriuchykhin has a comment on it.

ZmagoD:
https://github.com/biosistemika/scinote-web/blob/master/app/services/team_importer.rb#L151

Ducz0r:
Okay, let me be more precise: users can now reference individual repository items (this was previously **not possible**) using smart annotations. This results in a smart annotation that looks like this: `[#ILT/1~rep_item~1]`.

And these smart annotations for repository items are not updated when importing the data, they are left as it is, and this results in them pointing to either `nil` repository items, or possibly to items in other teams to which user doesn't have rights.

ZmagoD:
Aha OK, I will fix also that.

okriuchykhin:
S3_BUCKET is also ruby constant representing aws sdk s3 resource.

ZmagoD:
@okriuchykhin so here we can just check if ENV['S3_BUCKET'] ?

app/services/team_importer.rb
```diff
@@ -202,7 +190,7 @@ def update_smart_annotations(team)
   def update_annotation(text)
     return false if text.nil?
     updated = false
-    %w(prj exp tsk sam).each do |name|
+    %w(prj exp tsk).each do |name|
```
Ducz0r: You also need to include `rep_item` here; then the import should finally work as intended!

PR #1224, opened 2018-06-26
fixes setting sidebar teams link[fixes SCI-2429]

ZmagoD:
https://biosistemika.atlassian.net/browse/SCI-2429

### DIFF
--- a/app/views/users/settings/_sidebar.html.erb
+++ b/app/views/users/settings/_sidebar.html.erb
@@ -63,7 +63,12 @@
        <li class="<%= 'active' if on_settings_team_page? %>" >
          <span class="tree-link line-wrap no-indent">
            <% if on_settings_team_page? %>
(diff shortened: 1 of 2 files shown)


app/views/users/settings/_sidebar.html.erb
```diff
@@ -63,7 +63,12 @@
        <li class="<%= 'active' if on_settings_team_page? %>" >
          <span class="tree-link line-wrap no-indent">
            <% if on_settings_team_page? %>
```
mlorb: Do we still need this `if... else` sentence, or can we merge this together?

ZmagoD: I have no idea.

mlorb: You can give it a try :smiley_cat:

ZmagoD: it has to be like this

PR #1231, opened 2018-07-04
Tables sum error (tables now use big.js for +-/* calculations) [SCI-2587]

Zanz2:


Zanz2:
Try catch statements would be slow if the exception would be caught alot, but the only time an exception happens when i tested, is with huge numbers (10 to the power of 10000000 etc), thats when big.js freezes, so for that case i made the catch statement calculate it regularly, which will just return infinite.

vendor/assets/javascripts/ruleJS.js
```diff
@@ -1035,21 +1035,64 @@ var ruleJS = (function (root) {
...
+          } catch(e) {
+            result = number1 - number2
+          }
+          //MODIFICATION END
+
+          //ORIGINAL: result = number1 - number2;
           break;
         case '/':
-          result = number1 / number2;
+
```
Ducz0r: Get rid of whitespace here.

vendor/assets/javascripts/ruleJS.js
```diff
@@ -1035,21 +1035,64 @@ var ruleJS = (function (root) {
...
           if (result == Infinity) {
             throw Error('DIV_ZERO');
           } else if (isNaN(result)) {
             throw Error('VALUE');
           }
           break;
         case '*':
-          result = number1 * number2;
+
```
Ducz0r: Get rid of whitespace here.

PR #1248, opened 2018-07-17
Fixes duplicate checkbox entries in activity log [SCI-1621]

Zanz2:
The issue was that creating/updating/deleting steps and checkboxes caused the onclick jquery listeners to be reapplied ontop of the old ones, thus too many activities got logged, now it removes all listeners before applying new ones to the click .


okriuchykhin:
Might be fixed with https://github.com/biosistemika/scinote-web/pull/1251
Please re-check this bug after that PR is merged.

PR #1249, opened 2018-07-17
Fixes tinymce rtf editor not displaying (updated gem) [SCI-2305]

Zanz2:
Updated tinymce gem to 4.7.11


okriuchykhin:
Can you please update to the latest minor version? I think it is .13

PR #1266, opened 2018-08-13
Refactoring, deleted samples and samples related code [SCI-2228]

Zanz2:


app/controllers/my_modules_controller.rb
```diff
@@ -22,30 +20,24 @@ class MyModulesController < ApplicationController
...
                          assign_repository_records
-                         unassign_repository_records
-                         assign_samples
-                         unassign_samples)
+                         unassign_repository_records)
 
   layout 'fluid'.freeze
 
   # Define submit actions constants (used in routing)
   ASSIGN_SAMPLES = 'Assign'.freeze
```
Ducz0r: This can also be removed I think?

Ducz0r:
⚠️ And my next *todo*-list for the next review, after my previous issues will be resolved:

* Decide whether to remove this or not: `spec/lib/tasks/samples_to_repository_migration_spec.rb`, `spec/services/tasks/samples_to_repository_migration_service_spec.rb`
* Make sure `db/schema.rb` is valid

PR #1267, opened 2018-08-14
Changed some icons to fontawesome [SCI-2445]

Zanz2:


app/views/protocols/index.html.erb
```diff
@@ -70,9 +70,9 @@
                   </a>
                 </li>
                 <li>
-                <%= link_to "#modal-import-json-protocol", data: { toggle: 'modal', remote: true } do %>
-                  <span class="fas fa-file"></span>
-                  <span class="hidden-xs"><%= t("protocols.index.import_json") %></span>
+                <%= link_to "#modal-import-json-protocol", "data-toggle" => "modal" do %>
```
okriuchykhin: Can you please change it to new hash syntax and fix indent below

okriuchykhin: data: { toggle: 'modal' }

PR #1273, opened 2018-08-21
Added toggle checkbox for tooltips help in prefferences [SCI-2663]

Zanz2:


config/locales/en.yml
```diff
@@ -1801,6 +1801,11 @@ en:
 
   notifications:
     title: "Notifications"
+    tips:
+      title: "Help Tips"
+      text1: "When you place your mouse over specific element of SciNote, we show you a tip with description of this element."
+      text2: "You may want to turn that off if you feel confident in here."
+      toggle: "Show Help Tips NO/YES"
```
Ducz0r: This text should just say `"Show Help Tips"`.

config/locales/en.yml
```diff
@@ -1801,6 +1801,11 @@ en:
 
   notifications:
     title: "Notifications"
```
Ducz0r: I prefer to use the detailed settings path for these locales, e.g. `users.settings.account.preferences.edit.tips`. Tips are **not** part of notifications, but a separate section of settings.

app/views/users/settings/account/preferences/index.html.erb
```diff
@@ -133,6 +133,27 @@
               </div>
             </div>
           </div>
+          <br>
```
Ducz0r: I would move this out of the `notifications` segment - one level up, to be on the same level as notifications & time zone containers.

app/models/user.rb
```diff
@@ -43,7 +43,8 @@ class User < ApplicationRecord
       assignments_email: false,
       recent: true,
       recent_email: false,
-      system_message_email: false
+      system_message_email: false,
+      popover_v2_toggle: true
```
Ducz0r: First of all, why `v2`? I'd name this attribute in user preferences simply `popovers_enabled`. Secondly, I'd move it onto the first level, as `popovers` are **NOT** belonging under notifications.
E.g.
```ruby
default_settings(
    time_zone: 'UTC',
    notifications_settings: {
      assignments: true,
      assignments_email: false,
      recent: true,
      recent_email: false,
      system_message_email: false
    },
    popovers_enabled: true
  )
```

app/models/user.rb
```diff
@@ -456,7 +457,8 @@ def has_linked_account?(provider)
   NOTIFICATIONS_TYPES = %w(assignments_notification recent_notification
                            assignments_email_notification
                            recent_email_notification
-                           system_message_email_notification)
+                           system_message_email_notification
+                           popover_v2_toggle )
```
Ducz0r: I don't think we should modify `NOTIFICATIONS_TYPES`.

app/controllers/client_api/users/users_controller.rb
```diff
@@ -78,7 +78,8 @@ def user_params
                       :time_zone, :assignments_notification,
                       :assignments_email_notification, :recent_notification,
                       :recent_email_notification,
-                      :system_message_email_notification)
+                      :system_message_email_notification,
+                      :popover_v2_toggle)
```
Ducz0r: Again, change this name to e.g. `popovers_enabled`.

app/assets/javascripts/users/settings/account/preferences/index.js
```diff
@@ -71,7 +71,8 @@
   // Setup notification checkbox buttons
   function notificationsSettings() {
     var notification_settings = [ "recent_notification",
-                                  "assignments_notification" ]
+                                  "assignments_notification",
+                                  "popover_v2_toggle" ]
```
Ducz0r: Just from code structure point of view, I'd move this out of `notificationSettings()`, but into its own function. And I'd rename `popover_v2_toggle` to `popovers_enabled`.

app/models/user.rb
Zanz2: i named them that just because popovers exist already, and so that when debbugging it would be clear these are two seperate implementations, but i aggree popovers_enabled is better

app/assets/javascripts/users/settings/account/preferences/index.js
Zanz2: Hmm, the way im thinking about implementing its own settings function, it would require code duplication in index.js, the function for popovers_enabled would do the excact thing as the notifications function, but it would have a few different names, the new function would still read from the same place in prefferences controller as the old one does. Its around 50 lines of code duplicated

app/models/user.rb
Zanz2: if I remove it from here, it doesn work, since the user doesnt have this user.popovers_enabled attribute anymore.

app/assets/javascripts/users/settings/account/preferences/index.js
Zanz2: I can do it, but it would be alot more code, maybe i can add comments to old used function?

Ducz0r:
@Zanz2 I've pushed some fixes of my own of how I think this should be implemented. Can you check it out and test it, so we're sure it works?

Zanz2:
Tested it, your changes all work!

PR #1274, opened 2018-08-21
Added Tooltips (bootstrap popovers) for various elements in scinote [SCI-2662]

Zanz2:


Zanz2:
I am fixing hound

app/assets/javascripts/sitewide/popover_helper.js
```diff
@@ -0,0 +1,75 @@
...
+                var top = $(obj).offset().top;
+                $('.popover_'+i+'_window').css({
+                    top: (top) + 'px'
+                });
+                $(".popover_"+i+"_window").on("mouseleave", function () {
+                  $(obj).popover('hide');
+                  $(".popover_"+i+"_window").css("background-color","#F0F0F0");
+                });
+                $(".popover_"+i+"_window").on("mouseenter", function () {
+                  $(".popover_"+i+"_window").css("background-color","#DADADA");
```
Ducz0r: Also, regarding the styling when hovering over the popover, can you just style all this stuff via CSS, not using JavaScript?

app/assets/stylesheets/constants.scss
```diff
@@ -12,6 +12,10 @@ $color-gainsboro: #e3e3e3;
 $color-alto: #d2d2d2;
 $color-silver: #c5c5c5;
 $color-silver-chalice: #a0a0a0;
+$color-lightsilver: #f0f0f0;
```
Ducz0r: Regarding these 4 colors, are they copied from the specs epic? I'd like to include @mlorb in this debate - during last UX revamp, we've intentionally got rid of ~10+ grayscale variations due to not really making much of a difference, and this change here is introducing additional grayscale levels again. Do we really need to add these 4 colors, or can we reuse some of the existing grayscale colors?

mlorb: That's correct. We shouldn't be introducing new base colors, because be want to remain clean style guide.
My suggestions for replacing the colors:
- instead of `$color-lightsilver` use `$color-concrete`
- instead of `$color-darksilver` use `$color-alto`
- instead of `$color-lightgray` use `$color-gainsboro`
- instead of `$color-darkgray` use `$color-dove-gray`

And if these replacements are not good enough, then you can fine tune them in `app/assets/stylesheets/themes/scinote.scss` file like that:
`.popover-footer {`
`    background-color: lighten($color-gainsboro, 2%);`
`    a {`
`      color: lighten($color-dove-gray, 3%);`
`    }`
`}`

Zanz2: yeah, these colors were specified in epic, thats why i added them

PR #1284, opened 2018-08-27
Search page results from different teams than active not switching active team on click [SCI-2684]

Zanz2:
Its more of a hackish fix, but this is what i came up with, since opening the link_to in a new tab would for some reason cause a GET request, not a POST one, like i specified in the link_to constructor.


Zanz2:
Found a more elegant solution with Luka, closing this and will make new PR

PR #1288, opened 2018-08-30
Refactored .js.erb -> .js using gon gem [SCI-2620]

Zanz2:

PR #1289, opened 2018-08-31
Fixed bug with table name of step not being length validated [SCI-2719]

Zanz2:


Zanz2:
this implementation now uses gon gem, but the gon gem branch is not implemented in master yet

app/assets/javascripts/protocols/steps.js.erb
```diff
@@ -545,11 +545,29 @@
     var $tinyMCEInput = TinyMCE.getContent();
     var descriptionValid = textValidator(ev, $descrTextarea, 0,
       <%= Constants::TEXT_MAX_LENGTH %>, false, $tinyMCEInput);
-
+    var tableNamesValidArray = [];
+    // iterate over table titles and validate their length
+    $form.find(".table_name").each(function(index , tableName) {
+      if (tableName.value.length > 0) {
+        tableNamesValidArray[index] = textValidator(ev, tableName, 1,
+          <%= Constants::NAME_MAX_LENGTH %> );
```
Zanz2: constants are still used here, because this will get merge before js.erb issue #1288

PR #1291, opened 2018-09-07
Sci 2696

mikechea:
Close SCI-2696


spec/requests/api/v1/inventory_items_controller_spec.rb
```diff
@@ -126,4 +139,67 @@
       expect(hash_body).to match({})
     end
   end
+
+  describe 'POST inventory_item, #create' do
+    before :all do
+      @valid_headers['Content-Type'] = 'application/json'
```
okriuchykhin: I think you can move it to main @valid_headers definition

PR #1292, opened 2018-09-07
Created API endpoints for experiments and related data [SCI-2666]

Zanz2:
Created api endpoints for: experiments, my modules (tasks) , my module groups (task groups), connections ( input and output connections from tasks on canvas in experiments)
The new query routes are: ( 
api/v1/teams/1/projects/1/experiments
.../experiments/1
.../experiments/1/tasks
.../experiments/1/tasks/1
.../experiments/1/task_groups
.../experiments/1/task_groups/1
.../experiments/1/connections/
.../experiments/1/connections/1

### DIFF
--- a/app/controllers/api/v1/my_module_groups_controller.rb
+++ b/app/controllers/api/v1/my_module_groups_controller.rb
@@ -10,10 +10,11 @@ module Api
 
       def index
         my_module_groups = @experiment.my_module_groups
-                              .page(params.dig(:page, :number))
-                              .per(params.dig(:page, :size))
+                                      .page(params.dig(:page, :number))
+                                      .per(params.dig(:page, :size))
 
-        render jsonapi: my_module_groups, each_serializer: MyModuleGroupSerializer
+        render jsonapi: my_module_groups,
+          each_serializer: MyModuleGroupSerializer
       end
 
       def show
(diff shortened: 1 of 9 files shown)


app/serializers/api/v1/connection_serializer.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class ConnectionSerializer < ActiveModel::Serializer
+      type :connections
+      attributes :id, :input_id, :output_id
```
okriuchykhin: May be it makes sense instead just returning input_id and output_id as attributes, return those connected tasks as relationships?

app/controllers/api/v1/my_module_groups_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+        render jsonapi: {}, status: :forbidden unless can_read_experiment?(
+          @experiment
+        )
+      end
+
+      def load_task_groups
+        @my_module_group = @experiment.my_module_groups.find(
+          params.require(:id)
+        )
+        render jsonapi: {}, status: :not_found if @my_module_group.nil?
```
okriuchykhin: Here @my_module_group can't be nil as at previous line find will throw an exception if module_group is not found

okriuchykhin: And I think method should be named just load_task_group as you load only one

Zanz2:
I made changes according to the pull request review

PR #1293, opened 2018-09-07
close SCI-2698; unit tests for DELETE inventory item [SCI-2698]

mikechea:
Provided unit tests for DELETE inventory item

### DIFF
--- a/spec/requests/api/v1/inventory_items_controller_spec.rb
+++ b/spec/requests/api/v1/inventory_items_controller_spec.rb
@@ -161,7 +161,7 @@ RSpec.describe 'Api::V1::InventoryItemsController', type: :request do
       expect(RepositoryRow.where(id: deleted_id)).to exist
     end
 
-    it 'When invalid request, repository from another team' do
+    it 'When invalid request, repository from another team' do
       deleted_id = @teams.first.repositories.first.repository_rows.last.id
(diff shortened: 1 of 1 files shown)


spec/requests/api/v1/inventory_items_controller_spec.rb
```diff
@@ -126,4 +126,50 @@
...
+      delete api_v1_team_inventory_item_path(
+        id: deleted_id,
+        team_id: @teams.first.id,
+        inventory_id: @teams.second.repositories.first.id
+      ), headers: @valid_headers
+      expect(response).to have_http_status(404)
+      expect(RepositoryRow.where(id: deleted_id)).to exist
+    end
+
+    it 'When invalid request, repository from another team' do
```
Ducz0r: I have only this remark, can you get rid of the invalid symbol towards the end of this line?

mikechea: would that be line 171
`expect(response).to have_http_status(403)`?

Ducz0r: Nah, I mean the last character in this string: `it 'When invalid request, repository from another team'`. It renders as the funky unknown symbol to me.

PR #1294, opened 2018-09-09
add unit tests for Create inventory_column endpoint [SCI-2699]

mikechea:
close SCI-2699


mikechea:
can_manage_repository_column missing @inventory_column.

PR #1296, opened 2018-09-09
close mc-SC-2701; DELETE inventory column endpoint unit tests [SCI-2701]

mikechea:
added unit tests for DELETE inventory column endpoint


Ducz0r:
I've took the liberty to merge this one.

PR #1302, opened 2018-09-13
Add CREATE Result endpoint: Create Text result [SCI-2718]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-2718

### DIFF
--- a/app/controllers/api/v1/results_controller.rb
+++ b/app/controllers/api/v1/results_controller.rb
@@ -6,8 +6,8 @@ module Api
       include TinyMceHelper
 
       before_action :load_vars
-      before_action :load_result, only: %i(show update destroy)
-      before_action :check_manage_permissions, only: %i(create update destroy)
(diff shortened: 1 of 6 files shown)


app/controllers/api/v1/results_controller.rb
```diff
@@ -0,0 +1,116 @@
...
+              image.original_filename = image_params[:file_name]
+              tiny_img = TinyMceAsset.create!(image: image, team: @team)
+              result_text.text.sub!("[~tiny_mce_id:#{token}]",
+                                    "[~tiny_mce_id:#{tiny_img.id}]")
+            end
+          end
+          @result = Result.new(
+            user: current_user,
+            my_module: @task,
+            name: result_params[:name],
```
Ducz0r: This name is never retrieved, change this to `result_params[:attributes][:name]`, or change the return value of `result_params`.

PR #1312, opened 2018-09-24
Implement Projects view - navigation bar (tabs) [SCI-2729]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-2729


okriuchykhin:
I think it is better to implement export button as part of this task https://biosistemika.atlassian.net/browse/SCI-2645

As for Create project button, it looks unchanged in specs.

mlorb:
In the specs the position of Create project button is moved to the left. And Sort by button shouldn't be seen on table view.

PR #1323, opened 2018-10-03
Add endpoints for managing inventory list items [SCI-2741]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-2741

### DIFF
--- a/spec/requests/api/v1/inventory_list_items_controller_spec.rb
+++ b/spec/requests/api/v1/inventory_list_items_controller_spec.rb
@@ -85,6 +85,16 @@ RSpec.describe 'Api::V1::InventoryListItemsController', type: :request do
       expect { hash_body = json }.not_to raise_exception
       expect(hash_body).to match({})
     end
+
+    it 'When invalid request, items from text column' do
+      get api_v1_team_inventory_column_list_items_path(
+        team_id: @teams.first.id,
+        inventory_id: @valid_inventory.id,
+        column_id: @text_column.id
+      ), headers: @valid_headers
+      expect(response).to have_http_status(400)
+      expect(response.body).to be_empty
+    end
   end
 
   describe 'GET inventory_list_items, #show' do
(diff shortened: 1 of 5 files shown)


spec/requests/api/v1/inventory_list_items_controller_spec.rb
```diff
@@ -0,0 +1,351 @@
...
+
+    @valid_inventory = create(:repository, name: Faker::Name.unique.name,
+                              created_by: @user, team: @teams.first)
+
+    @wrong_inventory = create(:repository, name: Faker::Name.unique.name,
+                              created_by: @user, team: @teams.second)
+    create(:repository_column, name: Faker::Name.unique.name,
+           repository: @wrong_inventory, data_type: :RepositoryTextValue)
+
+    @text_column = create(:repository_column, name: Faker::Name.unique.name,
```
Ducz0r: This `@text_column` is never used from what I can see. Probably, it'd be good to test against this column as well, since error needs to be returned for list item endpoints for this column.

spec/requests/api/v1/inventory_list_items_controller_spec.rb
```diff
@@ -0,0 +1,351 @@
...
+          .as_json[:data]
+      )
+    end
+
+    it 'When invalid request, non existing list item' do
+      hash_body = nil
+      get api_v1_team_inventory_column_list_item_path(
+        id: 999,
+        team_id: @teams.first.id,
+        inventory_id: 123,
```
Ducz0r: Shouldn't `inventory_id` and `column_id` be same as in the previous test (e.g. non-random), since we're just testing the ID of the list item?

spec/requests/api/v1/inventory_list_items_controller_spec.rb
```diff
@@ -0,0 +1,351 @@
...
+        team_id: @teams.first.id,
+        inventory_id: @valid_inventory.id,
+        column_id: @list_column
+      ), params: invalid_request_body.to_json, headers: @valid_headers
+      expect(response).to have_http_status(400)
+      expect { hash_body = json }.to_not raise_exception
+      expect(hash_body).to match({})
+    end
+
+    it 'When invalid request, missing attributes values' do
```
Ducz0r: I think it'd be also good to write a test to check with an invalid `id` parameter here, for `PUT`.

PR #1327, opened 2018-10-07
Multiple fixes to bring the API to the JSONAPI standard [SCI-2768]

Ducz0r:
The whole current RESTful API has been checked for responses & endpoints, and fixed as per internal development meeting we had.

Closes [SCI-2768](https://biosistemika.atlassian.net/browse/SCI-2768).


Ducz0r:
@okriuchykhin In scope of this task, I've renamed the endpoints `users/:id/user_identities/...` into `users/:id/identities/...`? Do you see any problem with this?

okriuchykhin:
@Ducz0r should be fine, I think it is not used yet

PR #1329, opened 2018-10-10
Implement user state saving for projects overview [SCI-2733]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-2733

Datatables were also updated to recent version, some unused plugins removed


app/models/view_state.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal:true
+
+class ViewState < ApplicationRecord
+  belongs_to :user
```
Ducz0r: Nice, simple & efficient :+1:.

app/models/team.rb
```diff
@@ -35,11 +35,28 @@ class Team < ApplicationRecord
   has_many :reports, inverse_of: :team, dependent: :destroy
   has_many :datatables_reports,
            class_name: 'Views::Datatables::DatatablesReport'
+  has_many :view_states, as: :viewable, dependent: :destroy
 
   after_commit do
     Views::Datatables::DatatablesReport.refresh_materialized_view
   end
 
+  def current_view_state(user)
```
Ducz0r: I'm just thinking, could we isolate this (albeit not very big) code block into its own service, maybe even an abstracted one (e.g. have an abstract service interface for retrieving - and possibly saving - state for any user - object combo)?

I'm generally not a fan of Ruby's fat models paradigm (after seeing how bloated things can become). I just feel that this various state-keeping logic for GUI components doesn't really belong scattered into multiple `ActiveRecord` data classes. If anything, I myself conceptually relate these states more to users than to their related objects. That's why I think isolated services might be the most clean solution.

What do you think?

app/models/team.rb
```diff
@@ -35,11 +35,28 @@ class Team < ApplicationRecord
...
+    state = view_states.where(user: user).take
+    state || view_states.create!(user: user, state: default_view_state)
+  end
+
+  def default_view_state
```
Ducz0r: I just have a question: let's say that we also implement state save for e.g. individual team settings page (where there's a `DataTables`  listing all the users in a team). In this case, we're also storing a state for a `current_user` - `current_team` combo. Wouldn't it make sense to add additional root-level to this JSONB to designate which "view" for the user-object relation the state belongs to?

app/services/projects_overview_service.rb
```diff
@@ -118,15 +135,19 @@ def sortable_columns
...
+    order = @params[:order]&.values&.first
     if order
       dir = order[:dir] == 'desc' ? 'DESC' : 'ASC'
       column_index = order[:column]
     else
       dir = 'ASC'
       column_index = '1'
     end
+    if order_state != [column_index.to_i, dir.downcase]
+      @view_state.state['table']['order'][0] = [column_index.to_i, dir.downcase]
```
Ducz0r: Is this line totally correct? In default state, the order is like this: `order: [2, 'asc']`, but here, it changes to `order: [[2, 'asc'], 'asc']`.

okriuchykhin: Yes, this is correct, order should contain two-dimensional array, I will fix default value

app/models/team.rb
okriuchykhin: Yeah, it makes sense

app/models/team.rb
okriuchykhin: I can move it ot Concern

PR #1331, opened 2018-10-11
Add datalayer for associated oauth applications [SCI-2769]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-2769


Ducz0r:
I've closed this PR, as this is something we will not be implementing at this stage.

PR #1337, opened 2018-10-19
Add can_export_project? permission to allow archived projects to be exported [SCI-2805]

jbargu:
[SCI-2805](https://biosistemika.atlassian.net/browse/SCI-2805)

### DIFF
--- a/app/permissions/project.rb
+++ b/app/permissions/project.rb
@@ -38,6 +38,14 @@ Canaid::Permissions.register_for(Project) do
     user.is_owner_of_project?(project) && project.archived?
   end
 
+  # team: export projects
+  can :export_project do |user, project|
+    user.is_technician_or_higher_of_project?(project) ||
+      user.viewer?(project) ||
+      user.is_admin_of_team?(project.team) ||
+      (project.visible? && user.is_member_of_team?(project.team))
+  end
+
   # experiment: create
   can :create_experiments do |user, project|
     user.is_user_or_higher_of_project?(project)
--- a/app/controllers/teams_controller.rb
+++ b/app/controllers/teams_controller.rb
@@ -314,7 +314,7 @@ class TeamsController < ApplicationController
     if export_projects_params[:project_ids]
       projects = Project.where(id: export_projects_params[:project_ids])
       projects.each do |project|
-        render_403 unless can_read_project?(current_user, project)
+        render_403 unless can_export_project?(current_user, project)
       end
     end
   end


app/permissions/project.rb
```diff
@@ -38,6 +38,14 @@
     user.is_owner_of_project?(project) && project.archived?
   end
 
+  # team: export projects
+  can :export_project do |user, project|
+    user.is_technician_or_higher_of_project?(project) ||
+      user.viewer?(project) ||
+      user.is_admin_of_team?(project.team) ||
+      (project.visible? && user.is_member_of_team?(project.team))
+  end
```
mz3944: New permissions shouldn't be added unless absolutely necessary. Anyway, I think this permission is actually exactly the same as `read_project` permission:

 `is_member:of_project?` equals `user.viewer?(project)`, and I'm assuming if this is true, `.is_technician_or_higher_of_project?` will also be true, hence it's redundant.

In its comments you can see `read archive`, which means it's also meant for checking project export permissions. Are you 100% sure this solves the problem, because based on the written, it shouldn't?

Ducz0r: And I have another comment: if you _do_ implement a new permission, once the PR is merged, please notify Erik with the new permission - we have the permissions system documented for the customers.

jbargu: Well if it would be the same as `read_project` we wouldn't have this pull request, would we? :) 

There's a "decorator" on top that makes `read_project` true only if project is active, which prohibits exports of archived project. Actually **every** permission in this file is wrapped with this decorator. I don't see how to surpass this limitation to export archived projects, if it fails in the very beginning without adding another permission. I don't think I can call `read_project` without triggering the wrapper. Where am I wrong?

mz3944: I know we wouldn't, but it just didn't make sense - anyway, my mistake was that I mistakenly thought that the decorator is actually used with `||`, when in reality it's used with `&&`. Got confused I guess.

Well, the predisposition here is that if project is archived, you can't see it's content. so `can_read_*` permissions are actually meant for the described purpose.

So the way the permissions are structured, you should either look at the parent hierarchy read permission for it's descendent (e.g. `can_read_team?` for exporting samples). But this doesn't handle the export case as it should, so the only logical thing is to create a new permission (yes, you can call me Mr. Obvious).

But just use the same code as the read permission uses, minus the decorator rule. You can actually merge the two, which also logically makes sense:
```
  %i(read_project
     export_project)
    .each do |perm|
    can perm do |user, project|
      user.is_member_of_project?(project) ||
        user.is_admin_of_team?(project.team) ||
        (project.visible? && user.is_member_of_team?(project.team))
    end
  end
```

PR #1340, opened 2018-10-22
Fix helpers to use with export all [SCI-2793]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-2793

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -133,7 +133,6 @@ module ApplicationHelper
       match = el.match(sa_user)
       user = User.find_by_id(match[2].base62_decode)
       next unless user
-      team ||= current_team
       popover_for_user_name(user, team)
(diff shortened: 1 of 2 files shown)


app/helpers/application_helper.rb
```diff
@@ -166,7 +165,7 @@ def popover_for_user_name(user, team = nil, skip_user_status = false)
     end
 
     user_name = user.full_name
- 
+
```
mz3944: Maybe also consider removing `SmartAnnotations::TagToHtml.new(current_user, text).html`, just to be safe for the future. Only if it's possible/OK to do so.

okriuchykhin: Looks like ApplicationHelper is included only in controllers, so current_user should be available there, I assume we can leave it as it is for now.

PR #1341, opened 2018-10-23
Improve error messages in API [SCI-2773]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-2773

### DIFF
--- a/app/controllers/api/v1/base_controller.rb
+++ b/app/controllers/api/v1/base_controller.rb
@@ -5,6 +5,14 @@ module Api
     class BaseController < ApiController
       class TypeError < StandardError; end
       class IDMismatchError < StandardError; end
+      class PermissionError < StandardError
+        attr_reader :klass
+        attr_reader :mode
+        def initialize(klass, mode)
+          @klass = klass
+          @mode = mode
+        end
+      end
 
       rescue_from StandardError do |e|
         logger.error e.message
@@ -12,6 +20,15 @@ module Api
         render_error(I18n.t('api.core.errors.general.title'),
                      I18n.t('api.core.errors.general.detail'),
                      :bad_request)
+      end
+
+      rescue_from PermissionError do |e|
+        model = e.klass.name.underscore
+        render_error(
+          I18n.t("api.core.errors.#{e.mode}_permission.title"),
+          I18n.t("api.core.errors.#{e.mode}_permission.detail", model: model),
+          :forbidden
+        )
       end
 
       rescue_from TypeError do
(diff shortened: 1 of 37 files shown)


app/controllers/api/v1/inventories_controller.rb
```diff
@@ -48,34 +48,27 @@ def destroy
 
       private
 
-      def load_team
-        @team = Team.find(params.require(:team_id))
-        render jsonapi: {}, status: :forbidden unless can_read_team?(@team)
-      end
-
       def load_inventory
```
Ducz0r: Correct me if I'm wrong, but I have a possible suggestion: when we're loading various data objects in API controllers (e.g. this `load_inventory` right here), the only difference from the `BaseController.load_inventory` is that there, the parameter we're querying is `:inventory_id` in `BaseController`, but here it's `:id`.

Can we just update the `load_*` methods in `BaseController` to have a parameter key (which would default to e.g. `:inventory_id`), e.g.:

```ruby
def load_inventory(key = :inventory_id)
  @inventory = @team.repositories.find(params.require(key))
end
```

And then within dedicated controllers like this one, we can still reuse that `BaseController` method, we just pass a parameter to it, e.g.:
```ruby
before_action only: %i(show update destroy) { load_inventory(:id) }
```

This, of course, only works, if the methods are completely the same, with the exception of the parameter. If that's not the case, ignore this suggestion.

app/controllers/api/v1/base_controller.rb
```diff
@@ -3,14 +3,104 @@
...
+              id: request.uuid,
+              status: Rack::Utils.status_code(status),
+              title: title,
+              detail: message
+            }
+          ]
+        }, status: status
+      end
+
+      def permission_error(klass, mode)
```
Ducz0r: Wouldn't it be more streamlined if permission error is another sub-class of `StandardError` (with the 2 parameters, `klass` and `mode`). So whenever you want to render error from within descendants of `BaseController`, you always just use `raise`, instead of 2 ways (`raise` for majority of errors, and `permission_error` for permission errors). What do you think?

app/controllers/api/v1/inventories_controller.rb
okriuchykhin: Good point! Will implement it.

app/controllers/api/v1/base_controller.rb
okriuchykhin: Done

okriuchykhin:
Test were failing due to enabled request throttling in test environment

PR #1346, opened 2018-10-25
In project list table table is not updated when archiving/restoring project [SCI-2787]

mz3944:
[SCI-2787](https://biosistemika.atlassian.net/browse/SCI-2787)

Also fixes related bug: when all projects are archived, the table view will never display any projects.

### DIFF
--- a/app/assets/javascripts/projects/index.js
+++ b/app/assets/javascripts/projects/index.js
@@ -435,10 +435,10 @@
         sort: projectsViewSort
       },
       success: function(data) {
-        if (data.html.search('id="edit-project-cards-form-') > 0) {
+        if (data.html.search(/id="edit-project-cards-form-\d+"/)) {
+          viewContainer.html(data.html);
           $('#projects-absent').hide();
           $('#projects-present').show();
-          viewContainer.html(data.html);
         } else {
           $('#projects-present').hide();
           $('#projects-absent').show();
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/projects/index.js
```diff
@@ -435,7 +435,14 @@
...
       success: function(data) {
-        viewContainer.html(data.html);
+        if (data.html.search(/id="edit-project-cards-form-\d+"/)) {
+          viewContainer.html(data.html);
+          $('#projects-absent').hide();
+          $('#projects-present').show();
+        } else {
+          $('#projects-present').hide();
+          $('#projects-absent').show();
+        }
```
mz3944: This could possibly be solved in a more neat way, but I ran out of time.

app/assets/javascripts/projects/index.js
```diff
@@ -657,7 +664,16 @@
...
+      fnPreDrawCallback: function() {
+        $('#projects-absent').hide();
+        $('#projects-present').show();
+      },
       fnDrawCallback: function() {
+        var $table = TABLE.table().node();
+        if ($('.dataTables_empty', $table).length) {
+          $('#projects-present').hide();
+          $('#projects-absent').show();
+        }
```
mz3944: Same argument holds here.

PR #1357, opened 2018-10-30
Fix permission helpers in delayed export job [SCI-2834]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-2834

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -286,5 +286,7 @@ class Project < ApplicationRecord
       current_team: team,
       extra: '--keep-relative-links'
     )
+  ensure
+    report.destroy if report.present?
   end
 end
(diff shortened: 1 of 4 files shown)


app/models/project.rb
```diff
@@ -225,6 +225,7 @@ def notifications_count(user)
   def generate_report_pdf(user, team, pdf_name, obj_filenames = nil)
     ActionController::Renderer::RACK_KEY_TRANSLATION['warden'] ||= 'warden'
     proxy = Warden::Proxy.new({}, Warden::Manager.new({}))
+    proxy.set_user(user, scope: :user, store: false)
     renderer = ApplicationController.renderer.new(warden: proxy)
 
     report = Report.generate_whole_project_report(self, user, team)
```
mz3944: I would add a `begin`-`rescue` block up until the `report.destroy` line, and put the `report.destroy` in the `rescue` block, just to make sure the `report` is surely deleted in case we encounter another bug in the future.

PR #1358, opened 2018-11-04
Removal of the temporary files for export all [SCI-2646]

mz3944:
[SCI-2646](https://biosistemika.atlassian.net/browse/SCI-2646)

### DIFF
--- a/app/models/team_zip_export.rb
+++ b/app/models/team_zip_export.rb
@@ -18,21 +18,20 @@ class TeamZipExport < ZipExport
 
   def generate_exportable_zip(user, data, type, options = {})
     @user = user
-    FileUtils.mkdir_p(File.join(Rails.root, 'tmp/zip-ready'))
-    dir_to_zip = FileUtils.mkdir_p(
+    zip_input_dir = FileUtils.mkdir_p(
       File.join(Rails.root, "tmp/temp-zip-#{Time.now.to_i}")
     ).first
-    output_file = File.new(
-      File.join(Rails.root,
-                "tmp/zip-ready/projects-export-#{Time.now.to_i}.zip"),
+    zip_dir = FileUtils.mkdir_p(File.join(Rails.root, 'tmp/zip-ready')).first
+    zip_file = File.new(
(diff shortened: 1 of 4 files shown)


app/models/team_zip_export.rb
```diff
@@ -30,7 +30,10 @@ def generate_exportable_zip(user, data, type, options = {})
     fill_content(dir_to_zip, data, type, options)
     zip!(dir_to_zip, output_file.path)
     self.zip_file = File.open(output_file)
-    generate_notification(user) if save
+    if save
+      FileUtils.rm_rf([dir_to_zip, output_file], secure: true)
```
okriuchykhin: Maybe it makes sense to put it inside 'ensure' block to remove files even in case of error?

app/models/team_zip_export.rb
```diff
@@ -18,19 +18,20 @@ class TeamZipExport < ZipExport
...
     )
-    fill_content(dir_to_zip, data, type, options)
-    zip!(dir_to_zip, output_file.path)
-    self.zip_file = File.open(output_file)
+    fill_content(zip_input_dir, data, type, options)
+    zip!(zip_input_dir, zip_file)
+    self.zip_file = File.open(zip_file)
     generate_notification(user) if save
+  ensure
+    FileUtils.rm_rf([zip_input_dir, zip_file], secure: true)
```
mz3944: I didn't put an `if` check, since we create folders at the very beginning, so the creation shouldn't fail, and we can make the code a bit shorter.

app/models/zip_export.rb
```diff
@@ -47,18 +49,20 @@ def stored_on_s3?
...
     )
-    fill_content(dir_to_zip, data, type, options)
-    zip!(dir_to_zip, output_file.path)
-    self.zip_file = File.open(output_file)
+    fill_content(zip_input_dir, data, type, options)
+    zip!(zip_input_dir, zip_file)
+    self.zip_file = File.open(zip_file)
     generate_notification(user) if save
+  ensure
+    FileUtils.rm_rf([zip_input_dir, zip_file], secure: true)
```
mz3944: Same here.

PR #1360, opened 2018-11-07
Fix empty projects page bug and projects tree bug [SCI-2837]

mlorb:
[SCI-2837](https://biosistemika.atlassian.net/browse/SCI-2837)

Fixes also bug with not displaying projects navigation (project tree), when archive tab is selected.


app/controllers/projects_controller.rb
```diff
@@ -36,7 +36,8 @@ def index
           html: render_to_string(
             partial: 'projects/index/team_projects.html.erb',
                      locals: { projects: @projects }
-          )
+          ),
+          count: @projects.count
```
okriuchykhin: will perform additional query, please use 'size' instead as collection is already loaded

PR #1369, opened 2018-11-14
 Refactor result and step comments [SCI-2830]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-2830


jbargu:
Did check the number of requests, works wonderfully! Merge away!

PR #1377, opened 2018-11-19
Fix DataTable display for protocol modal [SCI-2863]

mz3944:
[SCI-2863](https://biosistemika.atlassian.net/browse/SCI-2863)


mz3944:
NOTE: Probably some addons have similar kind of modal with embedded DataTable, so I'll have to fix tehm also.

jbargu:
Great job, it's working. Merge away!

PR #1391, opened 2018-11-23
Fix smart annotations rendering issues [SCI-2880]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-2880

### DIFF
--- a/app/services/smart_annotations/html_preview.rb
+++ b/app/services/smart_annotations/html_preview.rb
@@ -36,7 +36,7 @@ module SmartAnnotations
           return "<span class='sa-type'>Tsk</span> <a href='" \
                  "#{ROUTES.module_archive_experiment_path(
                    object.experiment
-                 )}'>#{object.name}</a> #{I18n.t('atwho.res.archived')}"
+                 )}'>#{object.name} #{I18n.t('atwho.res.archived')}</a>"
         end
         "<span class='sa-type'>Tsk</span> " \
         "<a href='#{ROUTES.protocols_my_module_path(object)}'>" \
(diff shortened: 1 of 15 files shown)


app/views/my_modules/results.html.erb
```diff
@@ -48,9 +48,7 @@
 
   <div id="results" data-module-id="<%= @my_module.id %>">
   <% ordered_result_of(@my_module).each do |result| %>
-    <% cache [current_user, result] do %>
-      <%= render partial: "result", locals: { result: result } %>
-    <% end %>
+    <%= render partial: "result", locals: { result: result } %>
```
mz3944: Is this really needed?

app/services/smart_annotations/html_preview.rb
```diff
@@ -36,7 +36,7 @@ def generate_tsk_snippet(_, object)
           return "<span class='sa-type'>Tsk</span> <a href='" \
                  "#{ROUTES.module_archive_experiment_path(
                    object.experiment
-                 )}'>#{object.name}</a> #{I18n.t('atwho.res.archived')}"
+                 )}'>#{object.name} #{I18n.t('atwho.res.archived')}</a>"
```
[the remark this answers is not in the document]
mz3944: Actually, no, it's the other way around - you should fix the user smart annotations to match these, according to Jira task:

> 3. Color of smart annotation should be as it is task smart annotation. Item is blue and removed part grey including brackets.

app/views/my_modules/results.html.erb
okriuchykhin: Smart annotations(it is just signature in the text), so when object, which is referenced by SA, is changed, it is needed to invalidate cache for step/result containing such SA. But in order to find those steps/results it is needed to scan them all, and it is unacceptable.

PR #1392, opened 2018-11-23
Hide sign-in with third party applications (e.g. LinkedIn) in Oauth… [SCI-2676]

mlorb:
…flows

[SCI-2676](https://biosistemika.atlassian.net/browse/SCI-2676)

### DIFF
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -6,7 +6,7 @@ class Users::SessionsController < Devise::SessionsController
     # If user was redirected here from OAuth's authorize/new page (Doorkeeper
     # endpoint for authorizing an OAuth client), 3rd party sign-in buttons
     # (e.g. LinkedIn) should be hidden. See config/initializers/devise.rb.
-    @oauth_authorize = true if session['oauth_authorize'] == true
+    @oauth_authorize = session['oauth_authorize'] == true
     super
   end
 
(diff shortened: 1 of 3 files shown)


app/controllers/users/sessions_controller.rb
```diff
@@ -2,9 +2,16 @@ class Users::SessionsController < Devise::SessionsController
   # before_filter :configure_sign_in_params, only: [:create]
 
   # GET /resource/sign_in
-  # def new
-  #   super
-  # end
+  def new
+    # set @oauth_authorize to disable sign-ins
```
Ducz0r: I'd update this comment so it's even more descriptive:
```ruby
  def new
    # If user was redirected here from OAuth's authorize/new page (Doorkeeper
    # endpoint for authorizing an OAuth client), 3rd party sign-in buttons
    # (e.g. LinkedIn) should be hidden. See config/initializers/devise.rb.
    @oauth_authorize = false
```

app/controllers/users/sessions_controller.rb
```diff
@@ -2,9 +2,16 @@ class Users::SessionsController < Devise::SessionsController
...
+    # with third party applications (e.g. LinkedIn)
+    @oauth_authorize = false
+    if session['oauth_authorize'] == true
+      @oauth_authorize = true
+      session['oauth_authorize'] = false
```
Ducz0r: Why is this toggle neccesary? Because now, if user lands onto the log in page from the authorize page due to not being logged in, and inputs an invalid credentials, the LinkedIn button appears, which is not a good behavior. I think LinkedIn button should be hidden until user successfully logs in, or cancels the action and closes the browser window.

mlorb:
About `@oauth_authorize` definition check... I think it's working fine, because the check `@oauth_authorize != true` returns false even if the `@oauth_authorize` is `nil`.

Ducz0r:
Alright! Merge away 👍.

PR #1398, opened 2018-11-27
Fix projects table css [SCI-2786]

mlorb:
[SCI-2786](https://biosistemika.atlassian.net/browse/SCI-2786)

### DIFF
--- a/app/assets/stylesheets/themes/projects.scss
+++ b/app/assets/stylesheets/themes/projects.scss
@@ -11,4 +11,8 @@
   .archived {
     background-color: $color-concrete;
   }
+
+  #dropdownMenu1 {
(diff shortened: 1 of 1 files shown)


mlorb:
If we would be exactly following the bug description, then that's right. But the default row height (when dropdown is not clicked) is higher than normal, because of the bigger dropdown arrow padding. So my fix made the row lower.

Ducz0r:
aaa
I really don't like this "jumping" interaction. This is with your code (I painted the background blue so it's clear that height is still changing) 😟.

mlorb:
It's funny. This "jumping" is not happening on my screen. I implemented your suggestion anyway.

PR #1403, opened 2018-11-29
Enable caching for navigation sidebar [SCI-2757]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-2757

### DIFF
--- a/app/views/shared/sidebar/_my_modules.html.erb
+++ b/app/views/shared/sidebar/_my_modules.html.erb
@@ -1,21 +1,11 @@
 <% if experiment.active_my_modules.present? %>
   <ul>
-    <% experiment.active_my_modules.sort_by{ |m| m.workflow_order }.each do |my_module| %>
-
-      <li class="leaf <%= "active" if currently_active? my_module %>"
-           data-module-id="<%= my_module.id %>">
-        <span class="tree-link second-indent">
-          <% if currently_active? my_module %>
-              <%= my_module.name %>
-          <% else %>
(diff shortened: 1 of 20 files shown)


app/views/shared/sidebar/_projects.html.erb
```diff
@@ -1,27 +1,16 @@
 <ul>
-  <% if @projects_tree.present? then %>
+  <% if @projects_tree.present? %>
```
mz3944: Is this check even needed, since we have a loop afterwards?

app/views/shared/sidebar/_my_modules.html.erb
```diff
@@ -1,21 +1,11 @@
...
-        <span class="tree-link second-indent">
-          <% if currently_active? my_module %>
-              <%= my_module.name %>
-          <% else %>
-            <%= link_to my_module.name, module_action_to_link_to(my_module) %>
-          <% end %>
-          <% if is_canvas? %>
-            <a href="" class="canvas-center-on"><span class="fas fa-map-marker-alt"></span></a>
-          <% end %>
+    <% experiment.active_my_modules.each do |my_module| %>
```
mz3944: Also add caching here?

okriuchykhin: I don't think we need to go that deep and cache such small fragments.

app/views/shared/sidebar/_projects.html.erb
okriuchykhin: Probably not

PR #1413, opened 2018-12-06
Fix error at accepting invitation and login to scinote [SCI-2900]

mlorb:
https://biosistemika.atlassian.net/browse/SCI-2900


mlorb:
Will be solved with [SCI-2871](https://biosistemika.atlassian.net/browse/SCI-2871).

PR #1414, opened 2018-12-07
Fix invited member is added multiple times to team [SCI-2899]

mlorb:
[SCI-2899](https://biosistemika.atlassian.net/browse/SCI-2899)

I was not able to reproduce the bug, to test if the fix is working. So I am hoping that these changes will help to solve the problem.

### DIFF
--- a/app/views/users/settings/teams/show.html.erb
+++ b/app/views/users/settings/teams/show.html.erb
@@ -93,7 +93,7 @@
       <div class="panel-body">
         <div class="col-xs-24 col-sm-12">
           <a href="#" class="btn btn-primary pull-right row" data-trigger="invite-users"
-             data-modal-id="team-invite-users-modal">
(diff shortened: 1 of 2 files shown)


app/views/users/settings/teams/show.html.erb
```diff
@@ -93,7 +93,7 @@
       <div class="panel-body">
         <div class="col-xs-24 col-sm-12">
           <a href="#" class="btn btn-primary pull-right row" data-trigger="invite-users"
-             data-modal-id="team-invite-users-modal">
+             data-turbolinks="false" data-modal-id="team-invite-users-modal">
```
mz3944: This doesn't seem to be needed, so remove it.

mz3944: OK.

PR #1415, opened 2018-12-10
Use built-in methods for Postgres extensions enabling [SCI-2864]

okriuchykhin:
[SCI-2864](https://biosistemika.atlassian.net/browse/SCI-2864)

Rails built-in methods for extension enabling use IF NOT EXISTS options, which allows us to skip it if extensions are already there

Also adds AWS cli to production image, might be used for backups

### DIFF
--- a/db/migrate/20151103155048_add_btree_gist_extension.rb
+++ b/db/migrate/20151103155048_add_btree_gist_extension.rb
@@ -4,13 +4,7 @@ include DatabaseHelper
 class AddBtreeGistExtension < ActiveRecord::Migration[4.2]
   def up
     if db_adapter_is? "PostgreSQL" then
-      create_extension :btree_gist
-    end
-  end
-
-  def down
-    if db_adapter_is? "PostgreSQL" then
-      drop_extension :btree_gist
+      enable_extension :btree_gist
     end
   end
 end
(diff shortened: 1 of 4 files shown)


db/migrate/20151103155048_add_btree_gist_extension.rb
```diff
@@ -4,13 +4,7 @@
...
   def up
     if db_adapter_is? "PostgreSQL" then
-      create_extension :btree_gist
-    end
-  end
-
-  def down
-    if db_adapter_is? "PostgreSQL" then
-      drop_extension :btree_gist
+      enable_extension :btree_gist
```
mz3944: If we're editing existing migration files, it would be nicer to just delete this and `db/migrate/20151021082639_add_pg_trgm_support.rb`, and create a new migration for this, wouldn't it?

okriuchykhin: How do you plan to create gist_index in migrations if this extension is enabled only in the end?

Ducz0r: I don't think this is neccesary at this time-point. If this is the only issue, then I believe this one can be merged.

PR #1420, opened 2018-12-13
Replace reports materialized view with normal queries and re-enable fragment caching [SCI-2894]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-2894


Ducz0r:
@mz3944 @okriuchykhin Kind of randomly stumbled upon this already merged PR.

Just wanted to check, is it okay that we completely removed the migration file `db/migrate/20180417062042_create_datatables_reports.rb`?

PR #1434, opened 2018-12-23
Fix (unite) Date and Time display formats [SCI-2903]

mz3944:
[SCI-2903](https://biosistemika.atlassian.net/browse/SCI-2903)

Related to:
https://gitlab.com/biosistemika/scinote/addons/electronic_signatures/merge_requests/33
https://gitlab.com/biosistemika/scinote/addons/organization_management/merge_requests/38
https://gitlab.com/biosistemika/scinote/addons/gilson_iot/merge_requests/56

### DIFF
--- a/app/assets/javascripts/protocols/import_export/import.js.erb
+++ b/app/assets/javascripts/protocols/import_export/import.js.erb
@@ -115,8 +115,29 @@ function importProtocolFromFile(
         $("#protocol_authors").val($(this).children("authors").text());
         $("#import_protocol_description").val($(this).children("description").text());
       }
-      $("#protocol_created_at").val($(this).children("created_at").text());
-      $("#protocol_updated_at").val($(this).children("updated_at").text());
+
+      $.ajax({
+        url: "/helpers/to_user_date_format.json",
+        type: "GET",
+        data: { "timestamp": $(this).children("created_at").text(),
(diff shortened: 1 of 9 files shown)


mz3944:
Made another swipe through the code (checking all `*_at` variables). The only two files which still don't have the display format of time/date displayed in a uniform way are these two, which are in a `*.js` and `*.jsx` format, but I'm not sure if they're problematic:
`app/javascript/src/config/locales/messages.js`
`app/javascript/src/scenes/SettingsPage/scenes/team/index.jsx`

mlorb:
I get now the error when importing protocol.io

```
I18n::ArgumentError in Protocols#protocolsio_import_create

Showing /usr/src/app/app/views/protocols/import_export/_import_json_protocol_preview_modal.html.erb where line #64 raised:
Object must be a Date, DateTime or Time object. "01-Jan-1970 00:00 +0000" given.
Trace of template inclusion: app/views/protocols/protocolsio_import_create.js.erb
```

mz3944:
Fixed. I suggest we should enforce a way of handling date and time formats, i.e. exclusively on backend or frontend. Also, I've used full datetime format with timezone only for logging related things, e.g. electronic signatures, audit trails, system logs, etc., since it seems appropriate for such "meta" information. I hope this is OK.

PR #1435, opened 2018-12-23
An error at archiving experiment with locked task [SCI-2912]

mz3944:
[SCI-2912](https://biosistemika.atlassian.net/browse/SCI-2912)

### DIFF
--- a/app/views/experiments/_dropdown_actions.html.erb
+++ b/app/views/experiments/_dropdown_actions.html.erb
@@ -31,7 +31,7 @@
                       tooltipcontent: I18n.t('tooltips.text.experiment.move')
                     } %></li>
   <% end %>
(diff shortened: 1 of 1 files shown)


mz3944:
Really? It works for me... Can you double check? What exactly isn't working?

mlorb:
It's like in issue description: if you have a locked task in experiment...
image
...you still see the `Archive` experiment option (it shouldn't be available).

After the click you get:
image

mz3944:
For me it is hidden... What stack of add-ons are you using? I'm using the Pro Instance one.

mz3944:
Can you add another reviewer? Just so that maybe we can figure out why it works for me but not for you... Since we cannot move forward from the current situation we are in.

PR #1438, opened 2019-01-04
Add counters to task inventories dropdown [SCI-2776]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-2776


okriuchykhin:
Made changes to make dropdown work even when server is down

PR #1439, opened 2019-01-04
Rename CORE_API_V1_PREVIEW env. variable to CORE_API_V1_ENABLED [SCI-2888]

okriuchykhin:
https://biosistemika.atlassian.net/browse/SCI-2888


jbargu:
Checked with `grep`, everything appears to be replaced. Merge away!

PR #1446, opened 2019-01-11
Remove test folder

Ducz0r:
Remove the test folder (apart from mailer) & update addon generator.

This PR should be checked a bit, I've simply `git rm`-ed the folder contents. Hopefully we don't have any dependencies on the `test` code, but one can never be sure. Most of these folders, however, haven't been touched in last 2 years.


Ducz0r:
That's actually a good catch, this file is only used as a template, and since I removed the code that copies it, it can go out!

PR #1448, opened 2019-01-12
Can not switch team [SCI-2929]

mz3944:
[SCI-2929](https://biosistemika.atlassian.net/browse/SCI-2929)

Related to: [MR-67](https://gitlab.com/biosistemika/scinote/addons/billing/merge_requests/67)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -65,8 +65,10 @@ class ApplicationController < ActionController::Base
   private
 
   def update_current_team
-    if current_user.current_team_id.blank? &&
-       current_user.teams.count > 0
+    current_team = Team.find_by_id(current_user.current_team_id)
+    if (current_team.nil? || !current_user.is_member_of_team?(current_team)) &&
+       current_user.teams.count.positive?
+
       current_user.update(
         current_team_id: current_user.teams.first.id
       )


app/controllers/application_controller.rb
```diff
@@ -65,8 +65,10 @@ def render_404
   private
 
   def update_current_team
-    if current_user.current_team_id.blank? &&
-       current_user.teams.count > 0
+    current_team = Team.find_by_id(current_user.current_team_id)
+    if (current_team.nil? || !current_user.is_member_of_team?(current_team)) &&
+       current_user.teams.count.positive?
+
```
radiokills: Shouldn't this much logic be in a model/service object?

app/controllers/application_controller.rb
```diff
@@ -65,8 +65,10 @@ def render_404
   private
 
   def update_current_team
-    if current_user.current_team_id.blank? &&
-       current_user.teams.count > 0
+    current_team = Team.find_by_id(current_user.current_team_id)
+    if (current_team.nil? || !current_user.is_member_of_team?(current_team)) &&
```
radiokills: we're not checking if user doesn't have a current_team_id anymore, is that ok? :)

app/controllers/application_controller.rb
```diff
@@ -65,8 +65,10 @@ def render_404
...
+       current_user.teams.count.positive?
+
       current_user.update(
         current_team_id: current_user.teams.first.id
```
radiokills: Wha happens when user is not in any team? (just asking :))

mz3944: Well, theoretically this shouldn't happen :).

app/controllers/application_controller.rb
mz3944: Yes, since if it's `nil`, the `current_team` will be set as `nil`.

app/controllers/application_controller.rb
mz3944: While I agree, this should be done as a part of a refactoring task. Which we do have planned, in the future.

mz3944:
Just a note - when you assign yourself as a reviewer, also assign yourself as an "Assignee", because otherwise I cannot see that someone has reviewed my PR from the PR page.

mz3944:
Also, don't forget to review [MR-67](https://gitlab.com/biosistemika/scinote/addons/billing/merge_requests/67).

mz3944:
@radiokills is this OK to be merged? You didn't provide any response in the comments ;).

PR #1449, opened 2019-01-15
Update Ruby to 2.4.5 and add jemalloc package

okriuchykhin:
Updates Ruby version and adds optional implementation of memory allocator(jemalloc, can be enabled through env variables, can reduce total memory usage)


Ducz0r:
Just for reference, the reasoning for this pull request is that we need to investigate the memory consumption on our production instances, and `libjemalloc1` might help us with that. It's only enabled by specifying env. variables, so it shouldn't affect anything unless we toggle it on.

PR #1450, opened 2019-01-16
Add envirnoments to run app without docker, reorder Gemfile

urbanrotnik:
Jira ticket: No Ticket

**todo:**
Make a flexible environment for running and developing locally (without docker)

**What's done:** 
- Added extra ENVs for setup custom DB connection (`host`, `password`, `username`)
- Added extra ENVs to disable/enable `Paperclip`s spoof check method
- Reordered Gemfile (to match rubocop requirements)

- I also tried to include [Spring](https://github.com/rails/spring). Could speed up development, but I didn't add it because it didn't work on the first try, and I don't want to spend more time on this right now (test didn't pass). Maybe later...

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -19,15 +19,13 @@ default: &default
   encoding: unicode
   database: postgres
   pool: 5
-  username: postgres
-  password: mysecretpassword
-  host: db
+
   # For details on connection pooling, see rails configuration guide
   # http://guides.rubyonrails.org/configuring.html#database-pooling
 
 development:
   <<: *default
-  database: scinote_development
+  url: <%= ENV['DATABASE_DEV_URL'] || 'postgresql://postgres:mysecretpassword@db/scinote_development' %>
 
   # The specified database role being used to connect to postgres.
   # To create additional roles in postgres see `$ createuser --help`.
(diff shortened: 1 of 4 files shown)


config/database.yml
```diff
@@ -19,9 +19,9 @@ default: &default
   encoding: unicode
   database: postgres
   pool: 5
-  username: postgres
-  password: mysecretpassword
-  host: db
+  username: <%= ENV['RUNNING_LOCALLY'] ? ENV['DATABASE_USERNAME'] : 'postgres' %>
```
Ducz0r: I would try to change approach here. As you can see under `production` section, we already have a `DATABASE_URL` environmental variable in use there.

I would suggest to change this file that - for `development` and `test` - we can also reuse this variable, so if it's defined, it's simply used (the URL can contain all 3 things, username, password, and host), and overrides the default values; however, if it's not defined, these 3 values should fallback to the default values (`postgres`, `mysecretpassword`, `db`).

config/initializers/paperclip.rb
```diff
@@ -66,6 +66,8 @@ module Paperclip
   # Checks file for spoofing
   class MediaTypeSpoofDetector
     def spoofed?
+      return false if ENV['SPOOFING_CHECK_DISABLED'] && ENV['RUNNING_LOCALY']
```
Ducz0r: Does this "just" work? In any case, I would just change the variable to be called `DISABLE_SPOOF_CHECKING`.

PR #1455, opened 2019-01-17
Sci 2936 change indexes to bigint

radiokills:


db/migrate/20190117155006_change_indices_from_int_to_bigint.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+class ChangeIndicesFromIntToBigint < ActiveRecord::Migration[5.1]
```
urbanrotnik: Typo in class name? :)

mz3944:
Just a note: put a link into the PR description like this: [SCI-2936](https://biosistemika.atlassian.net/browse/SCI-2936). And the title should match the Jira ticket name with added ticket number, i.e. **Write a DB migration to convert index columns to bigint [SCI-2936]**. Also, if PR has dependent addons PRs, like this one has, you should link them in the description, and link back to this PR in each related addons PR.

mz3944:
It seems OK, but I didn't check if you actually got all of the IDs, since I would be basically doing double work. Merge away!

PR #1457, opened 2019-01-18
Change experiment cards look and feel [SCI-2933]

mlorb:
[SCI-2933](https://biosistemika.atlassian.net/browse/SCI-2933)

### DIFF
--- a/app/assets/javascripts/experiments/dropdown_actions.js
+++ b/app/assets/javascripts/experiments/dropdown_actions.js
@@ -112,8 +112,8 @@
   }
   // Initialize no description edit link
   function initEditNoDescription(){
-    var modal = "#edit-experiment-modal-";
-    $.each($(".experiment-no-description"), function(){
+    var modal = '#edit-experiment-modal-';
+    $.each($('.experiment-no-description'), function() {
       var id = modal + $(this).data("id");
       initializeModal($(this), id);
     });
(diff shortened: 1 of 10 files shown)


Ducz0r:
I haven't checked the code on this PR, just wanted to make sure that we don't run into the same issues we ran when we deployed refactored version of projects `/index` page, when we encountered significant load time issues on big, production database.

So I'm just pointing this out: if there were any changes on the queries, or how the representational layer (`.html.erb` views) lists data, just consider that these queries will be executed on quite large tables, and need to be fast :slightly_smiling_face:.

mlorb:
When I made this PR, I forgot to make the changes in Jira issue about some additional stuff that was discussed and decided. Sorry, I made a comment now.

Here is the list of your requests:

> * Why are there now 3 experiments per row? The epic doesn't mention this change.

We decided to make this change. Sorry, I forgot to make a comment in Jira.

> * I would make the whole area below the horizontal line as being clickable/link (also for new experiment card).

Done.

> * The experiment description should be below image.

It's specified that "The Dates and Experiment description are positioned above the Experiment graph image"

> * Copy as template & dropdown should be on the same horizontal level as dates.

We decided to make a change. I updated Jira.

> * Dropdown should still have copy as template option.

Fixed.

> * Edit details & Archive have no in-app Help, is that OK? I presume yes.

Yes.

> * I would horizontally center experiment image.

Done.

> * Is it correct that description height extends, or should it be of constant height? I assume it is, just want to make sure.

Yes, we changed that also.

> * On new experiment card, make "Create new experiment" also clickable/link.

Done.

mlorb:
@Ducz0r no changes were made on back-end.

mz3944:
You forgot to mention the related GitLab MR.

PR #1460, opened 2019-01-21
Refactor 'clone_experiment_as_template' method into service object [SCI-2925]

urbanrotnik:
Jira ticket: [SCI-2925](https://biosistemika.atlassian.net/secure/RapidBoard.jspa?rapidView=1&projectKey=SCI&modal=detail&selectedIssue=SCI-2925)

**todo:**
Experiments can be copied as a template into another project. This logic is encapsulated in the `app/models/experiment.rb`, in `deep_clone_to_project` method.

**What's done:** 
Moved to the service object, added `rspec` tests for it. 

**NOTE**
This is implemented on base of [SCI-2926](https://github.com/biosistemika/scinote-web/pull/1458), so that PR should be merged before.


app/services/experiments/copy_experiment_as_template_service.rb
```diff
@@ -0,0 +1,103 @@
+# frozen_string_literal: true
+
+module Experiments
+  class CopyExperimentAsTemplateService
```
jbargu: I do believe this whole class can be DRYed up together with the `MoveToProjectService`, because many things are repeating. May I suggest creating a parent class where common logic would reside?

app/services/experiments/copy_experiment_as_template_service.rb
```diff
@@ -0,0 +1,103 @@
...
+    extend Service
+
+    attr_reader :errors, :c_exp
+    alias cloned_experiment c_exp
+
+    def initialize(experiment_id:, project_id:, user_id:)
```
jbargu: This could be in parent class, additionally assigning `@c_exp` in this one.

app/services/experiments/copy_experiment_as_template_service.rb
```diff
@@ -0,0 +1,103 @@
...
+      @errors.merge!(@c_exp.errors.to_hash) unless @c_exp.valid?
+
+      @c_exp = nil unless succeed?
+      @c_exp.delay.generate_workflow_img if succeed?
+      track_activity if succeed?
+
+      self
+    end
+
+    def succeed?
```
jbargu: This could be in parent class.

app/services/experiments/copy_experiment_as_template_service.rb
```diff
@@ -0,0 +1,103 @@
...
+      @errors.none?
+    end
+
+    private
+
+    def find_uniq_name
```
jbargu: This could be in a parent class.

app/services/experiments/copy_experiment_as_template_service.rb
```diff
@@ -0,0 +1,103 @@
...
+      experiment_names = @project.experiments.map(&:name)
+      format = 'Clone %d - %s'
+      free_index = 1
+      free_index += 1 while experiment_names
+                            .include?(format(format, free_index, @exp.name))
+      format(format, free_index, @exp.name).truncate(Constants::NAME_MAX_LENGTH)
+    end
+
+    def valid?
+      unless @exp && @project && @user
```
jbargu: This part could be in a parent class, the bottom part is specific  for both child classes.

app/services/experiments/copy_experiment_as_template_service.rb
```diff
@@ -0,0 +1,103 @@
...
+            "Can't find #{key.capitalize}" if value.nil?
+          end.compact
+        return false
+      end
+
+      if @exp.projects_with_role_above_user(@user).include?(@project)
+        true
+      else
+        @errors[:user_without_permissions] =
+          ['You are not allowed to copy this experiment to this project']
```
jbargu: Still don't know if this should be localized due to being only for debugging purposes. What do you think?

app/services/experiments/copy_experiment_as_template_service.rb
urbanrotnik: If we decide for full refactoring (`move_to_proj`, `clone_proj`, `generate_workflow_image` + `Service.rb` module) with upper ideas, maybe we should open a new JIRA ticket and have separate PR for that?  
@jbargu 
@Ducz0r

app/services/experiments/copy_experiment_as_template_service.rb
urbanrotnik: 👍

app/services/experiments/copy_experiment_as_template_service.rb
jbargu: I might've mixed it up with some other PR, sorry!

jbargu: Right, this looks a similar code https://github.com/biosistemika/scinote-web/pull/1464/files#diff-5a91fb79d7efed2d7940a85be0ffd95fR161

app/services/experiments/copy_experiment_as_template_service.rb
jbargu: Create a parent class `MoveCloneService` which extends `Service.rb` then this one and the `MoveService` extend the `MoveCloneService`.  `succeed?` is then placed in the `MoveCloneService`, along with every common logic.

Your approach would probably be suitable if we had `@errors` and `succeed?` in every service.

I hope I'm not missing something here, I feel a bit under the weather.

urbanrotnik: As you said. But yes, `errors` and `succeed?` should be in every service, I guess?

app/services/experiments/copy_experiment_as_template_service.rb
Ducz0r: Yeah I don't know, I don't really see big issue in this not being localized. It's really more of a logging/debugging nature. Does it appear in the logs?

app/services/experiments/copy_experiment_as_template_service.rb
Ducz0r: Hmm, I see 2 things here (which I agree with you @urbanrotnik that we should probably open a separate task/s about it):
* extension of existing `Service` with some additional interfacing (e.g. `success`, `errors`, ...) that could be usable for any service
* introducing a parent Service just for `CopyExperimentAsTemplateService`, `MoveExperimentService`, just to DRY these 2 services a bit.

PR #1461, opened 2019-01-23
Add recaptcha to invitation modal [SCI-2935]

urbanrotnik:
Jira ticket: [SCI-2935](https://biosistemika.atlassian.net/browse/SCI-2935?filter=-1)

**todo:**
Add ReCaptcha to invitation modal 

**What's done:** 
Added `before_filter`, updated `ajax` call `invite_users`

**NOTE**
I didn't use existing implementation with `recaptcha_tag` helper because I am not inside of `form` element. 
Now we have included `recaptcha` at 3 diffrent users experiences, everyone has own `before_filter` for checking captcha. If we will use `recaptcha` on more places, implementing universal `check_captcha` on higher controller level should be a better way and keep other logic after validation outside of that `before_filter`.

### DIFF
--- a/app/assets/javascripts/users/settings/teams/invite_users_modal.js
+++ b/app/assets/javascripts/users/settings/teams/invite_users_modal.js
@@ -1,6 +1,16 @@
 (function() {
   'use strict';
 
+  function reloadRecaptchaIfExists() {
+    if (typeof grecaptcha !== 'undefined') {
+      grecaptcha.reset();
+    }
+  }
+
+  window.recaptchaCallback = function recaptchaCallback(response) {
+    $('#recaptcha-invite-modal').val(response);
+  };
+
   function initializeModal(modal) {
     var modalDialog = modal.find('.modal-dialog');
     var type = modal.attr('data-type');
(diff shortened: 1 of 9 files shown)


app/controllers/users/invitations_controller.rb
```diff
@@ -188,6 +190,14 @@ def generate_notification(user, target_user, role, team)
       end
     end
 
+    def check_captcha_for_invite
+      if Rails.configuration.x.enable_recaptcha
+        unless verify_recaptcha
+          render json: { recaptcha_error: t('invite_users.errors.recaptcha') }
```
radiokills: would it make sense to send a status code with this (i.e. 400)?

app/assets/javascripts/users/settings/teams/invite_users_modal.js
```diff
@@ -123,23 +137,31 @@
...
-            stepResultsDiv.html(data.html);
-            stepResults.show();
-            // Add 'data-invited="true"' status to modal element
-            modal.attr('data-invited', 'true');
+            if (data.recaptcha_error) {
+              animateSpinner(modalDialog, false);
+              recaptchaErrorMsgDiv.removeClass('hidden');
+              recaptchaErrorText.text(data.recaptcha_error);
+              reloadRecaptchaIfExists();
+            } else {
```
radiokills: In the ideal world this should be handled by error callback. For this controller should return an HTTP error code (for example 422)

app/controllers/users/invitations_controller.rb
urbanrotnik: 👍 Yeap

app/assets/javascripts/users/settings/teams/invite_users_modal.js
urbanrotnik: Agree 👍

PR #1472, opened 2019-01-30
System Notifications - Migration of existing data

radiokills:
Jira ticket: [SCI-2953](https://biosistemika.atlassian.net/browse/SCI-2953)

### What was done
Added "copy" and "delete" rake task. Removed obsolete "new_system" task. 

### Note:
This code is a bit ugly. We have to transform and copy approx 620K of user_notification record. In order to make this script faster I'm building raw SQL string and then executing it. However, this is a one-trick-pony, so I believe it's acceptable. Maybe we should delete these tasks after they are ran. 

**Note 2:** This pull request also includes @urbanrotnik commits.

### DIFF
--- a/lib/tasks/notifications.rake
+++ b/lib/tasks/notifications.rake
@@ -17,7 +17,8 @@ namespace :notifications do
         modal_title: system_notification.title,
         modal_body: system_notification.message,
         show_on_login: false,
-        source_created_at: system_notification.created_at
+        source_created_at: system_notification.created_at,
+        last_time_changed_at: system_notification.created_at
       )
 
       created_at = system_notification.created_at
--- a/lib/tasks/notifications.rake
+++ b/lib/tasks/notifications.rake
@@ -1,23 +1,72 @@
-namespace :notifications do
-  desc 'Creates new system notification for all active users'
-  task :new_system, [:title, :message] => :environment do |_, args|
-    include NotificationsHelper
+# frozen_string_literal: true
 
-    if args.blank? ||
-       args.empty? ||
-       args[:title].blank? ||
-       args[:message].blank?
-      puts 'One or both of arguments are missing'
-      return
+namespace :notifications do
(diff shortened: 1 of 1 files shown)


radiokills:
> Nice SQL generator :)
> How many records we expect with this `.where.not('title like ?', 'Congratulations%')` query?
> 
> Friendly advice: Install `overcommit` hooks

It's exactl 33 records, however there are ~620K of user_messages.

I was checking this with rubocop, but as I said: this script is one-time only, and it should maybe even be removed once it's ran. So I really wouldn't bother with refactoring of too long blocks. But if you think it's necessary I'll do it.

urbanrotnik:
> > Nice SQL generator :)
> > How many records we expect with this `.where.not('title like ?', 'Congratulations%')` query?
> > Friendly advice: Install `overcommit` hooks
> 
> It's exactl 33 records, however there are ~620K of user_messages.
> 
> I was checking this with rubocop, but as I said: this script is one-time only, and it should maybe even be removed once it's ran. So I really wouldn't bother with refactoring of too long blocks. But if you think it's necessary I'll do it.

Ahh.. ok then. So you don't need to process them in batches. Everything else looks good. 

Don't worry about rcop... ;)

lib/tasks/notifications.rake
```diff
@@ -1,23 +1,72 @@
...
+          )
+          VALUES
+        '
+      user_notifications = UserNotification
+                           .where(notification_id: system_notification.id)
+      values_array = user_notifications.map do |user_notification|
+        user_notification
+          .slice(:user_id, :created_at, :updated_at)
+          .merge(system_notification_id: new_notification.id)
+          .merge(seen_at: created_at, read_at: created_at)
```
urbanrotnik: Do we care if someone saw the notification in the old system or not? 
We have `checked` flag there I think.

radiokills: o-oh, thanks for heads up!

radiokills: fixed

PR #1476, opened 2019-02-01
Add test for templates service [SCI-2944]

okriuchykhin:
Jira ticket: [SCI-2944](https://biosistemika.atlassian.net/browse/SCI-2944)

###What was done
added test for templates service based on demo project's experiment export/import

### DIFF
--- a/spec/services/templates_service_spec.rb
+++ b/spec/services/templates_service_spec.rb
@@ -14,6 +14,8 @@ describe TemplatesService do
         create(:user_team, user: admin_user, team: main_team)
         seed_demo_data(main_team.created_by, main_team)
         dj_worker = Delayed::Worker.new
+        # Two runs are needed to execute additional jobs which get generated
+        # during first run
         Delayed::Job.all.each { |job| dj_worker.run(job) }
         Delayed::Job.all.each { |job| dj_worker.run(job) }
         demo_exp = main_team.projects.first.experiments.first
--- a/spec/services/templates_service_spec.rb
+++ b/spec/services/templates_service_spec.rb
@@ -36,8 +36,8 @@ describe TemplatesService do
         expect(tmpl_exp.uuid).to_not eq(nil)
         expect(tmpl_exp.my_modules.pluck(:name))
           .to match_array(demo_exp.my_modules.pluck(:name))
-        tmpl_tasks = tmpl_exp.my_modules.order(:id)
-        demo_tasks = demo_exp.my_modules.order(:id)
+        tmpl_tasks = tmpl_exp.my_modules
+        demo_tasks = demo_exp.my_modules
         demo_tasks.each do |demo_task|
           tmpl_task = tmpl_tasks.find_by_name(demo_task.name)
           expect(tmpl_task.name).to eq(demo_task.name)
@@ -70,7 +70,7 @@ describe TemplatesService do
               expect(demo_step.assets.pluck(:file_file_name))
                 .to match_array(tmpl_step.assets.pluck(:file_file_name))
             end
-            tmpl_step.assets.order(:id).each do |asset|
+            tmpl_step.assets.each do |asset|
               expect(asset.file.exists?).to eq(true)
             end
             if demo_step.tables.present?
--- a/spec/services/templates_service_spec.rb
+++ b/spec/services/templates_service_spec.rb
@@ -1,0 +1,100 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require 'fileutils'
+
+describe TemplatesService do
+  include FirstTimeDataGenerator
+  let!(:main_team) { create :team }
+  let!(:admin_user) { create :user }
+
+  describe '#update_project' do
+    context 'update templates project' do
+      it 'experiment is added to templates project' do
+        create(:user_team, user: admin_user, team: main_team)
+        seed_demo_data(main_team.created_by, main_team)
+        dj_worker = Delayed::Worker.new
+        Delayed::Job.all.each { |job| dj_worker.run(job) }
+        demo_exp = main_team.projects.first.experiments.first
+        exp_exporter = ModelExporters::ExperimentExporter.new(demo_exp.id)
+        exp_dir = exp_exporter.export_template_to_dir
+        tmplts_dir = "#{Rails.root}/tmp/testing_templates"
+        FileUtils.remove_dir(tmplts_dir) if Dir.exist?(tmplts_dir)
+        FileUtils.mkdir(tmplts_dir)
+        FileUtils.mv(exp_dir, "#{tmplts_dir}/experiment_#{demo_exp.id}")
+        templates_project = create :project, name: 'Templates', template: true
+        create(
+          :user_project, :owner, project: templates_project, user: admin_user
+        )
+        ts = TemplatesService.new(tmplts_dir)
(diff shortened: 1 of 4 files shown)


spec/services/templates_service_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+          demo_task.protocol.steps.each do |demo_step|
+            tmpl_step = tmpl_task.protocol.steps.find_by_name(demo_step.name)
+            expect(demo_step.name).to eq(tmpl_step.name)
+            expect(demo_step.description).to eq(tmpl_step.description)
+            if demo_step.assets.present?
+              expect(demo_step.assets.pluck(:file_file_name))
+                .to match_array(tmpl_step.assets.pluck(:file_file_name))
+            end
+            tmpl_step.assets.order(:id).each do |asset|
+              expect(asset.file.exists?).to eq(true)
```
mz3944: Do the same for `demo_step`. Also compare Assets contents, or at least size of the assets. For or assets also.

spec/services/templates_service_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+          expect(tmpl_task.name).to eq(demo_task.name)
+          expect(tmpl_task.task_comments.size)
+            .to eq(demo_task.task_comments.size)
+          demo_task.results.each do |demo_res|
+            tmpl_res = tmpl_task.results.find_by_name(demo_res.name)
+            expect(tmpl_res.name).to eq(demo_res.name)
+            if demo_res.asset
+              expect(tmpl_res.asset.file.exists?).to eq(true)
+              expect(demo_res.asset.file_file_name)
+                .to eq(tmpl_res.asset.file_file_name)
```
mz3944: Also check `asset.file.exists?` as you did for for steps.

spec/services/templates_service_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+  let!(:admin_user) { create :user }
+
+  describe '#update_project' do
+    context 'update templates project' do
+      it 'experiment is added to templates project' do
+        create(:user_team, user: admin_user, team: main_team)
+        seed_demo_data(main_team.created_by, main_team)
+        dj_worker = Delayed::Worker.new
+        Delayed::Job.all.each { |job| dj_worker.run(job) }
+        Delayed::Job.all.each { |job| dj_worker.run(job) }
```
mz3944: Why duplicate?

spec/services/templates_service_spec.rb
```diff
@@ -0,0 +1,93 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require 'fileutils'
+
+describe TemplatesService do
```
mz3944: Add `, type: :background_job`.

spec/services/templates_service_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+        tmpl_exp = templates_project.experiments.first
+
+        expect(tmpl_exp.name).to eq(demo_exp.name)
+        expect(tmpl_exp.uuid).to_not eq(nil)
+        expect(tmpl_exp.my_modules.pluck(:name))
+          .to match_array(demo_exp.my_modules.pluck(:name))
+        tmpl_tasks = tmpl_exp.my_modules.order(:id)
+        demo_tasks = demo_exp.my_modules.order(:id)
+        demo_tasks.each do |demo_task|
+          tmpl_task = tmpl_tasks.find_by_name(demo_task.name)
```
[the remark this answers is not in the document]
okriuchykhin: removed

spec/services/templates_service_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+            end
+            expect(demo_res.result_comments.size)
+              .to eq(tmpl_res.result_comments.size)
+          end
+          unless demo_task.protocol.present? &&
+                 demo_task.protocol.steps.size.positive?
+            next
+          end
+          demo_task.protocol.steps.each do |demo_step|
+            tmpl_step = tmpl_task.protocol.steps.find_by_name(demo_step.name)
```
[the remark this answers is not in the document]
okriuchykhin: it appears that order is not consistent on different systems, so I removed loop with index

spec/services/templates_service_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+            end
+            if demo_step.tables.present?
+              expect(demo_step.tables.pluck(:contents))
+                .to match_array(tmpl_step.tables.pluck(:contents))
+            end
+            if demo_step.checklists.present?
+              expect(demo_step.checklists.pluck(:name))
+                .to match_array(tmpl_step.checklists.pluck(:name))
+            end
+            expect(demo_step.step_comments.size)
```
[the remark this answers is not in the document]
okriuchykhin: Import service has it's own tests, so here I just test main attributes just to check that all objects have been imported

spec/services/templates_service_spec.rb
okriuchykhin: First run generates another set of delayed jobs

spec/services/templates_service_spec.rb
okriuchykhin: but it is already checked: expect(tmpl_res.asset.file.exists?).to eq(true)

spec/services/templates_service_spec.rb
okriuchykhin: Why should I check it for demo_step? I test only templates here, and in any case if template asset has file present it means that it was successfully copied from demo.

spec/services/templates_service_spec.rb
okriuchykhin: This is a service, not a background job

spec/services/templates_service_spec.rb
mz3944: My bad.

spec/services/templates_service_spec.rb
mz3944: OK, then just add comment so it is known.

spec/services/templates_service_spec.rb
mz3944: Sure, but this makes the delayed jobs (which you do trigger inside the tests) to run immediately. Or am I wrong?

spec/services/templates_service_spec.rb
mz3944: Pardon, I meant the other way around - do the same for steps as you did for the result, i.e. on at _line 96_.

spec/services/templates_service_spec.rb
mz3944: If I'm correct, then maybe you wouldn't need to call delayed jobs twice at _line 17_?

spec/services/templates_service_spec.rb
okriuchykhin: Sorry, but file has only 96 lines. Files attached to step are checked on line 74.

spec/services/templates_service_spec.rb
okriuchykhin: config.around(:each, type: :background_job) do |example|
    run_background_jobs_immediately do
      example.run
    end
  end

It just wraps test, jobs than are executed in the end

spec/services/templates_service_spec.rb
okriuchykhin: Apparently it doesn't

PR #1479, opened 2019-02-04
Optionally disable Tika background mode through env variable [SCI-2967]

okriuchykhin:
Jira ticket: [SCI-2967](https://biosistemika.atlassian.net/browse/SCI-2967)

### What was done
Introduce env variable which will prevent Tika server running as a daemon.

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -242,6 +242,7 @@ class Asset < ApplicationRecord
         file_path = fa.path
       end
 
+      # Start Tika as a server
       if !ENV['NO_TIKA_SERVER'] && Yomu.class_variable_get(:@@server_pid).nil?
         Yomu.server(:text)
       end
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -242,12 +242,11 @@ class Asset < ApplicationRecord
         file_path = fa.path
       end
 
-      unless Yomu.class_eval('@@server_pid')
-        Yomu.server(:text, nil)
-        sleep(5)
+      if !ENV['NO_TIKA_SERVER'] && Yomu.class_variable_get(:@@server_pid).nil?
+        Yomu.server(:text)
       end
-      y = Yomu.new file_path
 
+      y = Yomu.new file_path
       text_data = y.text
 
       if asset.asset_text_datum.present?


app/models/asset.rb
```diff
@@ -242,12 +242,11 @@ def self.extract_asset_text(asset_id)
         file_path = fa.path
       end
 
-      unless Yomu.class_eval('@@server_pid')
-        Yomu.server(:text, nil)
-        sleep(5)
+      if !ENV['NO_TIKA_SERVER'] && Yomu.class_variable_get(:@@server_pid).nil?
```
mz3944: Add comment `Start Tika as a server`

app/models/asset.rb
```diff
@@ -242,12 +242,11 @@ def self.extract_asset_text(asset_id)
...
+        Yomu.server(:text)
       end
-      y = Yomu.new file_path
 
+      y = Yomu.new file_path
       text_data = y.text
```
mz3944: Add `Yomu.kill_server!` to `ensure` block.

app/models/asset.rb
mz3944: So `NO_TIKA_SERVER` will need to be set for every sciNote instance, based on the fact if server has Tika installed or not. Will this be done in scope of this Jira task or as a separate one? Just mentioning it so that we won't forget it.

app/models/asset.rb
okriuchykhin: Why we should kill server if it was not started? And why we should kill it if we decided to start it?

app/models/asset.rb
okriuchykhin: there is no separate server package, it is always just one jar file which can be used in different ways. CF template is already updated

okriuchykhin:
> Some changes requested. A question though:
> 
> Yomu by default starts the JVM for each document and then stops it automatically. The server mode, where Tika starts as a server and processes whatever documents are thrown at it over a socket, is useful when processing several files at once, not just one, and it has time overhead when starting Java (but only happens once for many files). So is `Yomu.server(:text)` even needed (overkill)?

Sorry, I don't get this. We will have two options, starting tika as background server thus saving start up time, or calling tika per each file separately(without keeping it running in background) thus saving resources.

mz3944:
> > Some changes requested. A question though:
> > Yomu by default starts the JVM for each document and then stops it automatically. The server mode, where Tika starts as a server and processes whatever documents are thrown at it over a socket, is useful when processing several files at once, not just one, and it has time overhead when starting Java (but only happens once for many files). So is `Yomu.server(:text)` even needed (overkill)?
> 
> Sorry, I don't get this. We will have two options, starting tika as background server thus saving start up time, or calling tika per each file separately(without keeping it running in background) thus saving resources.

OK, so to sum up your intention - each instance will have either of the following behaviors, based on whether the `NO_TIKA_SERVER` is set (and true) or not:

- Tika will be running in a background in all of the cases, at all times, so it will be run only once and never stopped.
- Tika will be called for each file separately, in all cases, and then automatically stopped.

**My confusion was**, the Jira ticket didn't mention that both options should be preset. So the ticket should be updated to match your implementation. I updated the Jira task with the above description.

Merge Away!

PR #1483, opened 2019-02-06
Test experiment import [SCI-2946]

jbargu:
Jira ticket: [SCI-2946](https://biosistemika.atlassian.net/browse/SCI-2946)

### What was done
Create experiment import tests, focusing that everything gets imported. Export was partially addressed [here](https://github.com/biosistemika/scinote-web/pull/1476).

### Note:
I tried to implement this as DRY as possible and traceable, but just gave up, since it took too much effort fiddling with the rspec.


spec/services/model_importers/team_importer_spec.rb
```diff
@@ -0,0 +1,399 @@
...
+            # User assigns to the the module
+            unless my_module['user_my_modules'].empty?
+              expect(db_module.user_my_modules.first.user_id).to eq USER_ID
+            end
+          end
+        end
+      end
+    end
+  end
+end
```
radiokills: I bow to you for this exhaustive test.

jbargu: *bows back*. Could be much better though I believe :0

PR #1484, opened 2019-02-06
System notification modal window [SCI-2957 and SCI-2958]

aignatov-bio:
Jira ticket: https://biosistemika.atlassian.net/browse/SCI-2957

### What was done
Added modal window for system notifications, when clicking by news icon. Showing last system notification after user sign in.

#### ToDo:
In addons make override for method `show_login_system_notification?` to prevent show system notification modal, if others modals appear.

#### Description:
To Devise session controller added method `after_sign_in` that fires up after sign in action. Currently there is added `flash[:system_notification_modal] = true`, to show system notification modal after sign in. To main application layout added new condition, that render system notification modal.


urbanrotnik:
@aignatov-bio 
I think PR should be targeting to `feature/system-notifications` branch? 
If I am right, please add label for feature on PR :)

app/controllers/system_notifications_controller.rb
```diff
@@ -26,6 +30,13 @@ def mark_as_seen
     render json: { result: 'failed' }
   end
 
+  def mark_as_read
```
urbanrotnik: `mark_as_read` and `mark_as_seen` could be moved to `service` object. 
If it stays as it is, maybe `mark_as_read` could be changed from class to an instance method and you don't need to passing ID.

app/models/user_system_notification.rb
```diff
@@ -8,4 +8,42 @@ def self.mark_as_seen(notifications)
...
+    end
+  end
+
+  def self.modal(notification_id)
+    select(:modal_title, :modal_body, 'system_notifications.id')
+      .joins(:system_notification)
+      .find_by_system_notification_id(notification_id)
+  end
+
+  def self.show_on_login(update_read_time = false)
```
urbanrotnik: So we are flagging as read as soon as we are returning it. Shouldn't we `mark_as_read` with reusing `mark_read` method (or maybe service)?

app/controllers/system_notifications_controller.rb
aignatov-bio: Main idea, that we can pass multiple id’s and update seen_at with one SQL query.

app/controllers/system_notifications_controller.rb
```diff
@@ -26,6 +30,15 @@ def mark_as_seen
     render json: { result: 'failed' }
   end
 
+  def mark_read
+    current_user.user_system_notifications.mark_read(params[:id])
+    render json: { result: 'ok' }
```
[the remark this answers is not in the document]
aignatov-bio: In current example it isn’t matter, cause I am not handling response from this methods.

app/models/user_system_notification.rb
aignatov-bio: If we reuse mark_read method, than we will need perform 2 updates - 1 for update seen_at and 1 for read_at.

app/controllers/system_notifications_controller.rb
```diff
@@ -17,6 +17,10 @@ def index
     end
   end
 
+  def show
+    render json: current_user.user_system_notifications.modal(params[:id]) || {}
```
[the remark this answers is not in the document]
aignatov-bio: Yeah, it might be more clear code, cause we can implement this without additional method.

app/models/user_system_notification.rb
urbanrotnik: Yes, I know. I was thinking to mark as_read with ajax when user get a page. Probably overkill for this case...

app/controllers/system_notifications_controller.rb
urbanrotnik: @radiokills your opinion here?

aignatov-bio: This model currently small and this logic we can only use in this model, i think we need move it to service if we will use it in another model. Also i think leave it like class method and give ability to receive multiple id's, cause someone in future may want "Mark All As Read" button.

urbanrotnik:
@aignatov-bio Looks good. Maybe just wait for another review from @radiokills  before merging.

PR #1487, opened 2019-02-07
User counter on task card is not updated [SCI-3005]

aignatov-bio:
Jira ticket: https://biosistemika.atlassian.net/browse/SCI-3005

### What was done
Added missing attribute to module block


radiokills:
merge it plz

PR #1492, opened 2019-02-12
Fix protocols in team export service [SCI-3037]

okriuchykhin:
Jira ticket: [SCI-3037](https://biosistemika.atlassian.net/browse/SCI-3037)

### What was done
Refactored and fixed team export service


app/services/model_exporters/team_exporter.rb
```diff
@@ -57,7 +54,9 @@ def team(team)
         custom_fields: team.custom_fields,
         repositories: team.repositories.map { |r| repository(r) },
         tiny_mce_assets: team.tiny_mce_assets,
-        protocols: team.protocols.where(my_module: nil).map { |pr| protocol(pr) },
+        protocols: team.protocols.where(my_module: nil).map do |pr|
+          protocol(pr)
+        end,
```
mlorb: I think this change is not needed. One liner is better.

app/services/model_exporters/experiment_exporter.rb
```diff
@@ -3,10 +3,8 @@
 module ModelExporters
   class ExperimentExporter < ModelExporter
     def initialize(experiment_id)
-      @experiment = Experiment.find_by_id(experiment_id)
-      raise StandardError, 'Can not load experiment' unless @experiment
```
mlorb: We don't need to handle error anymore?

app/services/model_exporters/team_exporter.rb
```diff
@@ -3,11 +3,8 @@
 module ModelExporters
   class TeamExporter < ModelExporter
     def initialize(team_id)
-      @team = Team.includes(:user_teams).find_by_id(team_id)
-      raise StandardError, 'Can not load team' unless @team
```
mlorb: same here

app/services/model_exporters/experiment_exporter.rb
okriuchykhin: .find() throws an exception if object not found

app/services/model_exporters/team_exporter.rb
okriuchykhin: .find() throws an exception if object not found

app/services/model_exporters/team_exporter.rb
okriuchykhin: than it will not fit into the line

PR #1493, opened 2019-02-12
Add flag for archived tasks to experiment exporter [SCI-3033]

mlorb:
Jira ticket: [SCI-3033](https://biosistemika.atlassian.net/browse/SCI-3033)

### What was done
_Flag for archived tasks was added to disable exporting of archived tasks when exporting the templates. When exporting the team, archived tasks should be added to export._

### DIFF
--- a/app/models/my_module_group.rb
+++ b/app/models/my_module_group.rb
@@ -10,7 +10,7 @@ class MyModuleGroup < ApplicationRecord
              optional: true
   has_many :my_modules, inverse_of: :my_module_group, dependent: :nullify
 
-  scope :active, (lambda do
+  scope :without_archived_modules, (lambda do
     joins(:my_modules).where('my_modules.archived = ?', false).distinct
   end)
 
--- a/app/models/my_module_group.rb
+++ b/app/models/my_module_group.rb
@@ -10,6 +10,10 @@ class MyModuleGroup < ApplicationRecord
              optional: true
   has_many :my_modules, inverse_of: :my_module_group, dependent: :nullify
 
+  scope :active, (lambda do
+    joins(:my_modules).where('my_modules.archived = ?', false).distinct
+  end)
+
   def deep_clone_to_experiment(current_user, experiment)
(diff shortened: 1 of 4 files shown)


app/services/model_exporters/experiment_exporter.rb
```diff
@@ -35,10 +37,17 @@ def export_template_to_dir
     end
 
     def experiment
+      if @include_archived
+        my_modules = @experiment.my_modules
+        my_module_groups = @experiment.my_module_groups
+      else
+        my_modules = @experiment.active_my_modules
```
urbanrotnik: You should scope `my_modules` like this... `@experiment.my_modules.active` instead of using a method `active_my_modules`

app/services/model_exporters/experiment_exporter.rb
```diff
@@ -35,10 +37,17 @@ def export_template_to_dir
...
+        my_module_groups = @experiment.active_module_groups
```
urbanrotnik: Same here...

app/models/my_module_group.rb
```diff
@@ -10,6 +10,10 @@ class MyModuleGroup < ApplicationRecord
              optional: true
   has_many :my_modules, inverse_of: :my_module_group, dependent: :nullify
 
+  scope :active, (lambda do
+    joins(:my_modules).where('my_modules.archived = ?', false).distinct
```
urbanrotnik: Ok cool. Maybe only naming is a bit confused here. This is why I have been confused before and didn't get what I was looking on... 
We are scoping `my_modules_groups` but on `my_module` condition. 
What do you think to rename scope to something like `without_archived_modules`... 

Than `my_moudle_groups.without_archived_modules` will return you exactly what you read. 

Am I right and are you agree? :)

radiokills: :+1:

urbanrotnik:
PartyHard 🎉

PR #1498, opened 2019-02-15
System notification navbar menu drop down [SCI 2955]

aignatov-bio:
Jira ticket: https://biosistemika.atlassian.net/browse/SCI-2955

### What was done
Added dropdown menu to navbar with system notifications and specs

### DIFF
--- a/app/models/user_system_notification.rb
+++ b/app/models/user_system_notification.rb
@@ -18,12 +18,6 @@ class UserSystemNotification < ApplicationRecord
     end
   end
 
-  def self.modal(notification_id)
-    select(:modal_title, :modal_body, 'system_notifications.id')
-      .joins(:system_notification)
-      .find_by_system_notification_id(notification_id)
-  end
-
   def self.show_on_login(update_read_time = false)
     # for notification check leave update_read_time empty
     notification = joins(:system_notification)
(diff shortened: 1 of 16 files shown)


app/models/user_system_notification.rb
```diff
@@ -16,6 +18,16 @@ def self.mark_as_read(notification_id)
     end
   end
 
+  def self.unseen_count
+    unseen.count
+  end
+
+  def self.modal(notification_id)
```
urbanrotnik: Didn't you move this to scope?

aignatov-bio: I think maybe remove at all, cause we have unseen scope. And if we need find unseen count, we can simply use unseen.count.

spec/models/system_notification_spec.rb
```diff
@@ -69,4 +70,44 @@
...
+             system_notification: notifcation_one
+      create :user_system_notification,
+             user: user,
+             system_notification: notifcation_two
+    end
+
+    it 'get last notifications without search' do
+      result = SystemNotification.last_notifications(user)
+      expect(result.length).to eq 2
+      expect(result.first).to respond_to(
```
urbanrotnik: This test doesn't make sense here I think

aignatov-bio: It's more for test right result format. Cause if response structure will be changed, system notifications view may failed.

app/models/user_system_notification.rb
radiokills: @aignatov-bio Yes, that's a great idea.

PR #1505, opened 2019-02-20
Check for recaptcha enabled boolean value

jbargu:
Jira ticket: [SCI-2928](https://biosistemika.atlassian.net/browse/SCI-2928)

### What was done
Current check for `RECAPTCHA_ENABLED` only looked if the value was defined, but not its boolean value.

### DIFF
--- a/app/views/shared/_invite_users_modal.html.erb
+++ b/app/views/shared/_invite_users_modal.html.erb
@@ -99,7 +99,7 @@ invite_to_team = type.in?(%w(invite_to_team invite_to_team_with_role))
               </div>
             <% end %>
           <% end %>
-          <% if ENV['ENABLE_RECAPTCHA'] %>
+          <% if ENV['ENABLE_RECAPTCHA'] == 'true' %>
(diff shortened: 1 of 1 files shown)


app/views/shared/_invite_users_modal.html.erb
```diff
@@ -99,7 +99,7 @@ invite_to_team = type.in?(%w(invite_to_team invite_to_team_with_role))
               </div>
             <% end %>
           <% end %>
-          <% if ENV['ENABLE_RECAPTCHA'] %>
+          <% if ENV['ENABLE_RECAPTCHA'] == 'true' %>
```
urbanrotnik: Few tasks ago I figure out that we are using `secrets`. Should we go with `Rails.secrets...` instiead of checking environment variables directly?

jbargu: Wouldn't this add the extra step of defining everything in `secrets.yml` or do you believe all the variables should be defined in `secrets.yml`?

PR #1516, opened 2019-02-22
Due date is not red for today anymore [SCI-3069]

jbargu:
Jira ticket:
[SCI-3069](https://biosistemika.atlassian.net/browse/SCI-3069)

### What was done
Change due_date to be shifted 1 day forward

#### ToDo:
Maybe better solution would be to change time to 23:59:59 on due_date
update, but would require migration. I can do it this way as well, since
I don't like this solution.

### DIFF
--- a/app/models/my_module.rb
+++ b/app/models/my_module.rb
@@ -248,14 +248,14 @@ class MyModule < ApplicationRecord
   end
 
   def is_overdue?(datetime = DateTime.current)
-    due_date.present? and datetime.utc > due_date.utc
+    due_date.present? and datetime.utc > due_date_shifted
   end
 
   def overdue_for_days(datetime = DateTime.current)
-    if due_date.blank? or due_date.utc > datetime.utc
-      return 0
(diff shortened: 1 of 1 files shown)


jbargu:
Cause I'm dumb. Fixed it with the new commit.

app/models/my_module.rb
```diff
@@ -248,14 +248,14 @@ def number_of_samples
   end
 
   def is_overdue?(datetime = DateTime.current)
-    due_date.present? and datetime.utc > due_date.utc
+    due_date.present? && datetime.utc > due_date.end_of_day
```
radiokills: I'm not completely sure here, so this is more of a question: shouldn't we compare 2 UTC dates here? Or is this on purpose because we're considering some local date?

jbargu: No, you're completely right, I'm being sloppy here, just did a quick search and replace.

jbargu: My apologizes for taking your time.

PR #1523, opened 2019-02-25
Activity model update (original JIRA issue is migration for lagacy activity)[3026]

urbanrotnik:
Jira ticket: [SCI-3026](https://biosistemika.atlassian.net/browse/SCI-3026)

### What was done
Ok, done everything else than JIRA's content 
- Model update
- set indexes for enums
### Note:
- We (dev team with Matjaz) decided that we won't migrate  old activities to the new "format"
It's hard to figure out what is subject and what are values. 
- At this point, we upgrade modal to accept both types of activities (old and new). The difference is if the activity has a `subject` association (new way) or not (old way). 

relates to [SCI-3025](https://github.com/biosistemika/scinote-web/pull/1494)


app/models/activity.rb
```diff
@@ -1,135 +1,32 @@
...
-
-  validates :type_of, presence: true
+class Activity < ApplicationRecord
+  serialize :values, Hash
+  enum type_of: Extends::ACTIVITY_TYPES
 
-  belongs_to :project, inverse_of: :activities
+  belongs_to :owner, inverse_of: :activities, class_name: 'User'
+  belongs_to :subject, polymorphic: true, optional: true
+  belongs_to :project, inverse_of: :activities # for permissions
```
urbanrotnik: @okriuchykhin  
Correct me if I am wrong. 
In the current implementation of activities, `project` behaves sometimes as a `subject` and sometimes is there just for permissions check? Is that true? 

So are we going to store `project` from now on in `values` (when is a part of action but not main) or in `subject`(when is a main "actor") AND in `project` association always for permission check?
Please correct me if I am wrong.  
cc @radiokills

app/models/activity.rb
```diff
@@ -1,135 +1,32 @@
...
-    project.users.each do |project_user|
-      next if project_user == user
-      next if !project_user.assignments_notification &&
-              notification.type_of == 'assignment'
-      next if !project_user.recent_notification &&
-              notification.type_of == 'recent_changes'
-      UserNotification.create(notification: notification, user: project_user)
+  def activity_version
+    if (experiment || my_module) && subject
+      errors.add(:activity, 'wrong comination of associations')
```
okriuchykhin: typo in combination

app/models/activity.rb
```diff
@@ -1,135 +1,32 @@
...
-    :unassign_sample,
-    :complete_task,
-    :uncomplete_task,
-    :assign_repository_record,
-    :unassign_repository_record
-  ]
-
-  validates :type_of, presence: true
+class Activity < ApplicationRecord
+  serialize :values, Hash
```
okriuchykhin: Aren't we using JSON here?

app/models/activity.rb
urbanrotnik: The `project` is optional, the `team` is required always. Both are for permissions check.

app/models/activity.rb
urbanrotnik: jp. removed.

app/models/team.rb
```diff
@@ -34,6 +34,7 @@ class Team < ApplicationRecord
   has_many :tiny_mce_assets, inverse_of: :team, dependent: :destroy
   has_many :repositories, dependent: :destroy
   has_many :reports, inverse_of: :team, dependent: :destroy
+  has_many :activities
```
okriuchykhin: May be it makes sense to add dependent: :destroy

app/models/activity.rb
```diff
@@ -1,135 +1,30 @@
...
+  belongs_to :owner, class_name: 'User'
+  belongs_to :subject, polymorphic: true, optional: true
 
-  # Depricated in SCI-3025
-  belongs_to :experiment, inverse_of: :activities, optional: true
-  # Depricated in SCI-3025
-  belongs_to :my_module, inverse_of: :activities, optional: true
-  belongs_to :owner, inverse_of: :activities, class_name: 'User'
+  belongs_to :project, optional: true # for permissions
+  belongs_to :team # for permissions
```
okriuchykhin: May be add , inverse_of: :activities ?

PR #1526, opened 2019-02-26
Add main global activities bussiness logic [SCI-3014]

okriuchykhin:
Jira ticket: [SCI-3014](https://biosistemika.atlassian.net/browse/SCI-3014)

### What was done
Added business logic for global notifications


app/services/activities_service.rb
```diff
@@ -0,0 +1,58 @@
...
+  def self.load_activities(team_ids, filters = {})
+    if filters[:subjects].present?
+      first_type, first_ids = filters[:subjects].first
+      query = Activity.where(subject_type: first_type, subject_id: first_ids)
+      if filters[:subjects].count > 1
+        filters[:subjects].except(first_type).each do |type, ids|
+          query = query.or(Activity.where(subject_type: type, subject_id: ids))
+        end
+      end
+    else
```
aignatov-bio: I think it can be simplify in this way:
query_where = filters[:subjects].map{ |v| "(subject_type = ? AND subject_id = ?)" }.join(" OR ")
query=Activity.where(query_where,filters[:subjects].map{|v| [v.subject_type, v.subject_id]}.flatten)

okriuchykhin: The working code will look like this:

query = Activity.where(
        filters[:subjects].map { '(subject_type = ? AND subject_id IN(?))' }
                          .join(' OR '),
        *filters[:subjects].flatten
      )

I doubt it will be more readable

PR #1527, opened 2019-02-26
Move templates project creation for new team to delayed job [SCI-3101]

okriuchykhin:
Jira ticket: [SCI-3101](https://biosistemika.atlassian.net/browse/SCI-3101)

### What was done
Move templates project creation for new team to delayed job

### DIFF
--- a/app/services/templates_service.rb
+++ b/app/services/templates_service.rb
@@ -32,13 +32,16 @@ class TemplatesService
     return unless owner.present?
     updated = false
     exp_tmplt_dir_prefix = "#{@base_dir}/experiment_"
-    existing = tmpl_project.experiments.where.not(uuid: nil).pluck(:uuid)
-    @experiment_templates.except(*existing).each_value do |id|
-      importer_service = TeamImporter.new
-      importer_service.import_experiment_template_from_dir(
-        exp_tmplt_dir_prefix + id.to_s, tmpl_project.id, owner.id
-      )
-      updated = true
+    # Create lock in case another worker starts to update same team
+    tmpl_project.with_lock do
+      existing = tmpl_project.experiments.where.not(uuid: nil).pluck(:uuid)
+      @experiment_templates.except(*existing).each_value do |id|
+        importer_service = TeamImporter.new
+        importer_service.import_experiment_template_from_dir(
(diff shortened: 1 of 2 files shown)


app/services/templates_service.rb
```diff
@@ -32,13 +32,16 @@ def update_team(team)
...
     exp_tmplt_dir_prefix = "#{@base_dir}/experiment_"
-    existing = tmpl_project.experiments.where.not(uuid: nil).pluck(:uuid)
-    @experiment_templates.except(*existing).each_value do |id|
-      importer_service = TeamImporter.new
-      importer_service.import_experiment_template_from_dir(
-        exp_tmplt_dir_prefix + id.to_s, tmpl_project.id, owner.id
-      )
-      updated = true
+    # Create lock in case another worker starts to update same team
+    tmpl_project.with_lock do
```
jbargu: Is this lock necessary? We had a discussion with Miha yesterday regarding the demo projects as well. I'm just wondering, isn't it guaranteed that workers will take jobs without causing any race conditions? I've read that workers can be on different computers, as long as they use the same DB and their clock is in sync.

If it is necessary, shouldn't it then be `team` locked instead? Because if we have a Team without template projects, 2 workers could take it, the first one context-switching after line 17, then the second one takes the same team and checks for existence of a template projects. Then the first worker switches back. The template project does not exist, so both workers create the template project. The lock does not help, because two different projects are created.

I'm not 100% certain if the lock are necessary though, if you have any docs on the issue, I'd be happy to read it.

okriuchykhin: Templates sync code can be run not only as DJ but as scheduled task in separate thread or manually with rake task, so there could be a little chance that this code can be executed simultaneously.
Project has validation for name uniqueness inside one team, so chances of creating 2 same projects are very low.
And how do you imagine locking the whole team for 15-30 seconds especially when we use 'touch' on the models?

jbargu: Alright thank you. Would putting `update_all_templates` into DJ as well cause any problems if it's in another thread?

okriuchykhin: It will be run as periodic task on smaller data sets and started manually as rake task on big, there is no need to put it to DJ

PR #1529, opened 2019-02-27
New inline tags editing in task [SCI 3093]

aignatov-bio:
Jira ticket: https://biosistemika.atlassian.net/browse/SCI-3093

### What was done
Made new tags editing to task view


urbanrotnik:
Tested on chrome, looks good. But on safari: 
- When you open modal `manage_tags` and close it back all tags disappear except first one
- When you get `searching..` line in the dropdown, text jumping a pixle up n down
- Cursor is sometimes not centered... 

I am not sure if we should spend time on this so I won't request the changes. Maybe we can add a ticket to backlog? @radiokills @Ducz0r

PR #1532, opened 2019-02-27
added uniquness validator and index on user_projects and user_teams r…

radiokills:
Jira ticket: [SCI-3031](https://biosistemika.atlassian.net/browse/SCI-3031)

### What was done
added uniquness validator and index on user_projects and user_teams relation


jbargu:
Looks great, I tested it and it works as well. Merge away.

PR #1541, opened 2019-02-28
Duplicates name protocol bug and rake task to rename duplicates [SCI-3047]

aignatov-bio:
Jira ticket: https://biosistemika.atlassian.net/browse/SCI-3047

### What was done
Problem was that, if you copy protocol from task to protocol repository and double click on submit button, it will run 2 ajax requests that create 2 records with same name. Solution is to block button after first click.

### DIFF
--- a/lib/tasks/rename_duplicates_protocols.rake
+++ b/lib/tasks/rename_duplicates_protocols.rake
@@ -23,7 +23,7 @@ namespace :protocols do
     end
 
     # checking for private protocols
-    private_protocols = Protocol.select(:name, :team_id)
+    private_protocols = Protocol.select(:name, :team_id, :added_by_id)
                                 .where.not(name: nil).where(protocol_type: 2)
                                 .group(
                                   :name,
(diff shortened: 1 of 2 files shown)


lib/tasks/rename_duplicates_protocols.rake
```diff
@@ -0,0 +1,74 @@
...
+                               .group(:name, :team_id, :protocol_type)
+                               .having('COUNT(*) > 1')
+    public_protocols.each do |dup_name|
+      protocols_to_update = Protocol.where(
+        name: dup_name.name,
+        team_id: dup_name.team_id,
+        protocol_type: 3
+      )
+      protocols_to_update.each_with_index do |protocol, index|
+        next if index.zero?
```
Ducz0r: Just double-checking, is it really okay to assume that the first protocol (`index.zero?`) is the `dup_name` protocol? Basically, we're relying on the default ordering being the same throughout these calls. Would it be safer to also enforce ordering on all `each`, `each_with_index` calls?

aignatov-bio: Add `.order(created_at: :asc)` to each `each_with_index`

Ducz0r: I've just realized that it doesn't really matter 😲.

I mean, we might _not_ always keep the first created protocol with non-renamed name, but that is still totally acceptable, we are only making sure our database state is clean.

So I think this one can be merged 🌈. We don't need to write `rspec` tests for this I believe.

PR #1544, opened 2019-02-28
Assigned items has lost blue dot when using filter [SCI-3087]

mz3944:
Jira ticket: [SCI-3087](https://biosistemika.atlassian.net/browse/SCI-3087)

### What was done
Fix fetching assigned repository rows when using filter.
Also, the query logic was refactored, to make it more readable, DRY and shorter/efficient.

### DIFF
--- a/app/services/repository_datatable_service.rb
+++ b/app/services/repository_datatable_service.rb
@@ -25,39 +25,42 @@ class RepositoryDatatableService
   end
 
   def process_query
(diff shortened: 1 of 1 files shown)


mz3944:
I was wondering why it said
`using distinct raises an error when combined with sort`
since the sorting worked - it just didn't work on custom columns. I updated the comment accordingly.

PR #1548, opened 2019-03-01
Improve speed of experiment import [SCI-3107]

okriuchykhin:
Jira ticket: [SCI-3107](https://biosistemika.atlassian.net/browse/SCI-3107)

### What was done
Improve speed of experiment import, used in templates service


app/services/team_importer.rb
```diff
@@ -217,14 +219,18 @@ def update_smart_annotations_in_project(project)
     project.project_comments.each do |comment|
       comment.save! if update_annotation(comment.message)
     end
-    project.experiments.each do |exp|
+    project.experiments
+           .preload(my_modules: [{ protocols: { steps: :step_comments } },
```
urbanrotnik: Any specific reason why you are using `preload` and not `include`?

PR #1549, opened 2019-03-01
Prepare a mechanism to schedule the template creation at user login

radiokills:
Jira ticket: [SCI-3119](https://biosistemika.atlassian.net/browse/SCI-3119)

### What was done
* added method to template service
* added callback to login action


aignatov-bio:
I think also need check how it will works with Common addon, cause there is some GDPR actions.

PR #1551, opened 2019-03-04
Global activities creating service [SCI-3103]

urbanrotnik:
Jira ticket: [SCI-3103](https://biosistemika.atlassian.net/browse/SCI-3103)

### What was done
- Service object for creating activities
- Tests for creating activities with different types of `message_items` - values 
- Add message_items params for call in excel document 

#### ToDo 
- @mlorb needs to finish her part
- add new `type_ofs`
- Check and update `.yml` strings where we need

#### Note:
This is part of reimplementation for creating activities.
I was trying to make service as flexible as we need to store different objects with different values and always have a fallback value (string value) for message item in case it's been deleted. 

If the code looks complex/ is not understandable enough, I can add some comments in the service file. Maybe is easier to understand with `rspec` descriptions.

### DIFF
--- a/spec/controllers/projects_controller_spec.rb
+++ b/spec/controllers/projects_controller_spec.rb
@@ -119,7 +119,7 @@ describe ProjectsController, type: :controller do
       end
 
       it 'never calls create activity service' do
-        expect(Activities::CreateActivityService).not_to receive(:call)
+        expect(Activities::CreateActivityService).to receive(:call)
 
         post :create, params: params, format: :json
       end
--- a/spec/controllers/projects_controller_spec.rb
+++ b/spec/controllers/projects_controller_spec.rb
@@ -117,6 +117,12 @@ describe ProjectsController, type: :controller do
         expect(response).to have_http_status(:unprocessable_entity)
         expect(response.content_type).to eq 'application/json'
       end
+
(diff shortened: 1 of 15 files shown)


urbanrotnik:
Thanks, @aignatov-bio for your comments. 
- I like the idea about checks, validating. Based on `type_of` we can check if we have the right `subject` and `message_items`. We can do this inside of `CreateActivityService` with `ActiveRecrodValidations` but I would add this as an improvement. 
- For message items, we already storing `type`, `id`, `value` and `value_for` what is actually getter method for value from an object with that `id`. 

@radiokills @Ducz0r

urbanrotnik:
@aignatov-bio rebased, conflicts solved.

spec/controllers/projects_controller_spec.rb
```diff
@@ -117,6 +117,12 @@
         expect(response).to have_http_status(:unprocessable_entity)
         expect(response.content_type).to eq 'application/json'
       end
+
+      it 'never calls create activity service' do
```
radiokills: Is this named correctly?

urbanrotnik: Nice catch. This was implemented just for testing, will be fixed in the next branch/PR

PR #1558, opened 2019-03-08
Protocol Description in Protocol Management becomes rich text field and is more exposed in U I [SCI-3073]

mz3944:
Jira ticket: [SCI-SCI-3073](https://biosistemika.atlassian.net/browse/SCI-SCI-3073)

### What was done
_Implemented loading of recently used repository protocols into task, and changed protocol header._

### Note:
_Buttons for the completing and locking of the task were removed, since they'll be moved up, above task details. I didn't leave them in, since it would ruin the expected UI. So this PR should be merged together when the part above protocol task is finished._

### DIFF
--- a/app/controllers/my_modules_controller.rb
+++ b/app/controllers/my_modules_controller.rb
@@ -312,6 +312,10 @@ class MyModulesController < ApplicationController
     ).limit(
       Constants::RECENT_REPOSITORIES_SEARH_LIMIT
     )
+    # Allow loading from recently used repository protocols only if protocol
+    # was unchanged and there are actually any such protocols available
+    @show_recent_protocols = @protocol.updated_at == @protocol.created_at and
+      @recent_protocols.present?
     current_team_switch(@protocol.team)
   end
 
--- a/app/controllers/my_modules_controller.rb
+++ b/app/controllers/my_modules_controller.rb
@@ -299,6 +299,19 @@ class MyModulesController < ApplicationController
 
   def protocols
     @protocol = @my_module.protocol
+    @recent_protocols = Protocol.where(
+      team: current_team,
+      protocol_type: Protocol.protocol_types[:in_repository_public]
+    ).or(
+      Protocol.where(
+        added_by: current_user,
+        protocol_type: Protocol.protocol_types.values_at(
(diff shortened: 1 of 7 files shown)


okriuchykhin:
I think it should go to features/note-section and not master

app/controllers/my_modules_controller.rb
```diff
@@ -256,6 +256,23 @@ def update
 
   def protocols
     @protocol = @my_module.protocol
+    @recent_protocols = Protocol.where(
+      team: current_team,
+      protocol_type: Protocol.protocol_types[:in_repository_public]
+    ).or(
+      Protocol.where(
```
okriuchykhin: Do we allow loading protocols from other teams?

app/views/my_modules/protocols/_protocol_buttons.html.erb
```diff
@@ -26,18 +26,24 @@
...
+      <a href="#" title="<%=t "my_modules.protocols.buttons.load_protocol_from_recent_repository" %>" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+        <span class="hidden-sm hidden-md"><%=t "my_modules.protocols.buttons.load_protocol_from_recent_repository" %></span>
+        <span class="hidden-xs hidden-lg fas fa-list-alt"></span>
+        <span class="caret"></span>
+      </a>
+      <ul class="dropdown-menu pull-right">
+        <% @recent_protocols.each do |repo_protocol| %>
+          <li>
+            <%= link_to load_from_repository_protocol_path(@protocol, source_id: repo_protocol.id, format: :json), method: :post, remote: true, data: { action: "load-recent-from-repository" }, title: t('my_modules.protocols.buttons.load_protocol_from_recent_repository_title', protocol_name: repo_protocol.name) do %>
+              <i class="fas fa-file-alt"></i><%= truncate(repo_protocol.name) %>
```
okriuchykhin: I think it should have grey color

app/controllers/my_modules_controller.rb
mz3944: Well I assumed we do, if user has access to them. @radiokills ?

PR #1560, opened 2019-03-08
Change line length in rubocop to 120

Ducz0r:
Jira ticket: (none)

### What was done

Per @urbanrotnik, he suggested a more senseful line length limit would be 120 characters (than the current 80). Because now, any method parametrized with hashes is quickly more than 80 characters long.

#### ToDo:

Fix some other troubling issues with the Rubocop warnings.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -363,7 +363,7 @@ Metrics/CyclomaticComplexity:
   Enabled: false
 
 Metrics/LineLength:
-  Max: 80
+  Max: 120
   AllowHeredoc: true
   AllowURI: true
   URISchemes:


urbanrotnik:
I think I just spend too much time thinking about how to nicely split lines because many function calls and 1 line `if`s  are longer than 80chars. 
And with many splits, it becomes less readable IMO.

Ducz0r:
@jbargu, can you comment on this - I don't like your confused emoji? 😄

jbargu:
@Ducz0r As a terminal purist this change aches my heart, but I understand my argument against the masses is void. :D

radiokills:
Can we merge this then?

Ducz0r:
@jbargu, I hope you find in your ❤️ of 💕 , to forgive me. For the greater good!

PR #1561, opened 2019-03-11
Improve the image file upload into RTE input fields [SCI-3131]

aignatov-bio:
Jira ticket: https://biosistemika.atlassian.net/browse/SCI-3131

### What was done
1) Every time when TinyMCE field initializing, than will be create field  #tiny-mce-images, which one store all images ids.
2) After adding new image, will be created new delay job, which delete image in one day, if image will not be saved.
3) Each time image record will be updated, than will be updated description of related object.
4) New concern that must be included to every model, where we using TinyMCE.
5) Images now resizeable.
6) Visual improvements of image uploading modal window.
7) Other small fixes.


db/schema.rb
```diff
@@ -10,7 +10,11 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
+<<<<<<< HEAD
```
okriuchykhin: Some artifacts here

app/assets/javascripts/sitewide/tiny_mce_file_upload_plugin.js.erb
```diff
@@ -36,6 +37,20 @@
         }, {
           plugin_url: url
         });
+        // Let's make image selection looks fancy
+        $('<div class="image-selection-container">'
+          +'<div class="select_button btn btn-primary">Choose file</div>'
```
okriuchykhin: Probably we need to use translation here

app/assets/javascripts/sitewide/tiny_mce_file_upload_plugin.js.erb
```diff
@@ -36,6 +37,20 @@
...
+          +'<input type="text" placeholder="No file chosen" disabled></input>'
```
okriuchykhin: And here

app/models/tiny_mce_asset.rb
```diff
@@ -25,16 +30,46 @@ class TinyMceAsset < ApplicationRecord
...
+  def self.update_images(object, images)
+    images = JSON.parse(images)
+    current_images = object.tiny_mce_assets.pluck(:id)
+    images_to_delete = current_images.reject { |x| (images.include? Base62.encode(x)) }
+    images.each do |image|
+      image_to_update = find_by_id(Base62.decode(image))
+      image_to_update&.update(object: object, saved: true)
+    end
+    where(id: images_to_delete).destroy_all
+  rescue StandardError
```
okriuchykhin: At least log error message
Rails::logger.error e.message

app/controllers/steps_controller.rb
```diff
@@ -32,7 +31,7 @@ def new
   def create
     @step = Step.new(step_params)
     # gerate a tag that replaces img tag in database
-    @step.description = parse_tiny_mce_asset_to_token(@step.description, @step)
+    @step.description = @step.description
```
okriuchykhin: I think it is not needed here

app/models/concerns/tiny_mce_images.rb
```diff
@@ -0,0 +1,34 @@
...
+             class_name: :TinyMceAsset,
+             dependent: :destroy
+
+    def prepare_for_report(field)
+      images = tiny_mce_assets
+      description = self[field]
+      images.each do |image|
+        tmp_f = Tempfile.open(image.image_file_name, Rails.root.join('tmp'))
+        begin
+          image.image.copy_to_local_file(:large, tmp_f.path)
```
okriuchykhin: A bit misleading naming, may be it is better to use tiny_mce_asset instead of image?

app/models/concerns/tiny_mce_images.rb
```diff
@@ -0,0 +1,34 @@
...
+  included do
+    has_many :tiny_mce_assets,
+             as: :object,
+             class_name: :TinyMceAsset,
+             dependent: :destroy
+
+    def prepare_for_report(field)
+      images = tiny_mce_assets
+      description = self[field]
+      images.each do |image|
```
okriuchykhin: You can use tiny_mce_assets straight away here instead of images variable

app/models/result_text.rb
```diff
@@ -1,9 +1,11 @@
...
+  include TinyMceImages
+  
   auto_strip_attributes :text, nullify: false
   validates :text,
             presence: true,
             length: { maximum: Constants::RICH_TEXT_MAX_LENGTH }
   validates :result, presence: true
   belongs_to :result, inverse_of: :result_text, touch: true, optional: true
-  has_many :tiny_mce_assets, inverse_of: :result_text, dependent: :destroy
+  #has_many :tiny_mce_assets, inverse_of: :result_text, dependent: :destroy
```
okriuchykhin: Can we delete it?

app/models/step.rb
```diff
@@ -30,7 +31,7 @@ class Step < ApplicationRecord
   has_many :tables, through: :step_tables
   has_many :report_elements, inverse_of: :step,
     dependent: :destroy
-  has_many :tiny_mce_assets, inverse_of: :step, dependent: :destroy
+  #has_many :tiny_mce_assets, inverse_of: :step, dependent: :destroy
```
okriuchykhin: Can we delete it?

app/models/tiny_mce_asset.rb
```diff
@@ -1,15 +1,20 @@
+# frozen_string_literal: true
+
 class TinyMceAsset < ApplicationRecord
   attr_accessor :reference
   before_create :set_reference, optional: true
-  after_create :update_estimated_size
+  after_create :update_estimated_size, :self_destroyer
```
okriuchykhin: self- destruct is more correct name for this

app/models/tiny_mce_asset.rb
```diff
@@ -65,21 +100,41 @@ def open
     end
   end
 
+  def delete_unsaved_image
+    destroy unless saved
```
okriuchykhin: I think it is better to implement it as class method with asset id as argument and handle inside it situation when this file was already deleted, otherwise it might end up as failed DJ

db/schema.rb
```diff
@@ -10,7 +10,11 @@
...
 ActiveRecord::Schema.define(version: 20190304153544) do
+=======
```
okriuchykhin: and here

app/models/tiny_mce_asset.rb
```diff
@@ -21,20 +28,55 @@ class TinyMceAsset < ApplicationRecord
   validates_attachment :image,
                        presence: true,
                        size: {
-                         less_than: Rails.configuration.x.file_max_size_mb.megabytes
+                         less_than: Rails.configuration.x\
```
okriuchykhin: is it correct?

app/assets/javascripts/sitewide/tiny_mce_file_upload_plugin.js.erb
```diff
@@ -237,9 +263,18 @@
...
+        updateActiveImages(ed)
+      })
+      var textAreaElement=$('#'+ed.id)
+      textAreaElement.parent().find('input#tiny-mce-images').remove()
+      $('<input type="hidden" id="tiny-mce-images" name="'+textAreaElement[0].name.split('[')[0]+'[tiny_mce_images]" value="[]">').insertAfter(textAreaElement)
+      }
+
+
+  });
+
```
okriuchykhin: Some lines in this file miss semicolons and indication

app/assets/stylesheets/hooks/tinymce.scss
```diff
@@ -0,0 +1,37 @@
+@import "constants";
+@import "mixins";
+
+.mce-image-loader {
+  display: none !important;
+
+  + .image-selection-container {
```
okriuchykhin: is it a typo?

app/controllers/api/v1/results_controller.rb
```diff
@@ -95,7 +96,7 @@ def tiny_mce_asset_params
         end
         file_tokens = prms.map { |p| p[:attributes][:file_token] }
         result_text_params[:text].scan(
-          /\[~tiny_mce_id:(\w+)\]/
+          /data-token="(\w+)\"/
```
okriuchykhin: I think there is no need to escape second double quote

app/assets/stylesheets/hooks/tinymce.scss
aignatov-bio: No, i need only next element

app/models/tiny_mce_asset.rb
aignatov-bio: Yeah, \ is like continue on next line.

"a" \
"b" \
"c"
=> "abc"

spec/requests/api/v1/results_controller_spec.rb
```diff
@@ -172,6 +174,7 @@
         experiment_id: @valid_experiment,
         task_id: @valid_task
       ), params: @valid_tinymce_hash_body.to_json, headers: @valid_headers
+      p response
```
okriuchykhin: is it needed here?

okriuchykhin:
So what are the benefits of storing rendered img tag with src inside of text field? Do we still need to re-create it on each time when we render text field?

aignatov-bio:
We need store whole image tag cause TinyMCE save all transformation to this tag. As i understand we need recreate src link for S3 images.

okriuchykhin:
Not only for S3, local links are hashed as well.

So basically image itself is not modified/resized so it's appearance is only changed using CSS?

aignatov-bio:
Yes, TinyMCE add CSS params to image:

``

I think we can remove `src` attribute on save and add it on read. But everything else we need to keep as is.

okriuchykhin:
Why can't we just resize image on save based on tag attributes?

okriuchykhin:
Will you implement clean up of SRC value in this PR?

aignatov-bio:
Yes, i will make it, when finish with global activities and image annotation tasks.

PR #1567, opened 2019-03-15
[SCI-3166] Finalize global activities front end

aignatov-bio:
https://biosistemika.atlassian.net/browse/SCI-3166

### What was done
Combined global activities front end and fixed bugs.


app/controllers/global_activities_controller.rb
```diff
@@ -57,8 +61,16 @@ def search_subjects
   private
 
   def activity_filters
+    begin
+      params[:types] = JSON.parse(params[:types])
+      params[:users] = JSON.parse(params[:users])
+      params[:teams] = JSON.parse(params[:teams])
+      params[:subjects] = JSON.parse(params[:subjects])
+    rescue StandardError
+    end
```
radiokills: Why  this?

radiokills: I mean: Rescuing from StandardError without any action can make debugging unpleasant :). And what could possibly happen here that needs this kind of error handling?

aignatov-bio: Theoretically this will never raise error cause we are creating all arrays... currently i don't know what kind of error handling can be performed here.

app/models/activity.rb
```diff
@@ -37,11 +38,45 @@ def self.activity_types_list
...
+  def generate_breadcrumbs
+    generate_breadcrumb subject if subject
   end
 
   private
 
+  def generate_breadcrumb(subject)
+    case subject
+    when Protocol
+      breadcrumbs[:protocol] = subject.name
```
aignatov-bio: You can simplify all blocks moving breadcrumbs out from case:
`breadcrumbs[subject.name.underscore.to_sym] = subject.name`
And after this group them by breadcrumb types.

urbanrotnik: After discussion with @aignatov-bio, we will leave as is for now. 
- When we want to save something else than `.name` from specific `subject` we will have "problems" 
- We can't just move out of switch because than we not covering `else` case than

PR #1585, opened 2019-03-25
removing activity creation from demo project

radiokills:
### What was done
Removed creation of activity from demo project.


aignatov-bio:
Merge it

PR #1588, opened 2019-03-26
Adding activity groups [SCI-3203]

aignatov-bio:
Jira ticket: https://biosistemika.atlassian.net/browse/SCI-3202

### What was done
Added activity groups

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -29,12 +29,25 @@ class Activity < ApplicationRecord
   )
 
   def self.activity_types_list
-    type_ofs.map  do |key, value|
-      {
-        id: value,
-        name: key.tr('_', ' ').capitalize
-      }
-    end.sort_by { |a| a[:name] }
+    activity_list = type_ofs.map do |key, value|
+      [
+        key.tr('_', ' ').capitalize,
+        value
+      ]
+    end.sort_by { |a| a[0] }
+    activity_groups = Extends::ACTIVITY_GROUPS
+
+    result = {}
+
+    activity_groups.each do |key, activities|
+      group_name = key.to_s.tr('_', ' ').capitalize
+      result[group_name] = []
+      activities.each do |activity_id|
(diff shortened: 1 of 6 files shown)


app/models/activity.rb
```diff
@@ -29,12 +29,25 @@ class Activity < ApplicationRecord
...
+
+    activity_groups.each do |key, activities|
+      group_name = key.to_s.tr('_', ' ').capitalize
+      result[group_name] = []
+      activities.each do |activity_id|
+        activity_hash = activity_list.select { |activity| activity[1] == activity_id }[0]
+        result[group_name].push(activity_hash) if activity_hash
+      end
+    end
+    result
```
urbanrotnik: Isn't ActivityGroup simply a type_of 'subject'?

app/models/activity.rb
```diff
@@ -78,8 +91,6 @@ def generate_breadcrumb(subject)
   end
 
   def activity_version
-    if (experiment || my_module) && subject
-      errors.add(:activity, 'wrong combination of associations')
-    end
+    errors.add(:activity, 'wrong combination of associations') if (experiment || my_module) && subject
```
urbanrotnik: So good to have more than 80 chars per line 😄 👍

config/initializers/extends.rb
```diff
@@ -199,4 +199,18 @@ class Extends
...
+    task: [*8..14, 35, 36, 37, 53, 54, *58..69],
+    task_protocol: [*15..22, 45, 46, 47, 38, 39, 100],
+    task_inventory: [55, 56],
+    experiment: [*27..31, 57],
+    reports: [48, 49, 50],
+    inventories: [*70..78],
+    protocol_repository: [*79..91],
+    team: [*92..96],
+    exports: [*97..98]
+  }.freeze
```
urbanrotnik: Hm.. Let's check if we can get activity group from 'subject', because this is really hard to maintain

PR #1589, opened 2019-03-26
[SCI-3232] Can not save new/edited text/table result, can not archive text/table result, can not save edited file result.

radiokills:
Jira ticket: [SCI-3232](https://biosistemika.atlassian.net/browse/SCI-3232)
Can not save new/edited text/table result, can not archive text/table result, can not save edited file result.

### What was done
SCI-3232 renaming the attribute name so it doesnt get 'constantized', fixed cringy naming


app/services/activities/create_activity_service.rb
```diff
@@ -66,12 +66,12 @@ def enrich_message_items
     def try_to_constantize(key)
       const_name = key.to_s.camelize
       # Remove last part from and with '_xxx..x'
-      const_name_shorted = key.to_s.sub(/_[^_]*\z/, '').camelize
+      const_name_short = key.to_s.sub(/_[^_]*\z/, '').camelize
```
urbanrotnik: 👍

PR #1613, opened 2019-03-29
Fix unconfirmed users clean-up issues [SCI-3172]

okriuchykhin:
Jira ticket: [SCI-3172](https://biosistemika.atlassian.net/browse/SCI-3172)

### What was done
Fixed issues with deleting unconfirmed users, also add confirmation for users signed up with Linkedin

### DIFF
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -58,6 +58,8 @@ module Users
                             reason: I18n.t('devise.linkedin.failed_to_save'))
           redirect_to after_omniauth_failure_path_for(resource_name) and return
         end
+        # Confirm user
+        @user.update_column(:confirmed_at, @user.created_at)
         redirect_to users_sign_up_provider_path(user: @user)
       end
     end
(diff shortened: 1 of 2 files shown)


app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -58,6 +58,8 @@ def linkedin
                             reason: I18n.t('devise.linkedin.failed_to_save'))
           redirect_to after_omniauth_failure_path_for(resource_name) and return
         end
+        # Confirm user
+        @user.update_column(:confirmed_at, @user.created_at)
```
radiokills: Why are you using update_column? Just curious.

app/models/user.rb
```diff
@@ -209,7 +209,7 @@ class User < ApplicationRecord
 
   has_many :user_notifications, inverse_of: :user
   has_many :notifications, through: :user_notifications
-  has_many :user_system_notifications
+  has_many :user_system_notifications, dependent: :destroy
```
radiokills: :+1:

app/controllers/users/omniauth_callbacks_controller.rb
okriuchykhin: Just to skip validations and callbacks

PR #1614, opened 2019-03-29
Adding missed activities [SCI-3253]

aignatov-bio:
Jira ticket: [SCI-3253](https://biosistemika.atlassian.net/browse/SCI-3253)

### What was done
Adding missed activities


app/controllers/protocols_controller.rb
```diff
@@ -262,6 +262,18 @@ def clone
     end
     respond_to do |format|
       if !cloned.nil?
+
```
urbanrotnik: We can move Creating Activity from controller to model to method `deep_clone_repository`

PR #1619, opened 2019-04-01
Adding preload filters for global activities [SCI-3254]

aignatov-bio:
Jira ticket: [SCI-3254](https://biosistemika.atlassian.net/browse/SCI-3254)

### What was done
Added preload filters for global activities


app/models/activity.rb
```diff
@@ -58,6 +58,25 @@ def generate_breadcrumbs
     generate_breadcrumb subject if subject
   end
 
+  def self.url_search_query(filters)
```
Ducz0r: This is construction of URL GET query. Isn't this always gonna be used on the front-end? In that case, I just think this would be better off being located in some `helper`, e.g. `activity_helper`. What do you think?

app/views/global_activities/index.html.erb
```diff
@@ -28,5 +28,8 @@
   </div>
 </div>
 
+<script>
+  var gaUrlQueryParams= <%= raw @filters.to_json %>
```
Ducz0r: Is there any vulnerability here? We are white-labeling the `activity_filters`, but the values can still be anything, or is that filtered out? Just double-checking, need to be careful when using `raw`.

app/assets/javascripts/global_activities/side_pane.js
```diff
@@ -124,6 +125,34 @@ $(function() {
...
+    }
+    if (filters.users) {
+      $('.ga-side .user-selector select').val(filters.users).trigger('change');
+    }
+    if (filters.from_date) {
+      $('#calendar-from-date').data('DateTimePicker').date(new Date(filters.from_date));
+    }
+    if (filters.to_date) {
+      $('#calendar-to-date').data('DateTimePicker').date(new Date(filters.to_date));
+    }
```
Ducz0r: Just double-check: are these 2 `new Date(string)` calls at any point problematic with regards to date conversion due to time zone conversions? I'd guess not, we will always be providing the date as `YYYY-MM-DD`, which should be future-proof, right?

app/assets/javascripts/global_activities/side_pane.js
```diff
@@ -91,7 +91,7 @@ $(function() {
   };
   // custom display function
   var subjectCustomDisplay = (state) => {
-    return state.label + ': ' + state.text;
+    return (state.label ? state.label + ': ' : '') + state.text;
```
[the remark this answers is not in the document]
aignatov-bio: It's for preload filters, cause when i preload, label will be inside text field with object name.

app/models/activity.rb
aignatov-bio: I think we can use anywhere, not only in front end, but also in back end for redirected links. And you don't need to include it everytime.

app/assets/javascripts/global_activities/side_pane.js
aignatov-bio: Aleksej use this format on backend, so i also use it on front end. I'll prefer to use everywhere unix GMT time and convert it only on front-end to user time zone.

app/views/global_activities/index.html.erb
aignatov-bio: It there is will be wrong format, back end will fail first one, because we except, that only we will generate this params. We need separate task to implement also back end checking. On front end if there is will be wrong format, it simply not set correct filter, but will continue working.

aignatov-bio:
from_date it is max date and to_date it is min date.
Test now types filter is working.
I think we don't need some special testing, because i reuse all filters logic as made before.

Ducz0r:
Okay. Then merge away!

PR #1631, opened 2019-04-02
Fix bug with update protocol from repository and modal disable button [SCI-3284]

aignatov-bio:
Jira ticket: [SCI-3284](https://biosistemika.atlassian.net/browse/SCI-3284)

Fix bug with update protocol from repository and protocol save modal disabled button.


jbargu:
Looks awesome, merge away!

PR #1632, opened 2019-04-02
Fix permission checks for file editing and make protocol preview files read only [SCI-3281]

okriuchykhin:
Jira ticket: [SCI-3281](https://biosistemika.atlassian.net/browse/SCI-3281)

### What was done
Fix permission checks for file editing and make protocol preview files read only

### DIFF
--- a/app/assets/javascripts/sitewide/file_preview.js
+++ b/app/assets/javascripts/sitewide/file_preview.js
@@ -3,13 +3,17 @@
 /* global fabric tui animateSpinner setupAssetsLoading I18n*/
 //= require assets
 
-(function(global) {
+var FilePreviewModal = (function() {
   'use strict';
 
-  global.initPreviewModal = function initPreviewModal() {
+  var readOnly = false;
+
+  function initPreviewModal(options = {}) {
     var name;
     var url;
     var downloadUrl;
+    readOnly = options.readOnly;
+
     $('.file-preview-link').off('click');
     $('.file-preview-link').click(function(e) {
       e.preventDefault();
@@ -18,7 +22,7 @@
       downloadUrl = $(this).attr('href');
       openPreviewModal(name, url, downloadUrl);
     });
-  };
+  }
 
   // Adding rotation icon
   function updateFabricControls() {
@@ -336,7 +340,7 @@
                 .click(function(ev) {
                   ev.stopPropagation();
                 }));
-            if (data.editable) {
+            if (!readOnly && data.editable) {
               modal.find('.file-edit-link').css('display', '');
               modal.find('.file-edit-link').off().click(function(ev) {
                 ev.preventDefault();
@@ -410,4 +414,8 @@
       }
     });
   }
+
+  return Object.freeze({
+    init: initPreviewModal
+  });
 }(window));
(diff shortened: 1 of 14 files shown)


app/assets/javascripts/sitewide/file_preview.js
```diff
@@ -349,8 +353,12 @@
             }
           }
         } else {
+          modal.find('.file-edit-link').css('display', 'none');
           modal.find('.file-preview-container').html(data['preview-icon']);
         }
+        if (readOnly) {
```
jbargu: I think it would be better to send `readOnly` as a parameter to `file-preview-url` and to have this `edit permission` logic all in one place.

app/assets/javascripts/sitewide/file_preview.js
```diff
@@ -349,8 +353,12 @@
...
+          modal.find('#wopi_file_edit_button').remove();
```
jbargu: I don't know what the requirements are but usually the edit button is disabled and the hover on shows the reason for not being able to edit.

app/controllers/assets_controller.rb
```diff
@@ -48,17 +48,24 @@ def file_preview
...
+      'download-url' => download_asset_path(@asset, timestamp: Time.now.to_i)
     }
 
+    can_edit = if @assoc.class == Step
+                 can_manage_protocol_in_module?(@protocol) || can_manage_protocol_in_repository?(@protocol)
+               elsif @assoc.class == Result
+                 can_manage_module?(@my_module)
+               elsif @assoc.class == RepositoryCell
+                 can_manage_repository_rows?(@repository.team)
+               end
```
jbargu: As a piece of mind, would it be better if we had additional 'else false` here? Just thinking about if this could be abused by anyone.

app/models/asset.rb
```diff
@@ -465,13 +465,8 @@ def update_contents(new_file)
...
 
-  def editable?(user)
-    objects = %w(step result)
-    my_module = send(objects.find { |object| send(object) }).my_module
-    Canaid::PermissionsHolder.instance.eval(:manage_experiment, user, my_module.experiment) &&
-      !locked? &&
-      %r{^image/#{Regexp.union(Constants::WHITELISTED_IMAGE_TYPES_EDITABLE)}} ===
-        file.content_type
+  def editable_image?
+    !locked? && %r{^image/#{Regexp.union(Constants::WHITELISTED_IMAGE_TYPES_EDITABLE)}} =~ file.content_type
```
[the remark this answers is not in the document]
okriuchykhin: just kept it for the future use, I think at the moment it is used only in wopi. As it is just model's filed it should not add any slowdown

app/assets/javascripts/sitewide/file_preview.js
okriuchykhin: This parameter is used to disable file editing on the whole page(like protocol preview), and not per file.

app/assets/javascripts/sitewide/file_preview.js
okriuchykhin: This parameter is used to disable file editing on the whole page(like protocol preview).

app/controllers/assets_controller.rb
okriuchykhin: As far as I can remember nil equals to false in 'if' checks, so if can_edit is not set to true, it will fail in latter check

PR #1639, opened 2019-04-03
Remove archive_task from extends.rb [SCI-3288]

jbargu:
Jira ticket: [SCI-3288](https://biosistemika.atlassian.net/browse/SCI-3288)

### What was done
Remove archive_task activity type.

### Note:
I updated activities with `type_of: :archive_task` in my-test instance DB.

### DIFF
--- a/config/initializers/extends.rb
+++ b/config/initializers/extends.rb
@@ -118,7 +118,7 @@ class Extends
     unassign_user_from_project: 7,
     create_module: 8,
     clone_module: 9,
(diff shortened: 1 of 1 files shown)


urbanrotnik:
@jbargu we do not have `archive_task ` activities on prod right?

jbargu:
> 
> 
> @jbargu we do not have `archive_task ` activities on prod right?

I'm not sure @okriuchykhin?

PR #1660, opened 2019-04-10
Date activity filter filters out activity after reaching bottom of the page [SCI-3302]

jbargu:
Jira ticket: [SCI-3302](https://biosistemika.atlassian.net/browse/SCI-3302)

### What was done
Reset scrollbar after selecting date filter


jbargu:
> Maybe put it inside `reloadActivities` function after `animateSpinner(null, false)`.

I'm not sure if this is desired behaviour for everything else? If you're scrolling down for newer activitiies, you don't want to be at the top.

PR #1663, opened 2019-04-11
Migration for old tinymce assets [SCI-3176]

aignatov-bio:
Jira ticket: [SCI-3176](https://biosistemika.atlassian.net/browse/SCI-3176)

### What was done
Migration for old tinymce assets.
Tests for tinymce model.


lib/tasks/migrate_old_tinymce_assets.rake
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+# rubocop:disable Metrics/LineLength
+namespace :tinymce_assets do
+  desc 'Migrate old TinyMCE images to new polymorphic format' \
+       'IT SHOULD BE RUN ONE TIME ONLY'
+  task migrate_tinymce_assets: :environment do
+    old_images = TinyMceAsset.where('step_id IS NOT NULL OR result_text_id IS NOT NULL')
```
radiokills: Maybe you could add additional condition, to have the object null, so we're completely safe if this task is (for whatever reason) run more than onece

spec/models/tiny_mce_asset_spec.rb
```diff
@@ -1,6 +1,19 @@
+# frozen_string_literal: true
+
 require 'rails_helper'
 
 describe TinyMceAsset, type: :model do
+  before :all do
+    @user = create(:user)
```
radiokills: use the let and let! helpers

spec/models/tiny_mce_asset_spec.rb
```diff
@@ -1,6 +1,19 @@
...
+    @teams = create_list(:team, 2, created_by: @user)
+    @valid_project = create(:project, name: Faker::Name.unique.name,
```
radiokills: Isn't the name already included in the factory?

PR #1664, opened 2019-04-11
Disable editing of office file in view mode when module is locked [SCI-2949, 3318]

jbargu:
Jira ticket: [SCI-2949](https://biosistemika.atlassian.net/browse/SCI-2949)
Jira ticket: [SCI-3318](https://biosistemika.atlassian.net/browse/SCI-3318)

This is connected to:
[Electronic signatures PR](https://gitlab.com/biosistemika/scinote/addons/electronic_signatures/merge_requests/37)

### What was done
Expose `@can_write` variable, so it can be overwritten in the addon.

### Note:
For the editing issue, it has to be first pushed onto the pro instance to test it.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,6 +4,12 @@ source 'http://rubygems.org'
 
 ruby '2.4.5'
 
+gem 'scinote_common', path: 'addons/Common'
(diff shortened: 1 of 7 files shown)


jbargu:
I screwed up something went amending... :cry:

PR #1667, opened 2019-04-12
Fix links in activity message for protocols in repository [SCI-3304]

mlorb:
Jira ticket: [SCI-3304](https://biosistemika.atlassian.net/browse/SCI-3304)

### What was done
_Links for protocols in repository activities were fixed._

### Note:
_Review also [electronic signatures](https://gitlab.com/biosistemika/scinote/addons/electronic_signatures/merge_requests/39), were missing activity was added._

### DIFF
--- a/app/views/global_activities/references/_protocol.html.erb
+++ b/app/views/global_activities/references/_protocol.html.erb
@@ -7,9 +7,12 @@
 <% end %>
 <% if subject %>
   <% if subject.in_repository? %>
+    <% type = :public if subject.in_repository_public? %>
+    <% type = :private if subject.in_repository_private? %>
+    <% type = :archive if subject.in_repository_archived? %>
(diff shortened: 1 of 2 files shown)


app/views/global_activities/references/_protocol.html.erb
```diff
@@ -7,9 +7,12 @@
 <% end %>
 <% if subject %>
   <% if subject.in_repository? %>
+    <% type = :public if subject.in_repository_public? %>
+    <% type = :private if subject.in_repository_private? %>
+    <% type = :archive if subject.in_repository_archived? %>
```
radiokills: Would it make sense to put this logint somewhere else, maybe helper method?

radiokills: Maybe we could handle it in `route_to_other_team protocols_path`

PR #1672, opened 2019-04-15
Move experiment image generation to sync job[SCI-3028]

aignatov-bio:
Jira ticket: [SCI-3028](https://biosistemika.atlassian.net/browse/SCI-3028)

### What was done
Move experiment image generation to sync job

### DIFF
--- a/app/models/experiment.rb
+++ b/app/models/experiment.rb
@@ -221,6 +221,12 @@ class Experiment < ApplicationRecord
     true
   end
 
+  def update_workflow_img
+    Timeout.timeout(5) { generate_workflow_img }
+  rescue StandardError
+    delay.generate_workflow_img
+  end
+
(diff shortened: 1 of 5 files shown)


app/models/experiment.rb
```diff
@@ -221,6 +221,12 @@ def update_canvas(
     true
   end
 
+  def update_workflow_img
+    Timeout.timeout(5) { generate_workflow_img }
+  rescue StandardError
+    delay.generate_workflow_img
```
urbanrotnik: Should we log timeout somewhere?

aignatov-bio: I will add some error loging, I think we need to catch any errors, not only timeout.

urbanrotnik: I agree, but we should know what we catch :)

PR #1675, opened 2019-04-15
Fix for exports_left, fix for some edge cases [SCI-3345]

urbanrotnik:
Jira ticket: [SCI-3345](https://biosistemika.atlassian.net/browse/SCI-3345)

### What was done
- Fix for showing `reports_left`, don't care about reports if are older than today
- Add some tests for edge cases

### DIFF
--- a/app/controllers/teams_controller.rb
+++ b/app/controllers/teams_controller.rb
@@ -251,15 +251,13 @@ class TeamsController < ApplicationController
 
   def export_projects_modal
     if @exp_projects.present?
-      limit = (ENV['EXPORT_ALL_LIMIT_24_HOURS'] || 3).to_i
-      curr_num = current_user.export_vars['num_of_export_all_last_24_hours']
       if current_user.has_available_exports?
         render json: {
           html: render_to_string(
             partial: 'projects/export/modal.html.erb',
             locals: { num_projects: @exp_projects.size,
-                      limit: limit,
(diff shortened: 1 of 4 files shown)


app/models/user.rb
```diff
@@ -529,13 +530,20 @@ def increase_daily_exports_counter!
   end
 
   def has_available_exports?
-    limit = (Rails.application.secrets.export_all_limit_24h || 3).to_i
-    last_export = export_vars[:last_export_timestamp]
+    last_export = export_vars[:last_export_timestamp] || 0
```
radiokills: I know this is a bit nitty gritty, but wouldn't it make more sense to name this variable `last_export_timestamp`. The way it is named now it implies we're dealing with whole export, not just timestamp.

radiokills: Or maybe last_export_at

app/controllers/teams_controller.rb
```diff
@@ -251,23 +251,21 @@ def export_projects
...
+                      num_of_requests_left: current_user.exports_left }
           ),
           title: t('projects.export_projects.modal_title')
         }
       else
         render json: {
           html: render_to_string(
             partial: 'projects/export/error.html.erb',
-            locals: { limit: limit }
+            locals: { limit: TeamZipExport.exports_limit }
```
radiokills: :+1:

PR #1686, opened 2019-04-18
Add activity for importing inventory items [SCI-3324]

mlorb:
Jira ticket: [SCI-3324](https://biosistemika.atlassian.net/browse/SCI-3324)

### What was done
_Activity for importing inventory items was added_

### Note:
_Activity export_inventory_items was refactored_

### DIFF
--- a/app/controllers/repositories_controller.rb
+++ b/app/controllers/repositories_controller.rb
@@ -242,10 +242,10 @@ class RepositoriesController < ApplicationController
           import_records = repostiory_import_actions
           status = import_records.import!
 
-          log_activity(:import_inventory_items,
-                       num_of_items: status[:nr_of_added])
-
           if status[:status] == :ok
+            log_activity(:import_inventory_items,
+                         num_of_items: status[:nr_of_added])
+
             flash[:success] = t('repositories.import_records.success_flash',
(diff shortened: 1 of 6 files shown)


app/controllers/repositories_controller.rb
```diff
@@ -242,6 +242,9 @@ def import_records
           import_records = repostiory_import_actions
           status = import_records.import!
 
+          log_activity(:import_inventory_items,
```
urbanrotnik: Do we care if the import was successfully finished? Or just call `log_activity` always?

spec/controllers/repositories_controller_spec.rb
```diff
@@ -100,4 +100,37 @@
...
+    let(:params) do
+      { id: repository.id,
+        team_id: team.id,
+        file_id: 'file_id',
+        mappings: mappings }
+    end
+    let(:action) { post :import_records, params: params, format: :json }
+
+    it 'calls create activity for importing inventory items' do
+      ImportRepository::ImportRecords.any_instance.stub(:import!)
```
urbanrotnik: better syntax here should be someting like: 
`allow_any_instance_of(ImportRepository::ImportRecords).to receive(:import!).and_return(status: :ok)` 
Try with this 1...

PR #1689, opened 2019-04-23
Update protocol repository for new description format [SCI-3371]

aignatov-bio:
Jira ticket: [SCI-3371]](https://biosistemika.atlassian.net/browse/SCI-3371])

### What was done
Update protocol repository for new description format


app/models/tiny_mce_asset.rb
```diff
@@ -163,6 +163,23 @@ def self.save_to_eln(ostream, dir)
...
+        estimated_size: tiny_img.estimated_size,
+        object: target,
+        team: team
+      )
+      tiny_img_clone.save
+
+      target.tiny_mce_assets << tiny_img_clone
+      cloned_img_ids << [tiny_img.id, tiny_img_clone.id]
+    end
+    reload_images(cloned_img_ids)
```
okriuchykhin: I've changed implementation of reload_images in https://github.com/biosistemika/scinote-web/pull/1693

app/models/tiny_mce_asset.rb
```diff
@@ -163,6 +163,23 @@ def self.save_to_eln(ostream, dir)
...
+  def self.clone_assets(source, target, team)
+    cloned_img_ids = []
+    source.tiny_mce_assets.each do |tiny_img|
+      tiny_img_clone = TinyMceAsset.new(
+        image: tiny_img.image,
+        estimated_size: tiny_img.estimated_size,
+        object: target,
+        team: team
+      )
+      tiny_img_clone.save
```
okriuchykhin: Probably makes sense to use save! instead as failed save is not handled here

app/models/tiny_mce_asset.rb
```diff
@@ -163,6 +163,23 @@ def self.save_to_eln(ostream, dir)
     ostream
   end
 
+  def self.clone_assets(source, target, team)
```
okriuchykhin: I would put this method to concerns/tiny_mce_images.rb

PR #1692, opened 2019-04-23
Comments section changes [SCI-3135]

urbanrotnik:
Jira ticket: [SCI-3135](https://biosistemika.atlassian.net/browse/SCI-3135)

### What was done
New comments look.

### Note
Focus out is working (saving) comment. Save and Cancel button has been added above textarea.


urbanrotnik:
@aignatov-bio 
Looks awesome! Just few comments
- When you reload page, list of comments is not scrolled to bottom. 
- Sometimes on selecting (clicking) on textbox for editing list jumps a bit. 
- With Matjaz we discussed and agreed to add blue on hover (I think is the same as somewhere else), but I think this frame around textbox looks even better so let's go with this for now.

PR #1699, opened 2019-04-25
Bulk insert for SystemNotifications, updated ordering [SCI-3346]

urbanrotnik:
Jira ticket: [SCI-3346](https://biosistemika.atlassian.net/browse/SCI-3346)

### What was done
- update ordering (from `last_time_changed_at` to `created_at`) 
- Bulk insert with `active_record-import`
- Add unique constraint for `source_id` (prevent save more than 1 same notification in DB when more than 1 worker) 
- Removed callback `after_commit` and send emails with an explicit call
- Add service object for notifying users. With @radiokills we agreed to add extendable service/class for sending emails and we can reuse it for upcoming feature "notifications".

### Note:
I tested sending (and receiving) notifications on my local machine with 100k users, received 6 notifications at once. Tested also with more than one job at once.

### DIFF
--- a/app/models/system_notification.rb
+++ b/app/models/system_notification.rb
@@ -30,7 +30,7 @@ class SystemNotification < ApplicationRecord
     user,
     query = nil
   )
-    notifications = order(last_time_changed_at: :DESC)
+    notifications = order(created_at: :DESC)
     notifications = notifications.search_notifications(query) if query.present?
     notifications.joins(:user_system_notifications)
                  .where('user_system_notifications.user_id = ?', user.id)
@@ -39,6 +39,7 @@ class SystemNotification < ApplicationRecord
                    :title,
                    :description,
                    :last_time_changed_at,
+                   :created_at,
                    'user_system_notifications.seen_at',
                    'user_system_notifications.read_at'
                  )
--- a/app/models/system_notification.rb
+++ b/app/models/system_notification.rb
@@ -51,6 +51,6 @@ class SystemNotification < ApplicationRecord
     SystemNotification
       .order(last_time_changed_at: :desc)
       .first&.last_time_changed_at&.to_i ||
-      User.order(created_at: :desc).first&.created_at&.to_i
+      User.order(created_at: :asc).first&.created_at&.to_i
   end
 end
--- a/app/services/notifications/sync_system_notifications_service.rb
+++ b/app/services/notifications/sync_system_notifications_service.rb
@@ -81,12 +81,27 @@ module Notifications
             .where(source_id: attrs[:source_id]).first_or_initialize(attrs)
 
         if n.new_record?
-          n.users = User.all
-          n.save!
+          save_notification n
         elsif n.last_time_changed_at < attrs[:last_time_changed_at]
           n.update_attributes!(attrs)
         end
       end
+    end
+
+    def save_notification(notification)
+      ActiveRecord::Base.transaction do
+        notification.save!
+
+        User.find_in_batches do |user_ids|
+          user_system_notifications = user_ids.pluck(:id).collect do |item|
+            Hash[:user_id, item, :system_notification_id, notification.id]
+          end
+          UserSystemNotification.import user_system_notifications, validate: false
+        end
+      end
+
+      Notifications::PushToCommunicationChannelService.delay.call(item_id: notification.id,
+                                                            item_type: notification.class.name)
     end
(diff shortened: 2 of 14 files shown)


app/models/system_notification.rb
```diff
@@ -50,6 +51,6 @@ def self.last_sync_timestamp
     SystemNotification
       .order(last_time_changed_at: :desc)
       .first&.last_time_changed_at&.to_i ||
-      User.order(created_at: :desc).first&.created_at&.to_i
+      User.order(created_at: :asc).limit(1).first&.created_at&.to_i
```
okriuchykhin: doesn't '.first' use LIMIT inside?

app/services/notifications/sync_system_notifications_service.rb
```diff
@@ -81,12 +81,26 @@ def save_new_notifications
...
         end
       end
     end
+
+    def save_notification(notification)
+      ActiveRecord::Base.transaction do
+        notification.save!
+
+        user_system_notifications = []
+        User.pluck(:id).each do |user_id|
```
okriuchykhin: Does it load IDs of all users in the single array?

app/models/system_notification.rb
urbanrotnik: Ahh, sure it does. Thank you!

urbanrotnik: True. For array with 100k users `ObjectSpace.memsize_of(obj)` returns `780 KB`. And limit for items nr in an array should be around 500mio. 
Do you see any issues?

app/services/notifications/sync_system_notifications_service.rb
urbanrotnik: True. For array with 100k users ObjectSpace.memsize_of(obj) returns 780 KB. And limit for items nr in an array should be around 500mio.
Do you see any issues?

radiokills: Use find_each

urbanrotnik: @radiokills It is just an array of ID-s, not active record objects... But yeah... I will.

okriuchykhin: User.in_batches(of: 10) do |user_ids| puts user_ids.pluck(:id) end

urbanrotnik: @okriuchykhin do we care about size of `User.pluck(:id)` or `user_system_notifications` array?

okriuchykhin: Default is 1000, it should be fine

urbanrotnik: On 100k rows in `users`
```
User.all.pluck(:id)
   (35.2ms)  SELECT "users"."id" FROM "users"
```
```
User.in_batches(of: 10) do |user_ids| puts user_ids.pluck(:id) end
   takes: 20seconds
```
```
User.in_batches(of: 10000) do |user_ids| puts user_ids.pluck(:id) end
   10 times around 8ms
```

urbanrotnik: Ok... but what is the idea? To make inside of `in_batches` loop `SystemNotification.import` to prevent to have too big `user_system_notifications[]` or just to have a smaller array of `user_ids`?

okriuchykhin: It is not about the speed, but to keep resource allocation under control

okriuchykhin: To make inside of in_batches loop SystemNotification.import to prevent to have too big user_system_notifications[], user_ids[] will be same size

okriuchykhin: As example:

User.in_batches(of:10) do |user_ids| user_system_notifications = user_ids.pluck(:id).collect { |item| Hash[:user_id, item, :system_notification_id, 33] }; pp user_system_notifications end

okriuchykhin: Better version:

User.find_in_batches(batch_size: 10) do |user_ids| user_system_notifications = user_ids.pluck(:id).collect { |item| Hash[:user_id, item, :system_notification_id, 33] }; pp user_system_notifications end

urbanrotnik:
@radiokills @okriuchykhin 
Can you check changes

PR #1701, opened 2019-04-25
Fix sorting by user name with search parameter in inventories [SCI-3391]

okriuchykhin:
Jira ticket: [SCI-3391](https://biosistemika.atlassian.net/browse/SCI-3391)

### What was done
Fix sorting by user name with search parameter in inventories


radiokills:
fire away!

PR #1705, opened 2019-04-26
Show icons for office files [SCI-3387]

urbanrotnik:
Jira ticket: [SCI-3387](https://biosistemika.atlassian.net/browse/SCI-3387)

### What was done
Update icon_class helper method. Show office files for now (as was before attachments redesign) until we do not have implemented preview office files.

### DIFF
--- a/app/helpers/file_icons_helper.rb
+++ b/app/helpers/file_icons_helper.rb
@@ -4,34 +4,20 @@ module FileIconsHelper
     %w(csv ods xls xlsb xlsm xlsx odp pot potm potx pps ppsm ppsx ppt pptm pptx doc docm docx dot dotm dotx odt rtf).include?(file_ext)
   end
 
-  # For showing next to file
   def file_fa_icon_class(asset)
     file_ext = asset.file_file_name.split('.').last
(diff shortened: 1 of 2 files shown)


app/helpers/file_icons_helper.rb
```diff
@@ -4,34 +4,20 @@ def wopi_file?(asset)
...
+      return 'fa-file-powerpoint'
     elsif %w(pdf).include?(file_ext)
-      fa_class = 'fa-file-pdf'
+      return 'fa-file-pdf'
     elsif %w(txt csv tab tex).include?(file_ext)
-      fa_class = 'fa-file-alt'
-    end
-
-    # Now check for custom mappings or possible overrides
-    if Extends::FILE_ICON_MAPPINGS[file_ext]
```
urbanrotnik: @okriuchykhin 
Was this mistake or I don't get the context here. You are checking `FILE_ICON_MAPPINGS ` but then assigning a value from `FILE_FA_ICON_MAPPINGS ` hash. 
If this is correct code, mine is wrong.

okriuchykhin: Looks like a type, should be FILE_FA_ICON_MAPPINGS

PR #1709, opened 2019-04-26
Fix TinyMCE for results [SCI-3398]

okriuchykhin:
Jira ticket: [SCI-3398](https://biosistemika.atlassian.net/browse/SCI-3398)

### What was done
Multiple fixes for TinyMCE on results page

### DIFF
--- a/app/assets/javascripts/sitewide/tiny_mce_file_upload_plugin.js
+++ b/app/assets/javascripts/sitewide/tiny_mce_file_upload_plugin.js
@@ -300,7 +300,9 @@
         images = $.map($('img', iframe), e => {
           return e.dataset.mceToken;
         });
-        $('#' + editor.id).next()[0].value = JSON.stringify(images);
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/sitewide/tiny_mce_file_upload_plugin.js
```diff
@@ -300,7 +300,9 @@
         images = $.map($('img', iframe), e => {
           return e.dataset.mceToken;
         });
-        $('#' + editor.id).next()[0].value = JSON.stringify(images);
+        if ($('#' + editor.id).next()[0]) {
```
aignatov-bio: Maybe we need some error handling here. Because if we don't find this `next()` element it will mean that we will not save images.

okriuchykhin: I assume that there are no images If it doesn't find next element

PR #1712, opened 2019-04-29
Add new tag activtiies [SCI-3341]

aignatov-bio:
Jira ticket: [SCI-3341](https://biosistemika.atlassian.net/browse/SCI-3341)

### What was done
Added new activities for tags.
Added tests for new activities.

### DIFF
--- a/app/controllers/tags_controller.rb
+++ b/app/controllers/tags_controller.rb
@@ -17,6 +17,7 @@ class TagsController < ApplicationController
     end
 
     if @tag.save
+      log_activity(:create_tag, @tag.project, tag: @tag.id, project: @tag.project.id)
       if params.include? "my_module_id"
         # Assign the tag to the specified module
         new_mmt = MyModuleTag.new(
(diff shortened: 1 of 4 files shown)


app/controllers/tags_controller.rb
```diff
@@ -25,17 +25,9 @@ def create
...
-                subject: my_module,
-                project:
-                  my_module.experiment.project,
-                team: current_team,
-                message_items: {
-                  my_module: my_module.id,
-                  tag: @tag.id
-                })
+
+        log_activity(:create_tag, @tag.project, tag: @tag.id, project: @tag.project.id)
```
radiokills: Shouldn't this be outside `if params.include? "my_module_id"'` block?

config/initializers/extends.rb
```diff
@@ -178,7 +178,7 @@ class Extends
     change_task_due_date: 62,
     remove_task_due_date: 63,
     add_task_tag: 64,
-    edit_task_tag: 65,
+    edit_tag: 65,
```
radiokills: Will we have any problems with existing data?

app/controllers/tags_controller.rb
aignatov-bio: Fix

config/initializers/extends.rb
aignatov-bio: It was create, but never used. And in Activity Excel had flag "not implement". So i rename it along with other tag activities.

PR #1729, opened 2019-05-07
Update protocol preview style [SCI-3416]

aignatov-bio:
Jira ticket: [SCI-3416](https://biosistemika.atlassian.net/browse/SCI-3416)

### What was done
Update protocol preview style. Reuse main step logic.


app/assets/stylesheets/protocols/protocol.scss
```diff
@@ -0,0 +1,32 @@
+// scss-lint:disable SelectorDepth
+// scss-lint:disable NestingDepth
+
+@import "constants";
+@import "mixins";
+
+.protocol-info {
+
+  .well {
```
jbargu: I'm not sure why this styles cannot be used from the `protocols/index.scss'. They seem pretty similar, at least this `.well` is exactly the same. Could this be somehow DRYed up so it does not appear on 2 different places?

app/views/protocols/index/_protocol_preview_modal_body.html.erb
```diff
@@ -68,88 +64,22 @@
...
+<%= javascript_include_tag "numeral" %>
+<%= javascript_include_tag "numeric" %>
+<%= javascript_include_tag "md5" %>
+<%= javascript_include_tag "jstat" %>
+<%= javascript_include_tag "formula" %>
+<%= javascript_include_tag "parser" %>
+<%= javascript_include_tag "ruleJS" %>
+<%= javascript_include_tag "handsontable.formula" %>
+<%= javascript_include_tag "big.min" %>
+<%= stylesheet_link_tag "handsontable.formula" %>
```
jbargu: Would it make sense to DRY up this as well (something like `_formuls_dependencies.html.erb`? It appears on couple of places as well.

app/assets/stylesheets/protocols/protocol.scss
aignatov-bio: DRYed

app/views/protocols/index/_protocol_preview_modal_body.html.erb
aignatov-bio: DRYed

PR #1746, opened 2019-05-10
Changes to create new office file button [SCI-3426]

Ducz0r:
Jira ticket: [SCI-3426](https://biosistemika.atlassian.net/browse/SCI-3426)

### What was done

* Added checks to hide WOPI controls if `WOPI_ENABLED == false` (env. variable)
* Change Office icon for the "New Office" button
* Added filename length validation when creating new Office file
* Refresh the current page after the Office file is created (to show the newly created result/step file)

#### ToDo:
_Validation for file name might be implemented in a better way, but I couldn't manage to hook validation directly onto the model - `asset.file_file_name`._

### Note:
@jbargu I'm curious about my last item on the 'What was done', do you see any issues with the way I've implemented that? (there was already a commented `location.reload()` call there)


app/controllers/assets_controller.rb
```diff
@@ -190,11 +191,25 @@ def create_wopi_file
...
+    if original_filename.length > Constants::FILENAME_MAX_LENGTH
+      render json: {
+        message: {
+          file: I18n.t(
+            'assets.create_wopi_file.errors.file_name_too_long',
+            limit: Constants::FILENAME_MAX_LENGTH
+          )
+        }
+      }, status: 400 and return
+    end
```
jbargu: I believe this could easily be put into `asset.rb` into `:wopi_file_creation` context. See the line 213, it already checks for presence of filename, you'd just have to perform the same check in `name_should_not_be_empty_without_extension` and would be less mouthful and cleaner code, since only the check on 213 would be performed. (probably the method name should be changed then to reflect the changes).

app/assets/images/office/office.svg
```diff
@@ -0,0 +1,11 @@
...
+ <g transform="translate(-216.07358,-549.28882)">
+  <g transform="matrix(1.8232952,0,0,1.8232952,-597.71681,-124.12247)">
+   <g transform="translate(0,-91.137241)">
+    <g fill="#eb3c00" transform="matrix(0.74069815,0,0,0.74069815,98.5698,-8.2505871)">
+     <path d="m469.87,671.03,0-28.52,25.229-9.3238,13.711,4.3877,0,38.392-13.711,4.133-25.229-9.0691,25.229,3.0361,0-33.201-16.454,3.8392,0,22.487z"/>
+    </g>
+   </g>
+  </g>
+ </g>
+</svg>
```
jbargu: I knew it will come back!

app/assets/javascripts/assets/wopi/create_wopi_file.js
```diff
@@ -20,9 +20,8 @@ function initCreateWopiFileModal() {
   $('#new-office-file-modal form')
     .on('ajax:success', function(ev, data) {
       window.open(data.edit_url, '_blank');
-      $('#new-office-file-modal').modal('hide');
-
-      // location.reload();
+      window.focus();
+      location.reload(); // Reload current page, to display the new element
```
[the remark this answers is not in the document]
jbargu: To answer your question. I didn't know what the behaviour should be after clicking the button. I found `location.reload()` to be something of my liking but wanted to hear the opinion. As long as the prompt for opening the file does not disappear (if SciNote blocks popups) then I see no issue.

config/initializers/constants.rb
```diff
@@ -23,6 +23,9 @@ class Constants
   COLOR_MAX_LENGTH = 7
   # Max characters for text in dropdown list element
   DROPDOWN_TEXT_MAX_LENGTH = 15
+  # Max characters limit for (on most operating systems, it's ~255 characters,
+  # but this is with a bit more safety margin)
+  FILENAME_MAX_LENGTH = 100
```
jbargu: Shouldn't this be then a global standard, not only for wopi files?

PR #1753, opened 2019-05-13
Fix report saving [SCI-3443]

aignatov-bio:
Jira ticket: [SCI-3443](https://biosistemika.atlassian.net/browse/SCI-3443)

### What was done
Fix report saving


Ducz0r:
I've pushed my changes directly to @aignatov-bio branch. I believe these changes now follow the reports structure in a better way.

My only concern - that we will need to test - is that currently, this report element is tied to `my_module`. I'm not sure what happens if you e.g. load a new protocol from the repository into the task, does it change the underlying `protocol`, or not. If this protocol always stays the same, this should not be an issue, but if it changes, probably the report element should be deleted, and right now it almost certainly isn't (as there is not optional foreign key relation between the `my_module_protocol` report element and the related `protocol`).

PR #1764, opened 2019-05-14
Add base class for normalizer [SCI-3431]

urbanrotnik:
Jira ticket: [SCI-3431](https://biosistemika.atlassian.net/browse/SCI-3431)

### What was done
Base class added

### DIFF
--- a/app/utilities/protocol_importers/protocol_normalizer.rb
+++ b/app/utilities/protocol_importers/protocol_normalizer.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+module ProtocolImporters
+  class ProtocolNormalizer
+    def load_all_protocols(_params: {})
+      raise NotImplementedError
+    end
+
+    def load_protocol(_id:, _params: {})
+      raise NotImplementedError
+    end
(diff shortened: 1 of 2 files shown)


app/utilities/protocol_importers/protocol_normalizer.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module ProtocolImporters
+  class ProtocolNormalizer
+    def load_all_protocols(_params = {})
+      raise NotImplementedError
+    end
+
+    def load_protocol(_id)
```
jbargu: I wonder if it would be more viable to also pass `params = {}` here of which one parameter can be `id`to account for generality?

urbanrotnik: Hm... `params = {}` or positional arguments like `load_protocol(id:, params: {})`. If we need anything more than ID in the future?

jbargu: Awesome yeah, even better I'm down! :D

PR #1767, opened 2019-05-14
Improve handling of CSRF token errors [SCI-2841]

okriuchykhin:
Jira ticket: [SCI-2841](https://biosistemika.atlassian.net/browse/SCI-2841)

### What was done
Improve handling of CSRF token errors

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -10,6 +10,10 @@ class ApplicationController < ActionController::Base
   around_action :set_time_zone, if: :current_user
   layout 'main'
 
+  rescue_from ActionController::InvalidAuthenticityToken do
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -10,6 +10,10 @@ class ApplicationController < ActionController::Base
   around_action :set_time_zone, if: :current_user
   layout 'main'
 
+  rescue_from ActionController::InvalidAuthenticityToken do
+    redirect_to root_path
+  end
```
radiokills: Should we somehow notify user?

okriuchykhin: I don't think we can display some nice message here and not confuse users

PR #1768, opened 2019-05-15
 Add smart annotation to comments [SCI-3458]

aignatov-bio:
Jira ticket: [SCI-3458](https://biosistemika.atlassian.net/browse/SCI-3458)

### What was done
Create new view element for inline editing to show Smart Annotation objects. Update logic of inline editing.

### DIFF
--- a/app/assets/javascripts/shared/inline_editing.js
+++ b/app/assets/javascripts/shared/inline_editing.js
@@ -35,7 +35,8 @@ function initInlineEditing(title) {
       if (inputString.value === editBlock.dataset.originalName) {
         inputString.disabled = true;
         editBlock.dataset.editMode = 0;
-        $inputString.addClass('hidden').prev().removeClass('hidden');
+        $inputString.addClass('hidden');
+        $editBlock.find('.view-mode').removeClass('hidden');
         return false;
       }
       params[editBlock.dataset.paramsGroup] = {};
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/shared/inline_editing.js
```diff
@@ -63,21 +74,26 @@ function initInlineEditing(title) {
...
     }
 
-    $editBlock.click(e => {
+    $editBlock.click((e) => {
+      if (e.target.tagName === 'A') return true;
       if (inputString.disabled) {
         saveAllEditFields();
         editBlock.dataset.editMode = 1;
         inputString.disabled = false;
+        $inputString.removeClass('hidden').prev().addClass('hidden');
```
jbargu: All this `prev` without selector is calling for some troubles down the road (not only here). Would it be possible to add selector inside so it's more readable for someone what this is doing, or maybe also add a comment

app/assets/javascripts/shared/inline_editing.js
```diff
@@ -63,21 +74,26 @@ function initInlineEditing(title) {
       return true;
     }
 
-    $editBlock.click(e => {
+    $editBlock.click((e) => {
+      if (e.target.tagName === 'A') return true;
```
jbargu: Please comment this, or it's hocus pocus. Which element is 'A'?

app/assets/javascripts/shared/inline_editing.js
```diff
@@ -42,9 +45,17 @@ function initInlineEditing(title) {
         type: 'PUT',
         dataType: 'json',
         data: params,
-        success: function() {
+        success: function(result) {
+          var viewData;
+          if (editBlock.dataset.responseField) {
```
jbargu: In what circumstance does this `if` make sense. Basically we just ignore the successful update?

app/assets/javascripts/shared/inline_editing.js
aignatov-bio: Move it to another line with selector

app/assets/javascripts/shared/inline_editing.js
aignatov-bio: 'A' it is tag name of <a></a> element

app/assets/javascripts/shared/inline_editing.js
aignatov-bio: By default we take string from input element, but if we want to modify this string on backend, we need to provide this data field.
In our case case in input string we must save raw text, but in preview show with html tags.

PR #1775, opened 2019-05-16
JSON structure and basic normalizer for ProtocolsIO [SCI-3432]

urbanrotnik:
Jira ticket: [SCI-3432](https://biosistemika.atlassian.net/browse/SCI-3432)

### What was done
Add basic normalizer for data from ProtocolsIO.

#### ToDo:
Extend to support features like tables, files, build descriptions and checklists. Some questions regarding this included in doc file.


app/utilities/protocol_importers/protocols_io/v3/protocol_normalizer.rb
```diff
@@ -0,0 +1,51 @@
...
+          normalized_data = {}
+          normalized_data[:source_id] = protocol_hash[:id]
+          normalized_data[:name] = protocol_hash[:title]
+          normalized_data[:description] = santz(protocol_hash[:description])
+          normalized_data[:authors] = protocol_hash[:authors].map { |e| e[:name] }.join(', ')
+
+          normalized_data[:steps] = protocol_hash[:steps].map do |e|
+            {
+              source_id: e[:id],
+              name: e[:components].find { |o| o[:type_id] == 6 }[:source][:title],
```
jbargu: Step component digging will be used a lot. Should probably be refactored into a special function.

Additionally it would be better to use descriptive function names for every component that we should use (e.g. `step_name(e[:components])`. Else with this magic numbering the code will be something to look at in a couple of months ([hint hint](https://github.com/biosistemika/scinote-web/blob/7812f70c2f04624a093a5247e4cf13c332c58cf8/app/views/protocols/import_export/_import_json_protocol_s_desc.html.erb)). You can go even a step further and define a mapping of all magic numbers to respective constants or function, which can then be used:

```
def get_component(id, components):
    if AVAILABLE_COMPONENTS.include?(id):
      components.find { |o| o[:type_id] == id}
    else
      raise ArgumentError
    end
end
       
def step_name(components)
    get_component(6, components)[:source][:title]
end
```

A utility module for handling components would be awesome. Saying this here, since it's going to get messier and messier with every step we take. (pun almost intended)

app/utilities/protocol_importers/protocols_io/v3/protocol_normalizer.rb
```diff
@@ -0,0 +1,51 @@
...
+            }
+          end
+
+          # set positions
+          first_step_id = normalized_data[:steps].find { |s| s[:position].zero? }[:source_id]
+          next_step_id = protocol_hash[:steps].find { |s| s[:previous_id] == first_step_id }.try(:[], :id)
+          current_position = 0
+          while next_step_id
+            current_position += 1
+            normalized_data[:steps].find { |s| s[:source_id] == next_step_id }[:position] = current_position
```
jbargu: This is unnecessary O(n^2) time complexity for searching. Create hash above which maps `previous_id: step_hash` and then have O(1) for searching element.

app/utilities/protocol_importers/protocols_io/v3/protocol_normalizer.rb
```diff
@@ -0,0 +1,51 @@
...
+          normalized_data[:source_id] = protocol_hash[:id]
+          normalized_data[:name] = protocol_hash[:title]
+          normalized_data[:description] = santz(protocol_hash[:description])
+          normalized_data[:authors] = protocol_hash[:authors].map { |e| e[:name] }.join(', ')
+
+          normalized_data[:steps] = protocol_hash[:steps].map do |e|
+            {
+              source_id: e[:id],
+              name: e[:components].find { |o| o[:type_id] == 6 }[:source][:title],
+              description: santz(e[:components].find { |o| o[:type_id] == 1 }[:source][:description]),
```
jbargu: I am not sure if we want to strip all tags from the description. It contains HTML and right now I believe it's rendered directly into scinote.

spec/fixtures/files/protocolio/normalized_single_protocol.json
```diff
@@ -0,0 +1,406 @@
+{
```
jbargu: This file and the one below should probably reside in `protocols_io/v3/normalized_single_protocol.json`. Mind the missing `s` and the API number.

spec/utilities/protocol_importers/protocols_io/v3/protocol_normalizer_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  let(:response_without_title) do
+    res_without_title = response
+    res_without_title[:protocol].reject! { |a| a == 'title' }
+    res_without_title
+  end
+
+  let(:normalized_result) do
+    JSON.parse(file_fixture('protocolio/normalized_single_protocol.json').read).to_h.with_indifferent_access
```
jbargu: Take a look at https://github.com/vcr/vcr if it can ease your job.

app/utilities/protocol_importers/protocols_io/v3/protocol_normalizer.rb
```diff
@@ -0,0 +1,51 @@
...
+          normalized_hash = normalize response[:protocol]
+          normalized_hash
+        end
+
+        private
+
+        def normalize(protocol_hash)
+          normalized_data = {}
+          normalized_data[:source_id] = protocol_hash[:id]
+          normalized_data[:name] = protocol_hash[:title]
```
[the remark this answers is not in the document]
urbanrotnik: To current structure, next step/class for sanitization is Intermediate Protocol Object. I would rather see to happen here. What do you think?

app/utilities/protocol_importers/protocols_io/v3/protocol_normalizer.rb
urbanrotnik: Right now it is rendered directly, but for normalized data, we should strip html and have our own styling on FE.

urbanrotnik: We discussed this way with @radiokills.

app/utilities/protocol_importers/protocols_io/v3/protocol_normalizer.rb
```diff
@@ -4,17 +4,28 @@ module ProtocolImporters
   module ProtocolsIO
     module V3
       class ProtocolNormalizer < ProtocolImporters::ProtocolNormalizer
+        AVAILABLE_COMPONENTS = {
```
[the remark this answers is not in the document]
urbanrotnik: Ahh, I miss this 1

PR #1777, opened 2019-05-17
Fix nil handling in different places [SCI-2840]

aignatov-bio:
Jira ticket: [SCI-2840](https://biosistemika.atlassian.net/browse/SCI-2840)

### What was done
```Users::InvitationsController#invite_users (insufficient check for null parameters in #check_invite_users_permission)```
Here i found issue that if `params[:emails]` is `nil` it will fail. So i changed logic for checking `params`.

```Users::InvitationsController#invite_users (poor handling of case when user with such email already exists)```
Here i can't reproduce error, so a little bit optimize logic.

```MyModuleTagsController#create, ParameterMissing: param is missing or the value is empty: my_module_tag```
Added new check for params exist and JS code for error handling.

```my_module_tags_controller.rb #destroy```
Added new check, that tag exist, before destroy.

```
step_comments_controller.rb #load_vars
steps_controller.rb #load_vars
```
Same issue, if step `nil` it will fail.

``` tiny_mce_helper #parse_tiny_mce_asset_to_token```
We don't use it anymore.

``` assets_controller.rb #append_wd_params ```
Better use `.to_query` to converting hash to URL params.

### ToDo
One more issue in Addons

### DIFF
--- a/app/controllers/my_module_tags_controller.rb
+++ b/app/controllers/my_module_tags_controller.rb
@@ -55,10 +55,7 @@ class MyModuleTagsController < ApplicationController
   end
 
   def create
-    unless mt_params[:tag_id]
-      render_403
-      return false
-    end
+    return render_403 unless mt_params[:tag_id]
 
     @mt = MyModuleTag.new(mt_params.merge(my_module: @my_module))
     @mt.created_by = current_user
@@ -89,10 +86,7 @@ class MyModuleTagsController < ApplicationController
   def destroy
     @mt = MyModuleTag.find_by_id(params[:id])
 
-    unless @mt
-      render_404
-      return false
-    end
+    return render_404 unless @mt
 
     Activities::CreateActivityService
       .call(activity_type: :remove_task_tag,
--- a/app/controllers/my_module_tags_controller.rb
+++ b/app/controllers/my_module_tags_controller.rb
@@ -55,11 +55,17 @@ class MyModuleTagsController < ApplicationController
   end
 
   def create
+    unless mt_params[:tag_id]
+      render_403
+      return false
+    end
+
     @mt = MyModuleTag.new(mt_params.merge(my_module: @my_module))
     @mt.created_by = current_user
     @mt.save
 
     my_module = @mt.my_module
+
     Activities::CreateActivityService
       .call(activity_type: :add_task_tag,
             owner: current_user,
@@ -82,6 +88,11 @@ class MyModuleTagsController < ApplicationController
 
   def destroy
     @mt = MyModuleTag.find_by_id(params[:id])
+
+    unless @mt
+      render_404
+      return false
+    end
 
     Activities::CreateActivityService
       .call(activity_type: :remove_task_tag,
--- a/app/controllers/steps_controller.rb
+++ b/app/controllers/steps_controller.rb
@@ -514,7 +514,7 @@ class StepsController < ApplicationController
 
   def load_vars
     @step = Step.find_by_id(params[:id])
-    @protocol = @step.protocol
+    @protocol = @step&.protocol
     if params[:checklistitem_id]
       @chk_item = ChecklistItem.find_by_id(params[:checklistitem_id])
     end
--- a/app/controllers/users/invitations_controller.rb
+++ b/app/controllers/users/invitations_controller.rb
@@ -69,15 +69,11 @@ module Users
         end
 
         result = { email: email }
-
-        if !(Constants::BASIC_EMAIL_REGEX === email) ||
-           validate_user(email, email, generate_user_password).count.positive?
-
+        unless Constants::BASIC_EMAIL_REGEX === email
           result[:status] = :user_invalid
           @invite_results << result
           next
         end
-
         # Check if user already exists
         user = User.find_by_email(email)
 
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

File: app/controllers/my_module_tags_controller.rb [hunk 1]
```diff
@@ -55,11 +55,17 @@ def canvas_index
   end
 
   def create
+    unless mt_params[:tag_id]
+      render_403
+      return false
```
okriuchykhin: you can simply 'return render_403'

okriuchykhin: and can do it in one line

app/controllers/my_module_tags_controller.rb
```diff
@@ -83,6 +89,11 @@ def create
   def destroy
     @mt = MyModuleTag.find_by_id(params[:id])
 
+    unless @mt
+      render_404
+      return false
+    end
+
```
okriuchykhin: same here

app/controllers/steps_controller.rb
```diff
@@ -514,7 +514,7 @@ def extract_destroy_params(params, update_params)
 
   def load_vars
     @step = Step.find_by_id(params[:id])
-    @protocol = @step.protocol
+    @protocol = @step&.protocol
```
okriuchykhin: what will happen in case when @protocol is nil?

app/controllers/users/invitations_controller.rb
```diff
@@ -59,58 +62,45 @@ def invite_users
       @invite_results = []
       @too_many_emails = false
 
-      cntr = 0
-      @emails.each do |email|
-        cntr += 1
-
-        if cntr > Constants::INVITE_USERS_LIMIT
+      @emails.each_with_index do |email, count|
```
okriuchykhin: 'count' is a bit misleading name in this method

app/controllers/users/invitations_controller.rb
```diff
@@ -59,58 +62,45 @@ def invite_users
...
         end
 
-        password = generate_user_password
-
-        # Check if user already exists
-        user = nil
-        user = User.find_by_email(email) if User.exists?(email: email)
-
         result = { email: email }
+        unless Constants::BASIC_EMAIL_REGEX === email
```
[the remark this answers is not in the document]
okriuchykhin: maybe =~ is more appropriate here?

app/controllers/users/invitations_controller.rb
```diff
@@ -212,13 +199,14 @@ def check_captcha_for_invite
...
       @team = Team.find_by_id(params['teamId'])
       @role = params['role']
 
-      render_403 if @emails && @emails.empty?
-      render_403 if @team && !can_manage_team_users?(@team)
-      render_403 if @role && !UserTeam.roles.keys.include?(@role)
+      render_403 unless @emails && @team && @role
+      render_403 if @emails.empty?
+      render_403 unless can_manage_team_users?(@team)
+      render_403 unless UserTeam.roles.key?(@role)
```
okriuchykhin: what will happen if 2 or more conditions will match?

app/controllers/users/invitations_controller.rb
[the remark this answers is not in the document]
aignatov-bio: Yes, i will change. I also find that current regexp not correctly check email format. I will add new regexp.

```
[23] pry(main)> URI::MailTo::EMAIL_REGEXP =~ 'test@mail'
=> 0
[24] pry(main)> URI::MailTo::EMAIL_REGEXP =~ 'test@mail.ru'
=> 0
[25] pry(main)> /\A[\w+\-.]+@[a-z\d\-]+(\.[a-z\d\-]+)*\.[a-z]+\z/i =~ 'test@mail'
=> nil
[26] pry(main)> /\A[\w+\-.]+@[a-z\d\-]+(\.[a-z\d\-]+)*\.[a-z]+\z/i =~ 'test@mail.ru'
=> 0
```

app/controllers/my_module_tags_controller.rb
aignatov-bio: fix

app/controllers/my_module_tags_controller.rb
aignatov-bio: fix

app/controllers/steps_controller.rb
aignatov-bio: There is check later:
````
    unless @protocol
      render_404
    end
````

File: app/controllers/users/invitations_controller.rb (hunk 4 above)
Comment (aignatov-bio) [reply app/controllers/users/invitations_controller.rb]: Fix

File: app/controllers/users/invitations_controller.rb (hunk 6 above)
Comment (aignatov-bio) [reply app/controllers/users/invitations_controller.rb]: Looks like it works, added `return` for safety reasons.

File: config/initializers/constants.rb [hunk 7]
```diff
@@ -908,7 +908,7 @@ class Constants
   IMPORT_REPOSITORY_ITEMS_LIMIT = 2000
 
   # Very basic regex to check for validity of emails
-  BASIC_EMAIL_REGEX = URI::MailTo::EMAIL_REGEXP
+  BASIC_EMAIL_REGEX = /\A[\w+\-.]+@[a-z\d\-]+(\.[a-z\d\-]+)*\.[a-z]+\z/i
```
okriuchykhin: Not sure about this regex. Maybe we still need keep support of local domains like user@lan

aignatov-bio: Revert

PR #1780, opened 2019-05-18
Add a startup check for WOPI_ENABLED == true variables

Ducz0r:
Jira ticket: _n/a_

### What was done
If env. variable `WOPI_ENABLED == 'true'`, but other `WOPI` variables (with the exception of `WOPI_BUSINESS_USERS`, which defaults to `false`) are not properly defined, SciNote can start behaving very oddly, failing on all `PUT` and `PATCH` requests (while being quite hard to debug). I've simply added a new `initializer` that checks if these variables are correctly defined on server startup.

### Note:
I believe I've checked our production servers, and they all have all these env. variables defined, so this should not cause any problems upon the release.

### DIFF
--- a/config/initializers/wopi_startup_check.rb
+++ b/config/initializers/wopi_startup_check.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+# Only check if connection works when server starts, and if WOPI is
+# enabled
+if defined? Rails::Server && ENV['WOPI_ENABLED'] == 'true'
+  missing_vars = []
+  %w(
+    WOPI_TEST_ENABLED WOPI_DISCOVERY_URL WOPI_ENDPOINT_URL USER_SUBDOMAIN
+    WOPI_SUBDOMAIN WOPI_USER_HOST
+  ).each do |var_name|
+    missing_vars << var_name if ENV[var_name].blank?
+  end
+
+  unless missing_vars.empty?
+    puts "WARNING: Due to WOPI_ENABLED == 'true', " \
+         "following env. variables MUST also be specified: " \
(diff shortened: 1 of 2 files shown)


config/initializers/wopi_startup_check.rb
```diff
@@ -0,0 +1,21 @@
...
+
+# Only check if connection works when server starts, and if WOPI is
+# enabled
+if defined? Rails::Server && ENV['WOPI_ENABLED'] == 'true'
+  missing_vars = []
+  %w(
+    WOPI_TEST_ENABLED WOPI_DISCOVERY_URL WOPI_ENDPOINT_URL USER_SUBDOMAIN
+    WOPI_SUBDOMAIN WOPI_USER_HOST
+  ).each do |var_name|
+    missing_vars << var_name if ENV[var_name].blank?
```
radiokills: `blank?` will cause you headache. Try checking for .nil? instead.

radiokills: Huh, I guess there could be a situation where value is empty string, so you have to check both.

radiokills: You could spare some lines using the mighty functional approach

` missing_vars = %w(
    WOPI_TEST_ENABLED WOPI_DISCOVERY_URL WOPI_ENDPOINT_URL USER_SUBDOMAIN
    WOPI_SUBDOMAIN WOPI_USER_HOST
  ).select{ |env_var| ENV[env_var].nil? || ENV[env_var].blank? }`

radiokills: Actually nevermind first 2 comments. Seems activesupport adds .blank? option to nil

PR #1781, opened 2019-05-18
Add delay job for handling unknown tinyMCE images

aignatov-bio:
### What was done
This job cover 2 scenarios:

1) User copy image from one tinyMCE field to another. Job check if image not link to object, than we copy original image and rewrite token.

2) User copy image from other website. Job check if image don't have token, than it try copy image to S3 and create new asset.

### DIFF
--- a/app/models/concerns/tiny_mce_images.rb
+++ b/app/models/concerns/tiny_mce_images.rb
@@ -80,15 +80,7 @@ module TinyMceImages
     end
 
     def copy_unknown_tiny_mce_images
-      asset_team_id = case self.class.name
-                      when 'Protocol'
-                        team_id
-                      when 'Step' || 'MyModule'
-                        protocol.team_id
-                      when 'ResultText'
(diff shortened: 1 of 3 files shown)


app/models/concerns/tiny_mce_images.rb
```diff
@@ -79,6 +79,49 @@ def clone_tinymce_assets(target, team)
...
+    def copy_unknown_tiny_mce_images
+      asset_team_id = case self.class.name
+                      when 'Protocol'
+                        team_id
+                      when 'Step' || 'MyModule'
+                        protocol.team_id
+                      when 'ResultText'
+                        result.my_module.protocol.team_id
+                      end
+
```
radiokills: Can this be extracted into separate method? I guess it could come handy in other situations as well.

aignatov-bio: Moved team id searching to Team model

PR #1786, opened 2019-05-20
Add type column to zip_exports for STI [SCI-3368]

okriuchykhin:
Jira ticket: [SCI-3368](https://biosistemika.atlassian.net/browse/SCI-3368)

### What was done
Added type column for proper functioning of STI. It is needed because paperclip uses class name for hash generation, so we need to know correct class name of exported zip for proper url generation.

### DIFF
--- a/app/models/zip_export.rb
+++ b/app/models/zip_export.rb
@@ -20,7 +20,22 @@ require 'csv'
 
 class ZipExport < ApplicationRecord
   belongs_to :user, optional: true
-  has_attached_file :zip_file
+
+  # Override path only for S3
+  if ENV['PAPERCLIP_STORAGE'] == 's3'
+    s3_path =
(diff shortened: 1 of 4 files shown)


app/models/zip_export.rb
```diff
@@ -20,7 +20,22 @@
...
+
+  # Override path only for S3
+  if ENV['PAPERCLIP_STORAGE'] == 's3'
+    s3_path =
+      if ENV['S3_SUBFOLDER']
+        "/#{ENV['S3_SUBFOLDER']}/zip_exports/:attachment/"\
+        ":id_partition/:hash/:style/:filename"
+      else
+        '/zip_exports/:attachment/:id_partition/:hash/:style/:filename'
+      end
```
radiokills: Can you put this code into some method, this just seems a bit messy.

PR #1795, opened 2019-05-22
Add importer base class [SCI-3430]

mlorb:
Jira ticket: [SCI-3430](https://biosistemika.atlassian.net/browse/SCI-3430)

### What was done
_Add importer base class_

### DIFF
--- a/app/utilities/protocol_importers/protocol_importer.rb
+++ b/app/utilities/protocol_importers/protocol_importer.rb
@@ -1,0 +1,13 @@
+# frozen_string_literal: true
+
+module ProtocolImporters
+  class ProtocolImporter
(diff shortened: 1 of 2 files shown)


jbargu:
Ok sorry, but I believe this class is not needed in the grand scale of things. Normalizer is the one that should be extended, Importer/ApiClient would be the specific to every page. I didn't realize this was still in the docs.

urbanrotnik:
> Ok sorry, but I believe this class is not needed in the grand scale of things. Normalizer is the one that should be extended, Importer/ApiClient would be the specific to every page. I didn't realize this was still in the docs.

Probably res... :thinking_face:

PR #1798, opened 2019-05-23
File preview on new attachments layout [SCI-3145]

jbargu:
Jira ticket: [SCI-3145](https://biosistemika.atlassian.net/browse/SCI-3145)

### What was done
Add file preview to `asset.rb`. Libreoffice dependencies are added to Dockerfile, and for heroku existing buildpack is modified.

Preview available on my-test2.

Related to:
* [buildpack PR](https://github.com/biosistemika/heroku-buildpack-mimeinfo-perl/pull/2)
* [AI addons PR](https://gitlab.com/biosistemika/scinote/addons/ai/merge_requests/41)


Dockerfile
```diff
@@ -11,7 +17,8 @@ RUN curl -sL https://deb.nodesource.com/setup_8.x | bash - && \
   nodejs \
   postgresql-client \
   default-jre-headless \
-  unison \
+  unison && \
+  apt-get install -y -t stretch-backports libreoffice=1:6.1.5-3~bpo9+1 \
```
okriuchykhin: will this version be always present in backports?

Dockerfile
```diff
@@ -2,7 +2,13 @@ FROM ruby:2.5.5
 MAINTAINER BioSistemika <info@biosistemika.com>
 
 # additional dependecies
+
+# Add Debian stretch backports repository
+RUN echo 'deb http://http.debian.net/debian stretch-backports main' \
```
okriuchykhin: Maybe it is better to use ´echo "deb http://http.debian.net/debian $(lsb_release -sc)-backports main"´ for more flexibility, in case base image changes

app/models/asset.rb
```diff
@@ -236,6 +279,33 @@ def post_process_file(team = nil)
...
+
+    service = CreateFilePreviewService.call(file_path)
+
+    if service.succeed?
+      asset.preview = File.open(service.output_file_path)
+      asset.save
+
+      # Delete temp file after it is saved
+      File.delete(service.output_file_path)
+    else
```
okriuchykhin: IN case of error, what will happen to temporary file?

app/models/asset.rb
```diff
@@ -20,11 +20,22 @@ class Asset < ApplicationRecord
                       all: '-background "#d2d2d2" -flatten +matte'
                     }
 
+  has_attached_file :preview,
```
okriuchykhin: Is there a specific reason why you moved preview to separate attachment?

jbargu: I'm not happy with it either, but I couldn't make it work with the PaperClip preprocessors (especially because it expects images and we can have double stuff going on in here, files/images) and I was already way over the estimated effort. Additionally I believe this might simplify migration to ActiveStorage when this will become redundant. How would you go about it?

okriuchykhin: We can create custom post processor for paperclip
https://github.com/thoughtbot/paperclip#dynamic-configuration

Or simply set 'processing' flag on asset update, create delayed job for preview generation and save that preview directly to required style (Asset.first.file.path(:small)).

Dockerfile
jbargu: If it's not we should probably migrate to a newer Debian version already, right?

okriuchykhin: No, I mean specific number, what if it will be replaced with newer build of the package?

jbargu: I wanted to make sure specific version is installed, but I'll remove the version and just install the latest version on the backports repo.

Dockerfile
```diff
@@ -2,7 +2,13 @@ FROM ruby:2.5.5
 MAINTAINER BioSistemika <info@biosistemika.com>
 
 # additional dependecies
+
+# Add Debian stretch backports repository
+RUN echo 'deb http://http.debian.net/debian $(lsb_release -sc)-backports main' \
+  | tee /etc/apt/sources.list.d/stretch-backports.list
```
okriuchykhin: You forgot to replace 'stretch' here

Dockerfile
```diff
@@ -11,7 +17,8 @@ RUN curl -sL https://deb.nodesource.com/setup_8.x | bash - && \
   nodejs \
   postgresql-client \
   default-jre-headless \
-  unison \
+  unison && \
+  apt-get install -y -t stretch-backports libreoffice \
```
okriuchykhin: And here

PR #1807, opened 2019-05-29
Fix some css on step panel heading [SCI-3508]

mlorb:
Jira ticket: [SCI-3508](https://biosistemika.atlassian.net/browse/SCI-3508)

### What was done
_Fix width of step number and name and padding for name._

### DIFF
--- a/app/assets/stylesheets/steps.scss
+++ b/app/assets/stylesheets/steps.scss
@@ -70,13 +70,14 @@
       font-size: 24px;
       font-weight: bold;
       margin: 0 10px;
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/steps.scss
```diff
@@ -88,7 +89,6 @@
           display: flex;
           flex-shrink: 1;
           height: 39px;
-          min-width: 100px;
```
aignatov-bio: Why you remove it? Without `min-width' step name will be covered with big author name.

PR #1814, opened 2019-05-30
Import protocol service [SCI-3433]

urbanrotnik:
Jira ticket: [SCI-3433](https://biosistemika.atlassian.net/browse/SCI-3433)

### What was done
Add service


spec/utilities/protocol_importers/protocol_intermediate_object_spec.rb
```diff
@@ -8,7 +8,7 @@
   let(:user) { create :user }
   let(:team) { create :team }
   let(:normalized_result) do
-    JSON.parse(file_fixture('protocols_importer/protocols_io/v3/normalized_single_protocol.json').read)
+    JSON.parse(file_fixture('protocols_importer/normalized_single_protocol.json').read)
```
jbargu: I've just noticed it's `protocolS_importer` here and `protocol_importerS` in the utilities. If you're doing any refactoring in the future, it could be one of the ideaa to change this to be consistent (no need to this right now!).

urbanrotnik: I agree. Done.

PR #1816, opened 2019-05-31
 Fix nil issue in tags_my_module controller [SCI-3525]

aignatov-bio:

### What was done
 Fix nil issue in tags_my_module controlle


okriuchykhin:
@aignatov-bio Have you found the reason why requests with empty parameters come to controller?

PR #1818, opened 2019-06-03
Move project creation to helper for controller tests [SCI-3520]

aignatov-bio:
Jira ticket: [SCI-3520](https://biosistemika.atlassian.net/browse/SCI-3520)

### What was done
Move project creation to helper for controller tests

### DIFF
--- a/spec/controllers/experiments_controller_spec.rb
+++ b/spec/controllers/experiments_controller_spec.rb
@@ -3,20 +3,12 @@
 require 'rails_helper'
 
 describe ExperimentsController, type: :controller do
-  login_user
-
-  let!(:user) { controller.current_user }
-  let!(:team) { create :team, created_by: user, users: [user] }
-  let!(:project) { create :project, team: team }
-  let!(:user_project) do
-    create :user_project, :owner, user: user, project: project
-  end
-  let(:experiment) { create :experiment, project: project }
+  project_generator
 
   describe 'POST create' do
     let(:action) { post :create, params: params, format: :json }
     let(:params) do
-      { project_id: project.id,
+      { project_id: @project[:project].id,
         experiment: { name: 'test experiment A1',
                       description: 'test description one' } }
     end
@@ -40,7 +32,7 @@ describe ExperimentsController, type: :controller do
     context 'when editing experiment' do
       let(:params) do
         {
-          id: experiment.id,
+          id: @project[:experiment].id,
           experiment: { title: 'new_title' }
         }
       end
@@ -65,7 +57,7 @@ describe ExperimentsController, type: :controller do
                archived: true,
                archived_by: (create :user),
                archived_on: Time.now,
-               project: project
+               project: @project[:project]
       end
 
       let(:params) do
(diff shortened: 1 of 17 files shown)


Ducz0r:
@urbanrotnik @aignatov-bio Thank you for this PR, both for your refactoring work on this, and for your comments.

I'm - sadly - far from expert on RSpec, but to summarize my opinion:

**(1)**
First of all, I think the helper method should have a more descriptive name, maybe `prepare_basic_project`, `generate_basic_project`, `mock_basic_project`, `let_basic_project`, maybe even `let_project`? Something along these lines.

**(2)**
I believe @urbanrotnik has a point. We should be following good RSpec practices (see [this SO topic](https://stackoverflow.com/questions/5359558/when-to-use-rspec-let/5359979#5359979)), and the mentioned performance hit is quite significant if that is indeed the case. So we should resort to using `let()` calls if we can.

**(3)**
This is a list of arguments that can be provided to the method:
```ruby
[:my_module_comments, :repositories, :repository_rows, :project_comments, :results, :result_assets, :result_texts, :result_comments, :result_tables, :steps, :step_comments, :tags]
```
But I don't see much reuse of these arguments (e.g. individual argument is not called from many different places).

---

Based on everything above, my suggestion would be to implement this the following way:
* Implement a helper function (called one of the names above, e.g. `prepare_basic_project(user, team)`). This helper method would prepare the basic **project - experiment - task - protocol** objects hierarchy. I'd prefer that the `user` and `team` objects are created either directly in tests (as right now) or there is an extra method (e.g. `prepare_basic_user_and_team`, or something like that), and that user and team are only passed into the function as an argument.
* This function would - if possible - be using the `let()` statements.
* I would keep the rest of the underlying objects (`:my_module_comments, :repositories, :repository_rows, :project_comments, :results, :result_assets, :result_texts, :result_comments, :result_tables, :steps, :step_comments, :tags`) be initialized the way they currently are, within the test itself - to help with clarity, and to avoid too much magic, as Urban is suggesting. They are mostly one or few-liners, and are not really reused all that much across different tests.

Ergo, a refactored piece of code could look something like this:
```ruby
describe MyModuleCommentsController, type: :controller do
  login_user

  let(:user) { subject.current_user }
  let(:team) { create :team, created_by: user }
  # Alternatively, call function e.g. prepare_basic_user_and_team

  prepare_basic_project(user, team)
  let(:task_comment) { create :task_comment, user: user, my_module: my_module }

  ...
```

The only problem I have is that there is still magic when it comes to naming mocked objects; e.g. when you will call `prepare_basic_project(user, team)`, you will silently be creating mock functions called `project, experiment, my_module, protocol` behind-the-curtains; it would be nice if these function names could be something passed as an argument; but that would then complicate the function invocation even more, and I believe it's over-complicating it at this point.

What do you two think?

aignatov-bio:
Thank you for comments.
I am also not big expert in Rspec, but what i found, `let` has quite limited usage. It is not so flexible if comparing with `before(:each)'. One way, how we can do it, use factories traits. But we will have issue, that in `let` variable we will have, for example, only project. And get other objects will be not so easy.
@Ducz0r maybe we can wait @urbanrotnik and will make some groomig?

Ducz0r:
That sounds like a good plan :+1:. Maybe you two can groom this early next week, once @urbanrotnik is back.

PR #1822, opened 2019-06-05
 Add tests for edit WOPI activities [SCI-3238]

aignatov-bio:
Jira ticket: [SCI-3238](https://biosistemika.atlassian.net/browse/SCI-3238)

### What was done
 Add tests for edit WOPI activities


app/controllers/wopi_controller.rb
```diff
@@ -2,7 +2,7 @@ class WopiController < ActionController::Base
   include WopiUtil
 
   before_action :load_vars, :authenticate_user_from_token!
-  before_action :verify_proof!
+  before_action :verify_proof! unless Rails.env.test?
```
jbargu: We could use either `skip_before_action` call in tests or
```
allow(controller).to receive(:verify_proof!).and_return(true)
```

both of which are less invasive.

config/routes.rb
```diff
@@ -686,4 +686,10 @@ def draw(routes_name)
     get 'wopi/files/:id', to: 'wopi#file_get_endpoint', as: 'wopi_rest_endpoint'
     post 'wopi/files/:id', to: 'wopi#post_file_endpoint'
   end
+
+  if Rails.env.test?
+    namespace :wopi do
+      post :post_file_endpoint
```
jbargu: What's the reason behind this? Wouldn't

```
post :post_file_endpoint, subdomain: ENV['WOPI_SUBDOMAIN']
```
work in the tests? or just setting `ENV['WOPI_SUBDOMAIN']` to nil?

app/controllers/wopi_controller.rb
aignatov-bio: Thank you, it is working.

config/routes.rb
aignatov-bio: Fix

PR #1826, opened 2019-06-11
Generate report in .docx format [SCI-3537]

aignatov-bio:
Jira ticket: [SCI-3537](https://biosistemika.atlassian.net/browse/SCI-3537)

### What was done
Added support for reports in DOCX format

### Addons
Electronic signatures - https://gitlab.com/biosistemika/scinote/addons/electronic_signatures/merge_requests/44
EC Europa - https://gitlab.com/biosistemika/scinote/addons/ec_europa/merge_requests/34


app/services/report/docx_action/experiment.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+# rubocop:disable  Style/ClassAndModuleChildren
+module Report::DocxAction::Experiment
```
Ducz0r: Do we have any good practice about how to name services? I'd say it's probably better to use e.g. `app/services/reports/generate_docx/...` or something like that.

Ducz0r: Do we have any good practice about how to name services? I'd say it's probably better to use e.g. `app/services/reports/generate_docx/...` or something like that.

aignatov-bio: In this approach we receive very complex names with duplicates naming. 
My structure expect, that we have main folder for reports. One main class for each type of report and folder with additional modules and classes. I will optimize it little bit.

Ducz0r: :+1: In any case, other folders use plural, so at least it should be `services/reports/...`

Ducz0r:
The TravisCI build is failing because Travis didn't update its branches to `master` after this PR has been re-targeted to `master` from `rc/16.5`.

PR #1828, opened 2019-06-11
Import protocol service with controller params [SCI-3545]

urbanrotnik:
Jira ticket: [SCI-3545](https://biosistemika.atlassian.net/browse/SCI-3545)

### What was done
Create and save protocol from controller params.

### DIFF
--- a/app/services/protocol_importers/import_protocol_service.rb
+++ b/app/services/protocol_importers/import_protocol_service.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+
+module ProtocolImporters
+  class ImportProtocolService
+    extend Service
+
+    attr_reader :errors, :protocol
+
+    def initialize(protocol_params:, steps_params:, team_id:, user_id:)
+      @user = User.find user_id
+      @team = Team.find team_id
+      @protocol_params = protocol_params
+      @steps_params = steps_params
+      @errors = {}
+    end
+
+    def call
+      return self unless valid?
+
+      @protocol = Protocol.new(@protocol_params.merge!(added_by: @user, team: @team))
+
+      @protocol.steps << @steps_params.collect do |step_params|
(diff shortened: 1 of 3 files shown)


spec/services/protocol_importers/import_protocol_service_spec.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe ProtocolImporters::ImportProtocolService do
+  let(:user) { create :user }
+  let(:team) { create :team }
+  let(:protocol_params) { attributes_for :protocol, :in_public_repository }
```
jbargu: Did you mean `in_repository_public` here?

app/services/protocol_importers/import_protocol_service.rb
```diff
@@ -0,0 +1,49 @@
...
+    def call
+      return self unless valid?
+
+      @protocol = Protocol.new(@protocol_params.merge!(added_by: @user, team: @team))
+
+      @protocol.steps << @steps_params.collect do |step_params|
+        Step.new(step_params.merge(user: @user, completed: false))
+      end
+
+      errors[:protocol] = @protocol.errors.messages unless @protocol.save
```
mlorb: Should here be `@errors`?

spec/services/protocol_importers/import_protocol_service_spec.rb
urbanrotnik: It's ok. It is the name of trait in the factory.

app/services/protocol_importers/import_protocol_service.rb
urbanrotnik: Nice 1. You just figure out that also `errors` working because we have set `attr_reader`. But I agree, should use `@` prefix to be consistent

PR #1835, opened 2019-06-13
Convert image to base64 in PDF reports [SCI-3557]

aignatov-bio:
Jira ticket: [SCI-3557](https://biosistemika.atlassian.net/browse/SCI-3557)

### What was done
Convert image to base64 in PDF reports

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -56,7 +56,7 @@ ActiveRecord::Schema.define(version: 20190520135317) do
     t.datetime "updated_at", null: false
     t.string "file_file_name"
     t.string "file_content_type"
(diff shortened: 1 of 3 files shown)


db/schema.rb
```diff
@@ -56,7 +56,7 @@
     t.datetime "updated_at", null: false
     t.string "file_file_name"
     t.string "file_content_type"
-    t.integer "file_file_size"
+    t.bigint "file_file_size"
```
urbanrotnik: Is ok to change schema without migration?

aignatov-bio: Yes

PR #1837, opened 2019-06-13
Add search service and normalizer for list [SCI-3543]

urbanrotnik:
Jira ticket: [SCI-3543](https://biosistemika.atlassian.net/browse/SCI-3543)

### What was done
Service for getting lists of protocols

#### ToDo:
`valid` method of `SearchProtcolsService` is not final. I added basic wrong arguments catching but it's open to Mojca's task about error handling how to organized them. Messages should go to translations if we will pass them to user.

### DIFF
--- a/app/services/protocol_importers/search_protocols_service.rb
+++ b/app/services/protocol_importers/search_protocols_service.rb
@@ -1,0 +1,70 @@
+# frozen_string_literal: true
+
+module ProtocolImporters
+  class SearchProtocolsService
+    extend Service
+
+    attr_reader :errors, :protocols_list
+
+    CONSTANTS = Constants::PROTOCOLS_IO_V3_API
+
+    def initialize(protocol_source:, query_params: {})
+      @protocol_source = protocol_source
+      @query_params = query_params
+      @errors = Hash.new { |h, k| h[k] = {} }
+    end
+
+    def call
+      return self unless valid?
+
+      api_response = api_client.protocol_list(@query_params)
+
+      @protocols_list = normalizer.normalize_list(api_response)
+
+      self
+    end
+
+    def succeed?
+      @errors.none?
+    end
+
+    private
+
+    def valid?
+      # try if key is not empty
+      @errors[:invalid_params][:key] = 'Key cannot be empty' if @query_params[:key].blank?
+
+      # try if page id is ok
+      if @query_params[:page_id] && !@query_params[:page_id].positive?
+        @errors[:invalid_params][:page_id] = 'Page needs to be positive'
+      end
+
+      # try if order_field is ok
+      if CONSTANTS[:available_order_fields]&.exclude? @query_params[:order_field].to_sym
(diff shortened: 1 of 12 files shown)


spec/fixtures/files/protocol_importers/normalized_list.json
```diff
@@ -0,0 +1,94 @@
...
+      "title": "Preparation of Virus DNA from Seawater for Metagenomics",
+      "created_on": 1434495014,
+      "authors": "Matthew Sullivan Lab",
+      "nr_of_steps": 11,
+      "nr_of_views": 457,
+      "uri": "Preparation-of-Virus-DNA-from-Seawater-for-Metagen-c28yhv"
+    },
+    {
+      "id": 14506,
+      "title": "MALE CIRCUMCISION FOR PREVENTION OF HETEROSEXUAL TRANSMISSION OF HIV IN ADULT MALES IN SOWETO, what do indicators and incidence rates show?untitled protocol",
```
jbargu: :face_with_head_bandage:

app/services/protocol_importers/search_protocols_service.rb
```diff
@@ -0,0 +1,70 @@
...
+      # try if key is not empty
+      @errors[:invalid_params][:key] = 'Key cannot be empty' if @query_params[:key].blank?
+
+      # try if page id is ok
+      if @query_params[:page_id] && !@query_params[:page_id].positive?
+        @errors[:invalid_params][:page_id] = 'Page needs to be positive'
+      end
+
+      # try if order_field is ok
+      if CONSTANTS[:available_order_fields]&.exclude? @query_params[:order_field].to_sym
```
jbargu: This breaks generality. It should most likely be addressed in the API client, as well other checks in here. If that's the case, all these checks could go to API client and the `call` part could easily be done in controller since only two lines are needed for this service. Thoughts?

urbanrotnik: I agree with you. Direct assigning constants breaks generality. I am thinking how much will be code 'inflated' when we add more than 1 endpoint. But not much if we handle params in ApiClient probably. And we can do mappings there if necessery.

urbanrotnik: After discussion with @jbargu, we will stay with the service class for now. Error handling will be implemented there, validations check stays there for now and there is a fair possibility to execute this searching inside of background job in the future.

app/utilities/protocol_importers/protocols_io/v3/protocol_normalizer.rb
```diff
@@ -50,6 +50,24 @@ def normalize_protocol(client_data)
 
           { protocol: normalized_data }
         end
+
+        def normalize_list(client_data)
```
mlorb: In the interface `ProtocolImporters::ProtocolNormalizer` this method is called `normalize_all_protocols`. Decide for one name.

PR #1839, opened 2019-06-14
Adding new activities for image editing [SCI-3369]

aignatov-bio:
Jira ticket: [SCI-3369](https://biosistemika.atlassian.net/browse/SCI-3369)

### What was done
Adding new activities for image editing


app/controllers/assets_controller.rb
```diff
@@ -155,6 +155,13 @@ def view
     render layout: false
   end
 
+  def start_edit
+    asset = Asset.find_by_id(params[:id])
```
jbargu: we could just use the `load_vars` in `before_action` as it is used for other actions as well. Then we should additionally check for permissions if `current_user` even has permissions to add activity.

app/controllers/assets_controller.rb
```diff
@@ -307,4 +315,52 @@ def asset_data_type(asset)
 
     'file'
   end
+
+  def create_edit_image_activity(asset, current_user, started_editing)
```
jbargu: This could be refactored either into a Service, or a method in `Asset` model, or some other utility class. We should keep controller lean.

Instead of passing `started_editing` as boolean, we could use symbols, type: :start_editing, type: finish_editing. Much more readable what's going on.

app/controllers/assets_controller.rb
```diff
@@ -155,6 +155,13 @@ def view
     render layout: false
   end
 
+  def start_edit
```
jbargu: This should be a more descriptive name, like `create_start_edit_image_activity`or along those lines, it's too generic here.

app/assets/javascripts/sitewide/file_preview.js
```diff
@@ -448,6 +449,7 @@ var FilePreviewModal = (function() {
             if (!readOnly && data.editable) {
               modal.find('.file-edit-link').css('display', '');
               modal.find('.file-edit-link').off().click(function(ev) {
+                $.post('/files/' + data.id + '/start_edit');
```
jbargu: Shouldn't we add this activity on `image_loaded` event or even better when user started drawing/editing the image to make sure user doesn't just mistakenly click on the edit button?

app/controllers/assets_controller.rb
```diff
@@ -307,4 +315,52 @@ def asset_data_type(asset)
...
+        .call(activity_type: :edit_image_on_result,
+              owner: current_user,
+              subject: asset.result,
+              team: my_module.experiment.project.team,
+              project: my_module.experiment.project,
+              message_items: {
+                result: asset.result.id,
+                asset_name: { id: asset.id, value_for: 'file_file_name' },
+                action: action
+              })
```
jbargu: We can also edit images on `RepositoryCell` model.

app/assets/javascripts/sitewide/file_preview.js
aignatov-bio: I think it will be very complicated. Now if user don't make any changes, than we will not have activity that he edited image. Cause second activity triggered only if he applied changes.

db/schema.rb
```diff
@@ -56,7 +56,7 @@
     t.datetime "updated_at", null: false
     t.string "file_file_name"
     t.string "file_content_type"
-    t.integer "file_file_size"
+    t.bigint "file_file_size"
```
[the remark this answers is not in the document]
aignatov-bio: I think it is migration from gem update. Because it is paperclip fields.

app/controllers/assets_controller.rb
aignatov-bio: It was not defined in task. Like we don't have activity for Office files in Inventory.

db/schema.rb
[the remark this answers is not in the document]
jbargu: Does this mean we have to run something like `rake paperclip:migrate` in deploy or you think it's not necessary?

app/assets/javascripts/sitewide/file_preview.js
jbargu: Yeah I meant to put into 285th line:
```
    imageEditor.on('image_loaded', () => {
      $('.file-save-link').css('display', '');
    });

```
but if you think this does not bring anything to the table, then discard this comment.

aignatov-bio: It is very close actions, for average images less than second.

db/schema.rb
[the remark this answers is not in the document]
aignatov-bio: It will run as normal migrations. I think it already on master.

jbargu:
Looks good but the tests are failing. You should probably rebase master onto this branch rather than merge it.

PR #1853, opened 2019-06-20
Trunc long protocol name on task page [SCI-3581]

aignatov-bio:
Jira ticket: [SCI-3581](https://biosistemika.atlassian.net/browse/SCI-3581)

### What was done
Trunc long protocol name on task page

### DIFF
--- a/app/helpers/protocol_status_helper.rb
+++ b/app/helpers/protocol_status_helper.rb
@@ -7,7 +7,7 @@ module ProtocolStatusHelper
     res << 'data-trigger="focus" data-placement="bottom" title="'
     res << protocol_status_popover_title(parent) +
            '" data-content="' + protocol_status_popover_content(parent) +
-           '">' + protocol_name(parent) + '</a>'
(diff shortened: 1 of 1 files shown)


app/helpers/protocol_status_helper.rb
```diff
@@ -7,7 +7,7 @@ def protocol_status_href(protocol)
     res << 'data-trigger="focus" data-placement="bottom" title="'
     res << protocol_status_popover_title(parent) +
            '" data-content="' + protocol_status_popover_content(parent) +
-           '">' + protocol_name(parent) + '</a>'
+           '">' + protocol_name(parent).truncate(Constants::NAME_TRUNCATION_LENGTH) + '</a>'
     res.html_safe
```
radiokills: Don't you think this is a bit too short now? Also the acceptable size depends on the width of the window/screen.

aignatov-bio: Well, it's default trunc length for most of fields. And if you click on it, you will see full name.

PR #1858, opened 2019-06-21
PDF thumbnail generation is not done by ImageMagick [SCI-3579]

jbargu:
Jira ticket: [SCI-3579](https://biosistemika.atlassian.net/browse/SCI-3579)

### What was done
Libreoffice has a bug where it cannot open a file (freezing even in GUI
mode), therefore we migrate to ImageMagick and ghostscript. On heroku
ghostscript is already included in the pack.


jbargu:
> Please also update Dockerfile.production

Nice catch!

PR #1870, opened 2019-06-28
Prepare styled listing of protocols [SCI-3535]

jbargu:
Jira ticket: [SCI-3535](https://biosistemika.atlassian.net/browse/SCI-3535)

### What was done
Create missing protocol card list UI.

### Note:
There are most certainly some bugs here and there, but hopefully it's
managable.


jbargu:
> Looks awesome! Ship it
> 
> Should we create a ticket for any of those... point_down
> 
>     * Should we avoid ajax request when having less than 3 chars in the search bar?
> 
>     * Let's try to get published_on instead on a created_on date for listview (I will send them an email)
> 
>     * We have to check why design is broken on safari.

Valid points, it wouldn't be bad if we limit the search but this is management decision. Additionally there should be some way to communicate errors to user, there's none at this points (minus the alert with 'Server error').

PR #1876, opened 2019-07-02
Refactor MarvinJS core over active storage [SCI-3624]

aignatov-bio:
Jira ticket: [SCI-3624](https://biosistemika.atlassian.net/browse/SCI-3624)

### What was done
Refactor MarvinJS core over active storage


app/controllers/assets_controller.rb
```diff
@@ -34,15 +33,22 @@ def file_preview
                elsif @assoc.class == RepositoryCell
                  can_manage_repository_rows?(@repository.team)
                end
-
-    if @asset.image?
-      if ['image/jpeg', 'image/pjpeg'].include? @asset.content_type
+    if response_json['type'] == 'image'
+      if ['image/jpeg', 'image/pjpeg'].include? @asset.file.content_type
```
okriuchykhin: Why do we need this double check here? I thing .image? already covers it

app/controllers/marvin_js_assets_controller.rb
```diff
@@ -1,82 +1,89 @@
...
   end
 
-  private
+  def check_edit_permission
+    if @assoc.class == Step
+      render_403 && return unless can_manage_protocol_in_module?(@protocol) ||
+                                  can_manage_protocol_in_repository?(@protocol)
+    elsif @assoc.class == Result || @assoc.class == MyModule
+      render_403 and return unless can_manage_module?(@my_module)
+    end
```
okriuchykhin: There is no default case here

app/controllers/marvin_js_assets_controller.rb
```diff
@@ -1,82 +1,89 @@
...
 
-  def team_sketches
-    result = ''
-    sketches = current_team.marvin_js_assets.where.not(object_type: 'TinyMceAsset')
-    sketches.each do |sketch|
-      result += render_to_string(
-        partial: 'shared/marvinjs_modal_sketch.html.erb',
-           locals: { sketch: sketch }
-      )
+  def check_read_permission
```
okriuchykhin: There is no default case here

app/controllers/marvin_js_assets_controller.rb
```diff
@@ -1,82 +1,89 @@
...
-    render json: { html: result, sketches: sketches.pluck(:id) }
   end
 
-  private
+  def check_edit_permission
+    if @assoc.class == Step
+      render_403 && return unless can_manage_protocol_in_module?(@protocol) ||
+                                  can_manage_protocol_in_repository?(@protocol)
+    elsif @assoc.class == Result || @assoc.class == MyModule
+      render_403 and return unless can_manage_module?(@my_module)
```
okriuchykhin: Use && instead and, as you did in above case. Or you can simply do 'return render_403'

app/controllers/marvin_js_assets_controller.rb
```diff
@@ -1,82 +1,89 @@
...
-      result += render_to_string(
-        partial: 'shared/marvinjs_modal_sketch.html.erb',
-           locals: { sketch: sketch }
-      )
+  def check_read_permission
+    if @assoc.class == Step
+      render_403 && return unless can_read_protocol_in_module?(@protocol) ||
+                                  can_read_protocol_in_repository?(@protocol)
+    elsif @assoc.class == Result || @assoc.class == MyModule
+      render_403 and return unless can_read_experiment?(@my_module.experiment)
```
okriuchykhin: Use && instead and, as you did in above case. Or you can simply do 'return render_403'

app/controllers/tiny_mce_assets_controller.rb
```diff
@@ -23,4 +28,88 @@ def create
...
+    end
+  end
+
+  def check_read_permission
+    if @assoc.class == Step || @assoc.class == Protocol
+      render_403 && return unless can_read_protocol_in_module?(@protocol) ||
+                                  can_read_protocol_in_repository?(@protocol)
+    elsif @assoc.class == Result || @assoc.class == MyModule
+      render_403 and return unless can_read_experiment?(@my_module.experiment)
+    end
```
okriuchykhin: No default case here

app/controllers/tiny_mce_assets_controller.rb
```diff
@@ -23,4 +28,88 @@ def create
...
+  end
+
+  def check_edit_permission
+    if @assoc.class == Step || @assoc.class == Protocol
+      render_403 && return unless can_manage_protocol_in_module?(@protocol) ||
+                                  can_manage_protocol_in_repository?(@protocol)
+    elsif @assoc.class == Result || @assoc.class == MyModule
+      render_403 and return unless can_manage_module?(@my_module)
+    end
```
okriuchykhin: No default case here

app/controllers/tiny_mce_assets_controller.rb
```diff
@@ -23,4 +28,88 @@ def create
...
+          source_type: result[:asset].image.metadata[:asset_type]
+        }
+      }, content_type: 'text/html'
+    else
+      render json: result[:asset].errors, status: :unprocessable_entity
+    end
+  end
+
+  def marvinjs_update
+    asset = MarvinJsService.update_sketch(marvin_params, current_user)
```
okriuchykhin: So it basically allows to attach MJS asset to any object on the system?

app/controllers/tiny_mce_assets_controller.rb
```diff
@@ -1,6 +1,11 @@
 # frozen_string_literal: true
 
 class TinyMceAssetsController < ApplicationController
+  before_action :load_vars, only: %i(marvinjs_show marvinjs_update)
+
+  before_action :check_read_permission, only: %i(marvinjs_show marvinjs_update)
+  before_action :check_edit_permission, only: %i(marvinjs_update)
+
```
okriuchykhin: There are no permission checking for download and create?

app/services/marvin_js_service.rb
```diff
@@ -1,11 +1,90 @@
...
+    end
+
+    def enabled?
+      !ENV['MARVINJS_URL'].nil? || !ENV['MARVINJS_API_KEY'].nil?
+    end
+
+    def create_sketch(params, current_user)
+      file = generate_image(params)
+      if params[:object_type] == 'TinyMceAsset'
+        asset = TinyMceAsset.new(team_id: current_user.current_team.id)
```
okriuchykhin: I think current_team can change, as it is not request scoped here

app/services/marvin_js_service.rb
```diff
@@ -1,11 +1,90 @@
...
+                          last_modified_by: current_user,
+                          team_id: current_user.current_team.id)
+      attach_file(asset.file, file, params)
+      asset.save!
+      connect_asset(asset, params, current_user)
+    end
+
+    def update_sketch(params, current_user)
+      if params[:object_type] == 'TinyMceAsset'
+        asset = current_user.current_team.tiny_mce_assets.find(Base62.decode(params[:id]))
```
okriuchykhin: I think current_team can change, as it is not request scoped here

app/services/marvin_js_service.rb
```diff
@@ -1,11 +1,90 @@
...
+        attachment = asset&.image
+      else
+        asset = current_user.current_team.assets.find(params[:id])
+        attachment = asset&.file
+      end
+      return unless attachment
+
+      file = generate_image(params)
+      attachment.purge_later
```
okriuchykhin: Isn't attachment getting overwritten with new 'attach' call when parent model has only 'has_one_attached' ?

app/controllers/assets_controller.rb
aignatov-bio: Only jpeg has quality parameter

app/controllers/tiny_mce_assets_controller.rb
aignatov-bio: No, MJS connected to Asset or TinyMceAsset only.

app/controllers/tiny_mce_assets_controller.rb
aignatov-bio: For create we can't check permission because here same issue how for TinyMce images. It is not connected to any object.

app/services/marvin_js_service.rb
aignatov-bio: fix

PR #1880, opened 2019-07-02
SCI-3644 Import button is always visible

radiokills:
Jira ticket: [SCI-3644](https://biosistemika.atlassian.net/browse/SCI-3644)

### What was done
The import button is only visible on hover or when protocol card is active. Also the button is hidden when user doesn't have enough permissions.

### DIFF
--- a/app/assets/stylesheets/protocol_management.scss
+++ b/app/assets/stylesheets/protocol_management.scss
@@ -159,7 +159,7 @@
     .protocol-card {
       border-bottom: 1px solid $color-gainsboro;
       margin-right: 20px;
-      padding: 12px 11px 12px 11px;
+      padding: 12px 11px;
 
       &.active {
         border: 2px solid $brand-primary;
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/protocol_management.scss
```diff
@@ -159,17 +159,28 @@
...
       &:hover {
         background-color: rgba(64,161,215,0.1);
+        padding-bottom: 7px;
       }
+
+      .external-import-btn {
+        display: none;
+      }
+
+      &:hover .external-import-btn, &.active .external-import-btn{
```
jbargu: I think this line could be nested within the 176 line to avoid repeating `.external-import-btn`.

app/controllers/external_protocols_controller.rb
```diff
@@ -9,12 +9,12 @@ def index
     service_call = ProtocolImporters::SearchProtocolsService
                    .call(protocol_source: index_params[:protocol_source],
                          query_params: index_params)
-
     if service_call.succeed?
+      show_import_button = current_user.is_normal_user_or_admin_of_team?(current_team)
```
jbargu: Better to use `can_create_protocols_in_repository?(@team)` to be consistent.  See `check_import_permissions` method.

radiokills: groovy, tnx

PR #1910, opened 2019-07-08
Fix styling for StepComponents in step description [SCI-3656]

urbanrotnik:
Jira ticket: [SCI-3656](https://biosistemika.atlassian.net/browse/SCI-3656)

### What was done
Removed printing HTML tags with ruby.

### Note:
Link for reagent moved out of `details` section because we do not print `<a>` tags in details partial. And we are using the link inside of reagent not inside of vendor from now on. This is a relevant field in the last published protocols.

### DIFF
--- a/app/views/protocol_importers/templates/_note.html.erb
+++ b/app/views/protocol_importers/templates/_note.html.erb
@@ -1,4 +1,10 @@
 <p class="step-description-component-note"><b><%= t('protocol_importers.templates.note.title') %></b><br/>
-  <%= "#{t('protocol_importers.templates.note.author')}: #{item[:author]}" %><br/>
-  <%= item[:body] %></p>
+  <% if item[:author].present? %>
+    <%= "#{t('protocol_importers.templates.note.author')}: #{item[:author]}" %><br/>
+  <% end %>
+
+  <% if item[:body].present? %>
+    <%= item[:body] %>
+  <% end %>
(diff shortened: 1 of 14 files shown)


app/views/protocol_importers/templates/_note.html.erb
```diff
@@ -1,4 +1,6 @@
 <p class="step-description-component-note"><b><%= t('protocol_importers.templates.note.title') %></b><br/>
-  <%= "#{t('protocol_importers.templates.note.author')}: #{item[:author]}" %><br/>
+  <% if item[:name] %>
```
jbargu: You probably meant `:author` here.

app/views/protocol_importers/templates/_note.html.erb
```diff
@@ -1,4 +1,6 @@
...
+    <%= "#{t('protocol_importers.templates.note.author')}: #{item[:author]}" %><br/>
+  <% end %>
   <%= item[:body] %></p>
```
jbargu: Should you the same check for `:body`? It appears it can be empty too https://apidoc.protocols.io/#comment-object

urbanrotnik: We do not have `Body: ` label there, so won't print anything if `:body` is not present. But I add a check. Let's be consistent.

PR #1911, opened 2019-07-09
[SCI-3660] Protocols preview (main pane): All the links in the preview are relative

jbargu:
Jira ticket: [SCI-3660](https://biosistemika.atlassian.net/browse/SCI-3660)

Add base tag to iframe <head>.


urbanrotnik:
@jbargu some tests failed

jbargu:
> Hmm.. Looks ok, but still not working for DominkMetz (case from JIRA ticket)
> `<a href="null/researchers/dominik-metz" target="_blank">dominik metz</a>`
> Works without `null` prefix....

I'm not sure if this has to do anything with this fix. `null/researchers/dominik-metz` hyperlink is flawed by their side. We only prepend the `http://www.protocols.io` to whatever they provide. Could this be connected with missing hrefs as well? Looks like they could have a bug here as well.

PR #1912, opened 2019-07-09
Protocols list preview - Show published_on instead of created_on [SCI-3659]

urbanrotnik:
Jira ticket: [SCI-3659](https://biosistemika.atlassian.net/browse/SCI-3659)

### What was done
Use `published_on` value for showing on the that on the list, because we are requesting data with `published_on` order. 

### Note
Sometimes order does not look correct. We assume that `published_on` we are displaying is the date when the protocol was published for the first time. If Protocol was published many times (someone, probably admin of ProtocolsIO has unpublished it), for an order is used the last published_time but in results, we have only "first" published_on date. This is a rare case.


app/views/protocol_importers/_protocol_card.html.erb
```diff
@@ -8,7 +8,10 @@
 
   <div class='row'>
     <div class='row col-md-12 info-line'>
-      <%= l(Time.at(protocol[:created_on]).to_datetime, format: :full_date) %> • <%= protocol[:authors] %>
+      <% if protocol[:published_on] %>
+        <%= l(Time.at(protocol[:published_on]).to_datetime, format: :full_date) %> •
+    <% end %>
```
radiokills: Indentation mismatch. Can .published_on fail for some reason?

urbanrotnik: For now, we are getting time in UNIX timestamp so it shouldn't fail if `if` pass. Indentation updated.

PR #1920, opened 2019-07-10
Switch to pdftoppm PDF preview maker

jbargu:
Switch to pdftoppm PDF preview maker to avoid licensing costs.

[Buildpack PR](https://github.com/biosistemika/heroku-buildpack-mimeinfo-perl/pull/4)

### DIFF
--- a/lib/paperclip_processors/custom_file_preview.rb
+++ b/lib/paperclip_processors/custom_file_preview.rb
@@ -12,17 +12,10 @@ module Paperclip
 
       begin
         if @file.content_type == 'application/pdf'
-          content = Paperclip.run(
+          Paperclip.run(
             pdftoppm_path,
-            "-singlefile -r 72 -png #{@file.path}"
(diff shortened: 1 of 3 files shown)


lib/paperclip_processors/custom_file_preview.rb
```diff
@@ -11,25 +12,29 @@ def make
 
       begin
         if @file.content_type == 'application/pdf'
-          # We use special convert options for PDFs to improve quality and
-          # background, we append [0] to convert only the first page
-          convert(
-            ":source -resize '#{options[:geometry]}' -format #{options[:format]} -flatten -quality 70 :dest",
-            source: File.expand_path(@file.path) + '[0]',
-            dest: File.expand_path(dst.path)
+          content = Paperclip.run(
```
okriuchykhin: If you provide base filename as last argument, pdftoppm will save image directly to that file and there will be no need to load everything to memory

jbargu: Was a long a day, thanks.

PR #1928, opened 2019-07-12
Improve "make run" command

NicolasCARPi:

Using the "-f" flag on "rm" makes it exit silently and with success if the file is not here. This little change prevents the user running the command from seeing: `rm: cannot remove 'tmp/pids/server.pid': No such file or directory`.

### What was done
Use the -f flag on the `rm` command so it exits with success if the file is not present.


radiokills:
@NicolasCARPi Thanks for your contribution.

PR #1932, opened 2019-07-12
Inventory sharing - data-model update, permissions update [SCI-3687]

urbanrotnik:
Jira ticket: [SCI-3687](https://biosistemika.atlassian.net/browse/SCI-3687)

### What was done
Data model updated, permissions checking updated

### DIFF
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class Repository < ApplicationRecord
   include SearchableModel
   include SearchableByNameModel
@@ -14,6 +16,8 @@ class Repository < ApplicationRecord
            inverse_of: :repository, dependent: :destroy
   has_many :report_elements, inverse_of: :repository, dependent: :destroy
   has_many :repository_list_items, inverse_of: :repository, dependent: :destroy
+  has_many :team_repositories, inverse_of: :repository, dependent: :destroy
+  has_many :teams_shared_with, through: :team_repositories, class_name: 'Team'
 
   auto_strip_attributes :name, nullify: false
   validates :name,
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -8,7 +8,7 @@ class Repository < ApplicationRecord
 
   attribute :discarded_by_id, :integer
 
-  belongs_to :team, optional: true
+  belongs_to :team
   belongs_to :created_by, foreign_key: :created_by_id, class_name: 'User'
   has_many :repository_columns, dependent: :destroy
   has_many :repository_rows, dependent: :destroy
--- a/app/models/team_repository.rb
+++ b/app/models/team_repository.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class TeamRepository < ApplicationRecord
+  enum permission_level: Extends::SHARED_INVENTORIES_PERMISSION_LEVELS
+
+  belongs_to :team
+  belongs_to :repository
+
+  validates :permission_level, presence: true
+end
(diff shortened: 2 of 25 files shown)


app/models/repository.rb
```diff
@@ -6,14 +8,16 @@ class Repository < ApplicationRecord
 
   attribute :discarded_by_id, :integer
 
-  belongs_to :team, optional: true
+  belongs_to :team
```
radiokills: What are we going to do with currently shared repos? There will be failed validations there, we need to think about it.

app/models/repository.rb
```diff
@@ -6,14 +8,16 @@ class Repository < ApplicationRecord
...
   belongs_to :created_by, foreign_key: :created_by_id, class_name: 'User'
   has_many :repository_columns, dependent: :destroy
   has_many :repository_rows, dependent: :destroy
   has_many :repository_table_states,
            inverse_of: :repository, dependent: :destroy
   has_many :report_elements, inverse_of: :repository, dependent: :destroy
   has_many :repository_list_items, inverse_of: :repository, dependent: :destroy
+  has_many :team_repositories, inverse_of: :repository, dependent: :destroy
+  has_many :teams_shared_with, through: :team_repositories, class_name: 'Team'
```
radiokills: shared_with_teams maybe?

app/models/repository.rb
urbanrotnik: It's ok

app/models/repository.rb
urbanrotnik: Wil stays `has_many :teams_...` for now...

config/initializers/extends.rb
```diff
@@ -236,4 +236,9 @@ class Extends
     protocol_repository: [80, 103, 89, 87, 79, 90, 91, 88, 85, 86, 84, 81, 82, 83, 101, 112],
     team: [92, 94, 93, 97, 104]
   }.freeze
+
+  SHARED_INVENTORIES_PERMISSION_LEVELS = {
+    read: 0,
+    write: 1
+  }.freeze
```
okriuchykhin: If it is frozen, can we than extend it?

app/models/team_repository.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class TeamRepository < ApplicationRecord
+  enum permission_level: Extends::SHARED_INVENTORIES_PERMISSION_LEVELS
+
+  belongs_to :team
+  belongs_to :repository
+
+  validates :permission_level, presence: true
```
okriuchykhin: Maybe it is worth also create scope validation, to check that there is only one team<=>repository pair

app/permissions/repository.rb
```diff
@@ -0,0 +1,34 @@
...
+    user.is_admin_of_team?(repository.team)
+  end
+
+  # repository: create/import record
+  can :create_repository_rows do |user, repository|
+    if user.teams.include?(repository.team)
+      user.is_normal_user_or_admin_of_team?(repository.team)
+    elsif (write_team_repos = repository
+                                .team_repositories
+                                .where(team_id: user.teams.pluck(:id))
```
okriuchykhin: Better use '.where(team: user.teams)' - it will use subquery instead 2 separate ones like here

app/permissions/repository.rb
```diff
@@ -0,0 +1,34 @@
...
+                                .where(permission_level: :write)).any?
```
okriuchykhin: As there should be only 0 or 1 record per repository-team pair, you can use '.take' to load only one record and later on you don't need to call 'first', which is a bit confusing

PR #1939, opened 2019-07-16
Add activity logging for sharing inventories [SCI-3692]

urbanrotnik:
Jira ticket: [SCI-3692](https://biosistemika.atlassian.net/browse/SCI-3692)

### What was done
Log activity when a user shares/unshare inventory.


app/controllers/team_repositories_controller.rb
```diff
@@ -20,9 +22,15 @@ def create
...
   def destroy
-    team_repository = @repository.team_repositories.find(destory_params[:id])
-    team_repository.destroy
-    render json: { message: 'destroyed!' }, status: :no_content
+    team_repository = @repository.team_repositories.find_by_id(destory_params[:id])
+
+    if team_repository
+      log_activity(:unshare_inventory, team_repository)
+      team_repository.destroy
+      render json: { message: 'destroyed!' }, status: :no_content
```
jbargu: 204 response (`:no_content`) by the definition should not contain a message body.

config/initializers/extends.rb
```diff
@@ -222,7 +222,9 @@ class Extends
     edit_image_on_result: 110,
     edit_image_on_step: 111,
     edit_image_on_step_in_repository: 112,
-  }
+    share_inventory: 113,
+    unshare_inventory: 114
```
jbargu: You should also add it to corresponding `ACTIVITY_GROUPS` below, probably `inventories` or `team`?

PR #1940, opened 2019-07-17
Show shared inventories in left navigation and fix read repository… [SCI-3689]

mlorb:
…permission

Jira ticket: [SCI-3689](https://biosistemika.atlassian.net/browse/SCI-3689)

### What was done
_Show shared inventories in left navigation. Also read repository permission was fixed._

### ToDo
_Icons are not final._


app/controllers/repositories_controller.rb
```diff
@@ -303,7 +303,7 @@ def load_vars
   def load_parent_vars
     @team = current_team
     render_404 unless @team
-    @repositories = @team.repositories.order(created_at: :asc)
+    @repositories = (@team.repositories + @team.shared_repositories).uniq.sort_by(&:created_at)
```
urbanrotnik: This could be scope in Repository model

app/permissions/repository.rb
```diff
@@ -1,6 +1,21 @@
 # frozen_string_literal: true
 
 Canaid::Permissions.register_for(Repository) do
+  # repository: read/export
+  can :read_repository do |user, repository|
+    if user.teams.include?(repository.team)
+      user.is_member_of_team?(repository.team)
```
urbanrotnik: Do we need this check if `user.teams.include?(repository.team)` is `true`?

app/permissions/repository.rb
```diff
@@ -1,6 +1,21 @@
...
+    elsif (read_team_repo = repository
+                                .team_repositories
+                                .where(team: user.teams).take)
+      # When has some repository's relations with read permissions for at least one of user's teams.
+
+      user.is_member_of_team?(read_team_repo.team)
```
urbanrotnik: same here?

PR #1959, opened 2019-07-29
Update repository index page [SCI-3736]

urbanrotnik:
Jira ticket: [SCI-3736](https://biosistemika.atlassian.net/browse/SCI-3736)

### What was done
Update repositories screen 

#### ToDo:
Export/import buttons weren't moved, commented in JIRA

### DIFF
--- a/app/assets/stylesheets/extend/bootstrap.scss
+++ b/app/assets/stylesheets/extend/bootstrap.scss
@@ -29,12 +29,12 @@
 @media (min-width: 1075px) {
   .col-sm-3-custom {
     float: left;
-    width: 50%;
+    width: 25%;
   }
 
   .col-sm-9-custom {
     float: left;
-    width: 50%;
+    width: 75%;
   }
 }
 
--- a/app/assets/stylesheets/extend/bootstrap.scss
+++ b/app/assets/stylesheets/extend/bootstrap.scss
@@ -29,12 +29,12 @@
 @media (min-width: 1075px) {
   .col-sm-3-custom {
     float: left;
-    width: 25%;
+    width: 50%;
   }
 
   .col-sm-9-custom {
     float: left;
-    width: 75%;
+    width: 50%;
   }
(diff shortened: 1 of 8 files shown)


app/assets/stylesheets/extend/bootstrap.scss
```diff
@@ -29,12 +29,12 @@
 @media (min-width: 1075px) {
   .col-sm-3-custom {
     float: left;
-    width: 25%;
+    width: 50%;
```
aignatov-bio: Don't break bootstrap logic sm-3 means 25% and etc.
Use sm-6 element for 50%

app/assets/stylesheets/extend/bootstrap.scss
```diff
@@ -29,12 +29,12 @@
...
   }
 
   .col-sm-9-custom {
     float: left;
-    width: 75%;
+    width: 50%;
```
aignatov-bio: Same here

app/views/repositories/show.html.erb
```diff
@@ -11,16 +11,34 @@
 
 <div class="content-pane">
   <div id="repository-toolbar">
+    <div class="repository-title">
+      <span class="repository-label"><%= t('repositories.index.title_label') %> </span>
+      <span class="repository-name" title="<%= @repository.name %>"><%= @repository.name %></span>
+      <span class="repository-share-icon">
+        <% if @repository.shared_with?(current_team) %>
+          <% if @repository.team_repositories.first[:permission_level] == 'write' %>
```
jbargu: Wouldn't this check also be true for second shared team with `permission_level = read`? 

Additionally could we write this as:
```
   <% if @repository.team_repositories.first.permission_level.write? %>
```

Just a suggestion.

urbanrotnik: Nice 1, thanks!

PR #1965, opened 2019-08-06
Share inventory popup [SCI-3737]

urbanrotnik:
Jira ticket: [SCI-3737](https://biosistemika.atlassian.net/browse/SCI-3737)

### What was done
Added BE logic for multiple updates (share, unshare, change/permission)
Added modal and JS for doing requests

#### ToDo:
When we get specs, we need to polish look, error messages? 

### Note:
Helper private methods in the controller `teams_to_share`, `teams_to_unshare`, `teams_to_update` are not tested. I didn't find the proper way. Any suggestions?

### DIFF
--- a/app/controllers/team_repositories_controller.rb
+++ b/app/controllers/team_repositories_controller.rb
@@ -33,6 +33,21 @@ class TeamRepositoriesController < ApplicationController
     end
   end
 
+  # POST :team_id/repositories/:repository_id/multiple_update
+  def multiple_update
+    service_call = Repositories::MultipleShareUpdateService.call(repository_id: @repository.id,
+                                                                 user_id: current_user.id,
+                                                                 team_id: current_team.id,
+                                                                 team_ids_for_share: teams_to_share,
+                                                                 team_ids_for_unshare: teams_to_unshare,
+                                                                 team_ids_for_update: teams_to_update)
+    if service_call.succeed?
+      render json: { warnings: service_call.warnings.join(', ') }, status: :ok
+    else
+      render json: { errors: service_call.errors.map { |_, v| v }.join(', ') }, status: :unprocessable_entity
+    end
+  end
+
   private
 
   def load_vars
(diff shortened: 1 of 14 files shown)


app/controllers/team_repositories_controller.rb
```diff
@@ -33,6 +33,21 @@ def destroy
     end
   end
 
+  # POST :team_id/repositories/:repository_id/multiple_update
+  def multiple_update
+    service_call = Repositories::MultipleShareUpdateService.call(repository_id: @repository.id,
+                                                                 user_id: -1,
+                                                                 team_id: current_team.id,
+                                                                 team_ids_for_share: teams_to_share,
```
jbargu: Honestly I'd put those methods in the service and there they can be tested an here you an just pass `multiple_update_params[x]`.

app/controllers/team_repositories_controller.rb
```diff
@@ -33,6 +33,21 @@ def destroy
     end
   end
 
+  # POST :team_id/repositories/:repository_id/multiple_update
+  def multiple_update
+    service_call = Repositories::MultipleShareUpdateService.call(repository_id: @repository.id,
+                                                                 user_id: -1,
```
jbargu: Why is this -1, wouldn't this resolve into `nil` in service?

app/services/repositories/multiple_share_update_service.rb
```diff
@@ -0,0 +1,115 @@
...
+          @errors[:shares] << I18n.t('repositories.multiple_share_service.unable_to_update',
+                                     repository: @repository.name, team: update[:id])
+        end
+      end
+
+      self
+    end
+
+    def succeed?
+      # Shares are not errors but more like warnings.
```
jbargu: Why not just put them into a special `warnings` variable?

urbanrotnik: Because I treat them as errors in the beginning. I will add extra warnings variable.

app/controllers/team_repositories_controller.rb
urbanrotnik: Wooow 😲... thanks.

PR #1983, opened 2019-08-13
Fix report image issues [SCI-3772]

aignatov-bio:
Jira ticket: [SCI-3772](https://biosistemika.atlassian.net/browse/SCI-3772)

### What was done
Fix report image issues

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -7,6 +7,7 @@ class Asset < ApplicationRecord
   include Encryptor
   include WopiUtil
   include ActiveStorageFileUtil
+  include ImageVariantProcessing
 
   require 'tempfile'
   # Lock duration set to 30 minutes
@@ -446,6 +447,10 @@ class Asset < ApplicationRecord
 
   def editable_image?
     !locked? && %r{^image/#{Regexp.union(Constants::WHITELISTED_IMAGE_TYPES_EDITABLE)}} =~ file.content_type
+  end
+
+  def generate_base64(style)
+    return convert_variant_to_base64(medium_preview) if style == :medium
   end
 
   private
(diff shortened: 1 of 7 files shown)


app/models/asset.rb
```diff
@@ -448,6 +449,10 @@ def editable_image?
     !locked? && %r{^image/#{Regexp.union(Constants::WHITELISTED_IMAGE_TYPES_EDITABLE)}} =~ file.content_type
   end
 
+  def generate_base64(style)
```
radiokills: Why are we even passing style here?

app/services/reports/docx/draw_step_asset.rb
```diff
@@ -5,7 +5,7 @@ def draw_step_asset(subject)
     asset = Asset.find_by_id(subject['id']['asset_id'])
     return unless asset
 
-    is_image = asset.is_image?
+    is_image = asset.image?
```
radiokills: Do you really need new variable here?

app/models/user.rb
```diff
@@ -573,19 +574,12 @@ def file_name
...
 
-    avatar_uri = if avatar.options[:storage].to_sym == :s3
-                   URI.parse(avatar.url(style)).open.to_a.join
-                 else
-                   File.open(avatar.path(style)).to_a.join
-                 end
-
-    encoded_data = Base64.strict_encode64(avatar_uri)
-    "data:#{avatar_content_type};base64,#{encoded_data}"
+    convert_variant_to_base64(avatar_variant(style))
```
radiokills: Nice! :+1:

app/models/asset.rb
aignatov-bio: Currently we use it only in one place, but in future we can use it same as avatars, with multiple styles.

app/services/reports/docx/draw_step_asset.rb
aignatov-bio: Fixed

PR #2013, opened 2019-08-26
SCI-3806 Add an icon indicating shared status to dropdown

radiokills:
Jira ticket: [SCI-3806](https://biosistemika.atlassian.net/browse/SCI-3806)

### What was done
- Added status icon to inventories dropdown on task
- Added used and then unshared inventories
- Minor refactoring

### DIFF
--- a/app/helpers/inventories_helper.rb
+++ b/app/helpers/inventories_helper.rb
@@ -8,8 +8,14 @@ module InventoriesHelper
       else
         draw_custom_icon('shared-read')
       end
-    elsif inventory.shared_with_anybody?
-      draw_custom_icon('i-shared')
+    else
+      # The icon should be hiden if repo is not shared (we're updating it dinamically)
+      css_classes = ["repository-share-status"]
+      css_classes.push("hidden") unless inventory.i_shared?(current_team)
+
+      content_tag :span, class: css_classes.join(" ") do
+        draw_custom_icon('i-shared')
+      end
     end
   end
 end
--- a/app/helpers/inventories_helper.rb
+++ b/app/helpers/inventories_helper.rb
@@ -3,7 +3,7 @@
 module InventoriesHelper
   def inventory_shared_status_icon(inventory, team)
     if inventory.shared_with?(team)
-      if inventory.shared_with_write?(team)
+      if can_manage_repository_rows?(inventory)
         draw_custom_icon('shared-edit')
       else
         draw_custom_icon('shared-read')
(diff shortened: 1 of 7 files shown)


app/helpers/inventories_helper.rb
```diff
@@ -0,0 +1,15 @@
...
+module InventoriesHelper
+  def inventory_shared_status_icon(inventory, team)
+    if inventory.shared_with?(team)
+      if inventory.shared_with_write?(team)
+        draw_custom_icon('shared-edit')
+      else
+        draw_custom_icon('shared-read')
+      end
+    elsif inventory.shared_with_anybody?
+      draw_custom_icon('i-shared')
```
aignatov-bio: For ability to dynamic change this icon, when we share/unshare inventory. We must draw it everytime, if it is not shared inventory. We need only swap `hidden` class.

```
<span class="repository-share-status <%= "hidden" unless @repository.i_shared?(current_team) %>"  >	
          <%= draw_custom_icon('i-shared') %>	
</span>
```

app/views/my_modules/_repositories_dropdown.html.erb
```diff
@@ -13,6 +13,27 @@
...
+      <% end %>
+          <%= inventory_shared_status_icon(repository, current_team) %>
+    </a>
+  </li>
+<% end %>
+
+
+<% Repository.used_on_task_but_unshared(@my_module, current_team).each do |repository| %>
+  <li>
+    <a class="disabled dropdown-item <%= "active-item" if @repository == repository %> help_tooltips"
```
aignatov-bio: i think we can remove this `active-item` condition, because this repository will never been active.

app/views/my_modules/_repositories_dropdown.html.erb
```diff
@@ -13,6 +13,27 @@
...
+       data-tooltiplink="<%= I18n.t('tooltips.link.protocol.inventories') %>"
+       data-tooltipcontent="<%= I18n.t('tooltips.text.protocol.inventories') %>"
```
aignatov-bio: Maybe we need some new tooltip text and link here ?

app/helpers/inventories_helper.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module InventoriesHelper
+  def inventory_shared_status_icon(inventory, team)
+    if inventory.shared_with?(team)
+      if inventory.shared_with_write?(team)
```
aignatov-bio: Can you change this to `can_manage_repository_rows?(inventory)`, because we also need to check that user not a guest.

PR #2041, opened 2019-09-10
Fix missing icon on shared inventories in global search [SCI-3847]

mlorb:
Jira ticket: [SCI-3847](https://biosistemika.atlassian.net/browse/SCI-3847)

### Note
_These was some old code for rendering shared icons on global search (forgot during refactoring)._


urbanrotnik:
@aignatov-bio or @mlorb Who is gonna fix the test? :) 
Merge when test is fixed :)

PR #2051, opened 2019-09-16
Create task API endpoint [SCI-3834]

urbanrotnik:
Jira ticket: [SCI-3834](https://biosistemika.atlassian.net/browse/SCI-3834)

### What was done
Added `POST /tasks` endpoint

### Note
Please review [api-documentation](https://github.com/biosistemika/scinote-api-v1-docs/pulls) also.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -81,6 +81,12 @@ en:
           attributes:
             team_id:
               same_team: "Inventory can't be shared to the same team as it belongs to"
+        my_module:
+          attributes:
+            x:
+             not_unique: "and Y position has already been taken by another task in the experiment."
+
+
 
   helpers:
     label:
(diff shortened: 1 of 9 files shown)


config/locales/en.yml
```diff
@@ -81,6 +81,12 @@ en:
           attributes:
             team_id:
               same_team: "Inventory can't be shared to the same team as it belongs to"
+        my_module:
+          attributes:
+            x:
+             not_unique: "and Y position has already been taken by another task in the experiment."
```
jbargu: Shouldn't this be `X and Y` or does it append `X` automatically?

urbanrotnik: X (attribute name) is appended automatically.

urbanrotnik:
> I did not run the code, not sure if this uniqueness constraint causes some issues but otherwise merge away.

MyModule could be not valid (has already taken coordinates) inside of `update_canvas` method before we call `updated_positions`. We are saving model with .save(validate: false) at this place. This is the only place I found issue with this validation.

PR #2058, opened 2019-09-17
Create Protocol step API endpoint [SCI-3835]

urbanrotnik:
Jira ticket: [SCI-3835](https://biosistemika.atlassian.net/browse/SCI-3835)

### What was done
Protocol steps (show, index, create) endpoints added. 

Please review [documentation PR](https://github.com/biosistemika/scinote-api-v1-docs/pull/11) also

### DIFF
--- a/app/controllers/api/v1/base_controller.rb
+++ b/app/controllers/api/v1/base_controller.rb
@@ -110,20 +110,26 @@ module Api
 
       def load_project(key = :project_id)
         @project = @team.projects.find(params.require(key))
-        unless can_read_project?(@project)
-          raise PermissionError.new(Project, :read)
-        end
+        raise PermissionError.new(Project, :read) unless can_read_project?(@project)
       end
 
       def load_experiment(key = :experiment_id)
         @experiment = @project.experiments.find(params.require(key))
-        unless can_read_experiment?(@experiment)
-          raise PermissionError.new(Experiment, :read)
-        end
+        raise PermissionError.new(Experiment, :read) unless can_read_experiment?(@experiment)
       end
 
       def load_task(key = :task_id)
         @task = @experiment.my_modules.find(params.require(key))
+      end
+
+      def load_protocol(key = :protocol_id)
+        @protocol = @task.protocols.find(params.require(key))
+        raise PermissionError.new(Protocol, :read) unless can_read_protocol_in_module?(@protocol)
+      end
+
+      def load_step(key = :step_id)
+        @step = @protocol.steps.find(params.require(key))
+        raise PermissionError.new(Protocol, :read) unless can_read_protocol_in_module?(@step.protocol)
       end
     end
   end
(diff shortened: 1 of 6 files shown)


app/controllers/api/v1/base_controller.rb
```diff
@@ -110,21 +110,26 @@ def load_inventory_item(key = :item_id)
...
         @task = @experiment.my_modules.find(params.require(key))
       end
+
+      def load_protocol(key = :protocol_id)
+        @protocol = @task.protocols.find(params.require(key))
+        raise PermissionError.new(Protocol, :read) unless can_read_protocol_in_module?(@protocol)
+      end
+
+      def load_step(key = :step_id)
+        @step = @protocol.steps.find(params.require(key))
```
jbargu: I am thinking we could also add the same permission check:
```
raise PermissionError.new(Protocol, :read) unless can_read_protocol_in_module?(step.protocol)
```

as above. This way we are conservative in our programming and avoid any security problems down the line if someone misuses this method.

app/controllers/api/v1/steps_controller.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class StepsController < BaseController
+      before_action :load_team, :load_project, :load_experiment, :load_task, :load_protocol
```
jbargu: I find it a bit cumbersome that user has to provide all of the above ids where task or protocol id would suffice. But I guess this is how it is.

app/controllers/api/v1/steps_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+      end
+
+      def show
+        render jsonapi: @step, serializer: StepSerializer
+      end
+
+      def create
+        raise PermissionError.new(Protocol, :create) unless can_manage_protocol_in_module?(@protocol)
+
+        step = @protocol.steps.create!(step_params.merge!(completed: false,
```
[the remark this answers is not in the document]
urbanrotnik: We do not have specified requirements. Because the user does not have the ability to create completed step over UI on the web app, I use the same principle here. But maybe I am wrong...?

app/controllers/api/v1/steps_controller.rb
urbanrotnik: Yeah, it's quite annoying, but I just follow the nested schema for resources we have. 🤔

urbanrotnik: Any ideas @jbargu ?

app/controllers/api/v1/steps_controller.rb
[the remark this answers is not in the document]
jbargu: I see, good point.

app/controllers/api/v1/steps_controller.rb
jbargu: Probably best to not reinvent the wheel and just be consistent. This should be a part of a bigger debate based on user feedback.

PR #2060, opened 2019-09-17
Fix active storage migration script

aignatov-bio:
### What was done
Fix active storage migration script

### DIFF
--- a/db/migrate/20190613134100_convert_to_active_storage.rb
+++ b/db/migrate/20190613134100_convert_to_active_storage.rb
@@ -25,6 +25,13 @@ class ConvertToActiveStorage < ActiveRecord::Migration[5.2]
 
     transaction do
       models.each do |model|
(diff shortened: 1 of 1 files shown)


db/migrate/20190613134100_convert_to_active_storage.rb
```diff
@@ -25,6 +25,13 @@ def up
 
     transaction do
       models.each do |model|
+        table_exists = ActiveRecord::Base.connection.execute("SELECT EXISTS (
+          SELECT 1
+          FROM   information_schema.tables
+          WHERE  table_name = '#{model.table_name}'
+        )").as_json[0]['exists']
+        next unless table_exists
+
```
okriuchykhin: next unless ActiveRecord::Base.connection.table_exists?(model.table_name)

aignatov-bio: Thank you, fixed.

PR #2062, opened 2019-09-17
Fix routes for addons installer

aignatov-bio:

### What was done
Fix routes for addons installer


config/routes.rb
```diff
@@ -3,6 +3,8 @@
     skip_controllers :applications, :authorized_applications, :token_info
   end
 
+  # Addons
```
urbanrotnik: Does this comment fix routes?

aignatov-bio: Yes :)

PR #2075, opened 2019-09-19
Attachments API endpoint  [SCI-3836]

urbanrotnik:
Jira ticket: [SCI-3836](https://biosistemika.atlassian.net/browse/SCI-3836)

### What was done
Added attachments API endpoint (index, show, create)

Please check [documentation PR](https://github.com/biosistemika/scinote-api-v1-docs/pull/12) as well

### DIFF
--- a/app/controllers/api/v1/assets_controller.rb
+++ b/app/controllers/api/v1/assets_controller.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class AssetsController < BaseController
+      before_action :load_team, :load_project, :load_experiment, :load_task, :load_protocol, :load_step
+      before_action :load_asset, only: :show
+
+      def index
+        attachments = @step.assets
+                           .page(params.dig(:page, :number))
+                           .per(params.dig(:page, :size))
+
+        render jsonapi: attachments, each_serializer: AssetSerializer
+      end
+
+      def show
+        render jsonapi: @asset, serializer: AssetSerializer
+      end
+
+      def create
+        raise PermissionError.new(Asset, :create) unless can_manage_protocol_in_module?(@protocol)
+
+        asset = @step.assets.create!(asset_params)
+
+        asset.reload
+
+        render jsonapi: asset,
+               serializer: AssetSerializer,
+               status: :created
+      end
+
+      private
+
+      def asset_params
+        raise TypeError unless params.require(:data).require(:type) == 'attachments'
+
+        attr_list = %i(file)
+        params.require(:data).require(:attributes).require(attr_list)
+        params.require(:data).require(:attributes).permit(attr_list)
+      end
+
+      def load_asset
+        @asset = @step.assets.find(params.require(:id))
+      end
+    end
+  end
+end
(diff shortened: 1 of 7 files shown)


urbanrotnik:
> Does file URL persist long enough for a user to download it?

This link is not expiring at all but it redirects to expirable URL.

app/controllers/api/v1/assets_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+                           .per(params.dig(:page, :size))
+
+        render jsonapi: attachments, each_serializer: AssetSerializer
+      end
+
+      def show
+        render jsonapi: @asset, serializer: AssetSerializer
+      end
+
+      def create
```
jbargu: I was able to upload 100MB file through API, which I cannot do on UI.

app/controllers/api/v1/assets_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+        raise PermissionError.new(Asset, :create) unless can_manage_protocol_in_module?(@protocol)
+
+        asset = @step.assets.create!(asset_params)
+
+        asset.reload
```
jbargu: You should also call `asset.post_process_file` to extract text.

spec/requests/api/v1/assets_controller_spec.rb
```diff
@@ -0,0 +1,157 @@
...
+        }
+      end
+
+      it 'renders 400' do
+        action
+
+        expect(response).to have_http_status(400)
+      end
+    end
+  end
```
jbargu: A test that checks asset permission with a wrong user (with no permissions) would be welcomed, since this is quite important.

urbanrotnik: We are checking permission with `can_read_protocol_in_module?(@asset.step.protocol)`
So we check same permission two times with `load_protocol` and `load_step` in the base controller. With the current structure, we won't hit PermissionError on load_asset. 
But asset could be not found, maybe this test makes more sense. 
What do you think? @jbargu

app/controllers/api/v1/assets_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+      def asset_params
+        raise TypeError unless params.require(:data).require(:type) == 'attachments'
+
+        attr_list = %i(file)
+        params.require(:data).require(:attributes).require(attr_list)
+        params.require(:data).require(:attributes).permit(attr_list)
+      end
+
+      def load_asset
+        @asset = @step.assets.find(params.require(:id))
```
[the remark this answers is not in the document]
urbanrotnik: Added, but with this structure (routes) we are double-checking permissions, right?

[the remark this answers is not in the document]
jbargu: It just does not feel right not checking permission here. I would understand if all the loading of all ancestors would be done in the same method. What is the point of having multiple methods then if you have to keep in mind that this method is dependent on the previous method. Imagine writing a documentation for the methods. `You can call this method but before this you have to call load_step and even before load_protocol...` I understand it is not public. I might be wrong here, I just look at the possible problems that could this cause somewhere down line.

spec/requests/api/v1/assets_controller_spec.rb
jbargu: Yeah as long as we can confirm that one user cannot access some other asset she cannot have access to I believe we are ok. In case someone changes something that breaches this assumption.

PR #2085, opened 2019-09-24
Create/Update result(file) API endpoint  [SCI-3837] [SCI-3838]

urbanrotnik:
Jira ticket: [SCI-3837](https://biosistemika.atlassian.net/browse/SCI-3837)
Jira ticket: [SCI-3838](https://biosistemika.atlassian.net/browse/SCI-3838)
### What was done
Added update endpoint and extended create endpoint

Check also docs [here](https://github.com/biosistemika/scinote-api-v1-docs/pull/13)


app/controllers/api/v1/results_controller.rb
```diff
@@ -92,10 +130,15 @@ def result_params
       # Partially implement sideposting draft
       # https://github.com/json-api/json-api/pull/1197
       def result_text_params
-        prms =
-          params[:included]&.select { |el| el[:type] == 'result_texts' }&.first
-        prms.require(:attributes).require(:text)
-        prms[:attributes]
+        prms = params[:included]&.select { |el| el[:type] == 'result_texts' }&.first
```
okriuchykhin: Can you simplify these blocks?

okriuchykhin: And make sure that appropriate exception is frown in case of corrupted attributes

app/controllers/api/v1/results_controller.rb
```diff
@@ -92,10 +130,15 @@ def result_params
...
+        prms&.require(:attributes)&.require(:text)
+        prms&.dig(:attributes)&.permit(:text)
+      end
+
+      def result_file_params
+        prms = params[:included]&.select { |el| el[:type] == 'result_files' }&.first
+        prms&.require(:attributes)&.require(:file)
+        prms&.dig(:attributes)&.permit(:file)
```
okriuchykhin: Same here

app/controllers/api/v1/results_controller.rb
```diff
@@ -80,6 +94,30 @@ def create_text_result
...
+          asset = Asset.create!(result_file_params)
+          ResultAsset.create!(asset: asset, result: @result)
+        end
+      end
+
+      def update_file_result
+        old_checksum, new_checksum = nil
+        Result.transaction do
+          old_checksum = @result.asset.file.blob.checksum
+          @result.asset.file.purge
```
okriuchykhin: There is no need to delete file manually, ActiveStorage will do it automatically with delayed job

app/controllers/api/v1/results_controller.rb
```diff
@@ -80,6 +94,30 @@ def create_text_result
...
+          @result.asset.file.attach(result_file_params[:file])
+          new_checksum = @result.asset.file.blob.checksum
+        end
+        @asset_result_updated = old_checksum != new_checksum
```
okriuchykhin: Why do you need this?

urbanrotnik: For response type (status 200 or 204) in the controller on the line :35

app/controllers/api/v1/results_controller.rb
urbanrotnik: Isn't this already doing what it should? Do you have any ideas about simplifying this?

app/controllers/api/v1/results_controller.rb
urbanrotnik: Added one test when changing/updating only attachment...

PR #2086, opened 2019-09-24
[SCI-3751] Improve the auto-save functionality on RTF fields

jbargu:
Jira ticket: [SCI-3751](https://biosistemika.atlassian.net/browse/SCI-3751)

### What was done
Figure out where autosave plugin saves the draft and the time of the
draft. I used `updated_at` for every model to compare the time since we
do not save description field updates separately.

[This PR](https://github.com/biosistemika/scinote-web/pull/2061) should be checked before that.

### Note:
Not sure if storage has to be per editor (in case of many steps, they
share the same draft), since the use case is that we navigate from the
active editor. If we need separate storage for every editor then we have
to change ids for every TinyMCE field. I expect many problems here. :cry:

### DIFF
--- a/app/assets/javascripts/protocols/steps.js.erb
+++ b/app/assets/javascripts/protocols/steps.js.erb
@@ -605,6 +605,7 @@
           toggleButtons(true);
           SmartAnnotation.preventPropagation('.atwho-user-popover');
 
+          tinyMCE.editors.step_description_textarea.plugins.autosave.removeDraft();
           tinyMCE.editors.step_description_textarea.remove();
 
           //Rerender tables
--- a/app/assets/javascripts/protocols/steps.js.erb
+++ b/app/assets/javascripts/protocols/steps.js.erb
@@ -366,6 +366,7 @@
       });
       toggleButtons(true);
       DragNDropSteps.clearFiles();
+
       tinyMCE.editors.step_description_textarea.remove();
     });
   }
@@ -604,7 +605,6 @@
           toggleButtons(true);
           SmartAnnotation.preventPropagation('.atwho-user-popover');
 
-          tinyMCE.editors.step_description_textarea.plugins.autosave.removeDraft();
           tinyMCE.editors.step_description_textarea.remove();
 
           //Rerender tables
--- a/app/assets/javascripts/sitewide/tiny_mce.js
+++ b/app/assets/javascripts/sitewide/tiny_mce.js
@@ -53,9 +53,9 @@ var TinyMCE = (function() {
       var notificationBar = $('<div class="restore-draft-notification"></div>');
 
       if (lastDraftTime < lastUpdated) {
-        notificationBar.text('Older version of the text below has been saved in the browser. Do you want to restore it?');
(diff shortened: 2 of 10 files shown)


app/assets/javascripts/sitewide/tiny_mce.js
```diff
@@ -226,17 +287,18 @@ var TinyMCE = (function() {
             });
 
             editor.on('Dirty', function() {
-              var editorForm = $(editor.getContainer()).closest('form');
-              editorForm.find('.tinymce-status-badge').addClass('hidden');
-              $(editor.getContainer())
-                .find('.tinymce-save-button').removeClass('hidden');
+              refreshSaveButton(editor);
```
aignatov-bio: I already made it here - https://github.com/biosistemika/scinote-web/pull/2082

app/assets/stylesheets/tiny_mce.scss
```diff
@@ -66,4 +66,32 @@
...
+  }
+}
+
+.restore-draft-notification {
+  background: $state-info-bg !important;
+  height: 25px !important;
+  padding-left: 10px !important;
+  padding-right: 10px !important;
+  padding-top: 5px !important;
+
```
aignatov-bio: Can you make it without `!important`

app/assets/javascripts/sitewide/tiny_mce.js
jbargu: Well this is on `master` branch and yours is on `activestorage_migration`. It should be fixed when merging.

aignatov-bio: Ok

app/assets/javascripts/protocols/steps.js.erb
```diff
@@ -604,7 +605,6 @@
           toggleButtons(true);
           SmartAnnotation.preventPropagation('.atwho-user-popover');
 
-          tinyMCE.editors.step_description_textarea.plugins.autosave.removeDraft();
```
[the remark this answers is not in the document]
jbargu: Because I thought draft should not be removed if something new happens.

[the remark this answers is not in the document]
aignatov-bio: It is hack for Steps, because after we perform save action draft must be destroyed.

app/assets/stylesheets/tiny_mce.scss
jbargu: The problem is that TinyMCE spreadsheet sets style for every descendant within the editor, so the toolbar is also affected and I am not sure how to force it without `!important`.

app/assets/javascripts/sitewide/tiny_mce.js
jbargu: I renamed my function to `makeItDirty` to easen the merge process.

app/assets/stylesheets/tiny_mce.scss
aignatov-bio: Ok, it's very long to write how to make it. I will show when you will be in office next time. At least can you combine padding in one line: `padding: 5px 10px !important;`

PR #2087, opened 2019-09-25
[SCI-3870] Change/fix asset locking mechanism

jbargu:
Close SCI-3870

Jira ticket: [SCI-3870](https://biosistemika.atlassian.net/browse/SCI-3870)

### What was done
When investigating the above ticket I found out that one of the asset is locked in the mentioned protocol. I believe the current asset locking implementation is bugged, especially with the wrong comparing sign in `unlock_expired`. This probably caused the aforementioned asset to be locked indefinitely. Additionally I am not sure if that `delay` is even needed. RFC

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -449,19 +449,18 @@ class Asset < ApplicationRecord
   # locked?, lock_asset and refresh_lock rely on the asset
   # being locked in the database to prevent race conditions
   def locked?
+    unlock_expired
     !lock.nil?
   end
 
   def lock_asset(lock_string)
     self.lock = lock_string
     self.lock_ttl = Time.now.to_i + LOCK_DURATION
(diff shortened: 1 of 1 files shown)


app/models/asset.rb
```diff
@@ -473,7 +472,7 @@ def unlock
 
   def unlock_expired
     with_lock do
-      if !lock_ttl.nil? && lock_ttl >= Time.now.to_i
+      if !lock_ttl.nil? && lock_ttl <= Time.now.to_i
```
okriuchykhin: Probably only '<' is needed here

jbargu: I believe it is ok as it is, lock should be still valid on `LOCK_DURATION`.

okriuchykhin: Exactly, when you use '<=' and if now == lock_ttl, the lock will be removed, but it should be still valid

PR #2089, opened 2019-09-26
Rake task and unique index for canvas postion [SCI-3887]

urbanrotnik:
Jira ticket: [SCI-3887](https://biosistemika.atlassian.net/browse/SCI-3887)

### What was done
- Updated rails validation
- Add rake task
- ~Add partial unique index on MyModule for `x, y, experiment where archived == false`~ In seperated PR [SCI-3923]

[SCI-3923]: https://biosistemika.atlassian.net/browse/SCI-3923

### DIFF
--- a/app/models/my_module.rb
+++ b/app/models/my_module.rb
@@ -17,7 +17,7 @@ class MyModule < ApplicationRecord
   validates :experiment, presence: true
   validates :my_module_group, presence: true,
             if: proc { |mm| !mm.my_module_group_id.nil? }
-  validate :coordinates, if: proc { |mm| !mm.archived? }
+  validate :coordinates_uniqueness_check, if: :active?
 
   belongs_to :created_by,
              foreign_key: 'created_by_id',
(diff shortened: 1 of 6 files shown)


app/models/my_module.rb
```diff
@@ -500,4 +500,12 @@ def uncomplete
   def create_blank_protocol
     protocols << Protocol.new_blank_for_module(self)
   end
+
+  def coordinates
```
okriuchykhin: May be rename it to something like 'coordinates_validator' or 'coordinates_uniqueness_check' ?

app/models/my_module.rb
```diff
@@ -500,4 +500,12 @@ def uncomplete
...
+    my_modules_on_same_position =
+      MyModule.where(experiment_id: experiment_id, x: x, y: y, archived: false).where.not(id: id)
+    if my_modules_on_same_position.any?
```
okriuchykhin: May be just call directly:  experiment.my_modules.active.where(x: x, y: y).where.not(id: id).any?

app/models/my_module.rb
```diff
@@ -17,7 +17,7 @@ class MyModule < ApplicationRecord
   validates :experiment, presence: true
   validates :my_module_group, presence: true,
             if: proc { |mm| !mm.my_module_group_id.nil? }
-  validates_uniqueness_of :x, scope: %i(y experiment_id), message: :not_unique
+  validate :coordinates, if: proc { |mm| !mm.archived? }
```
okriuchykhin: I think you can simply use 'if: :active?' here

lib/tasks/my_modules.rake
```diff
@@ -0,0 +1,40 @@
...
+    query = MyModule.select('COUNT(*) as duplicates', :x, :y, :experiment_id)
+                    .where(archived: false)
+                    .group(:x, :y, :experiment_id)
+                    .having('COUNT(*)>1')
+
+    Rails.logger.info '*********************************************************************************'
+    Rails.logger.info "You have to relocate #{query.sum { |a| a.duplicates - 1 }} tasks"
+    Rails.logger.info '*********************************************************************************'
+
+    query.each do |row|
```
okriuchykhin: better use find_each

PR #2106, opened 2019-09-30
[SCI-3916] Fix and enable cucumber tests

jbargu:
Jira ticket: [SCI-3916](https://biosistemika.atlassian.net/browse/SCI-3916)

### What was done
Using Alex chrome fix and setting `RAILS_ENV=test` (Puma set it to development) cucumber tests work as expected. Previously there was a rather non deterministic behaviour, everytime you ran the tests, the different number of them failed, but they worked separately. Had to do some digging to uncover the problem.

### DIFF
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -58,7 +58,7 @@ ActionController::Base.allow_rescue = false
 begin
   require 'database_cleaner'
   require 'database_cleaner/cucumber'
-  DatabaseCleaner.strategy = :transaction
+  DatabaseCleaner.strategy = :truncation
 rescue NameError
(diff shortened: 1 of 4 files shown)


features/support/env.rb
```diff
@@ -6,6 +6,7 @@
 # instead of editing this one. Cucumber will automatically load all features/**/*.rb
 # files.
 
+ENV['RAILS_ENV'] = 'test'
```
radiokills: how about `||=`

okriuchykhin: it explicitly sets rails env to 'test' instead of default 'development'

PR #2161, opened 2019-10-17
SCI-3994 fix for excel import

radiokills:
Jira ticket: [SCI-3994](https://biosistemika.atlassian.net/browse/SCI-3994)

### What was done
We're downloading file to disk now so it carries correct file extension.


radiokills:
closing this, it includes current master which it shouldn't

PR #2171, opened 2019-10-22
Fix avatars in comments and activities [SCI-3995]

aignatov-bio:
Jira ticket: [SCI-3995](https://biosistemika.atlassian.net/browse/SCI-3995)

### What was done
Fix avatars in comments and activities

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -182,7 +182,7 @@ module ApplicationHelper
            else
              raw("<span class=\"global-avatar-container smart-annotation\"><img src='#{user_avatar_absolute_url(user, :icon_small)}'" \
              "alt='avatar' class='atwho-user-img-popover'" \
-             " ref='#{'missing-img' if missing_avatar(user, :icon_small)}'></span>")
+             " ref='#{'missing-img' unless user.avatar.attached?}'></span>")
            end
 
     html =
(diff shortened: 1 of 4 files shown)


aignatov-bio:
> Why do you try to use base64 encoded images everywhere?

Avatar was before partially base64 in some places and image links in other places. I think it must be more consistent. Maybe we must decide what we want to use for avatars - image or base64.

app/helpers/application_helper.rb
```diff
@@ -200,22 +200,14 @@ def popover_for_user_name(user,
...
-    begin
-      unless missing_avatar(user, style)
-        image = File.open(user.avatar_variant(style))
-        encoded_data = Base64.strict_encode64(image)
-        avatar_base64 = "data:#{user.avatar_content_type};base64,#{encoded_data}"
-        return avatar_base64
-      end
-    rescue StandardError => e
-      Rails.logger.error e.message
+    avatar_link = user.avatar_variant(style)
```
okriuchykhin: Doesn't avatar_variant check if avatar is attached?

aignatov-bio: `avatar_variant` return variant or link to avatar placeholder. And if it return variant we need obtain `service_url`, but for avatar we can use link as is.

PR #2174, opened 2019-10-23
Add cucumber tests for projects [SCI-4016]

aignatov-bio:
Jira ticket: [SCI-4016](https://biosistemika.atlassian.net/browse/SCI-4016)

### What was done
Add cucumber tests for projects


features/support/shared_methods.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+def check_active_team(team)
```
urbanrotnik: just thinking...
Maybe rename this to `expect_active_team(team)`, so we know we have `expectation` here?

PR #2176, opened 2019-10-23
Enable Teams cucumber tests [SCI-4022]

urbanrotnik:
Jira ticket: [SCI-4022](https://biosistemika.atlassian.net/browse/SCI-4022)

### What was done
Enabled tests, fixed scenario, add some configs


features/step_definitions/team_steps.rb
```diff
@@ -32,3 +34,21 @@
...
+  end
+end
+
+Given('team settings page') do
+  visit teams_path
+  page.save_screenshot(full: true)
+end
+
+Then('I click leave team button') do
+  find("a[data-action='leave-user-team']").click
```
aignatov-bio: Better use text identifier and you can use it as general element:
```
Then("I click {string} button") do |button|
  find('.btn', text: button).click
end
```

And then call it 
```
And I click "Leave Team" button
```

features/step_definitions/team_steps.rb
```diff
@@ -32,3 +34,21 @@
...
+    team = Team.find_by(name: row[:team])
+    team ||= FactoryBot.create(:team, name: row[:team])
+    user = FactoryBot.create(:user, row.slice('name', 'email'))
+    FactoryBot.create(:user_team, user: user, team: team, role: row[:role])
+  end
+end
+
+Given('team settings page') do
+  visit teams_path
+  page.save_screenshot(full: true)
```
aignatov-bio: remove it.
Also we can add support step:
```
Then('I make screenshot') do
  page.save_screenshot(full: true)
end
```
And you can easy add it between steps.

aignatov-bio: Also we can also add gem to automatically open this screenshot.
`page.save_and_open_screenshot(full: true)`

features/step_definitions/team_steps.rb
urbanrotnik: The button contains 2 spans with an icon, is not text-only btn.

PR #2178, opened 2019-10-25
Add render views for datatables [SCI-4033]

aignatov-bio:
Jira ticket: [SCI-4033](https://biosistemika.atlassian.net/browse/SCI-4033)

### What was done
_In a sentence or two describe what was done, focus on the tech aspect_


app/assets/javascripts/application.js.erb
```diff
@@ -42,8 +42,9 @@
 //= require select2_customization
 //= require shared/inline_editing
 //= require activestorage
-//= require turbolinks
 //= require marvinjslauncher
+//= require_directory ./repositories/renders
```
okriuchykhin: isn't it working without './'?

and I think 'renderers' is more correct name

app/helpers/repository_datatable_helper.rb
```diff
@@ -71,29 +72,29 @@ def default_table_columns
...
-    if (DisplayCellValue.respond_to? value_type) && defined?(render)
-      render DisplayCellValue.public_send(value_type, cell, team)
-    elsif DisplayCellValue.respond_to? "#{value_type}_text"
-      DisplayCellValue.public_send("#{value_type}_text", cell, team)
-    else
-      custom_auto_link(display_tooltip(cell.value.data, Constants::NAME_MAX_LENGTH),
-                       simple_format: true,
-                       team: team)
-    end
+    cell_type = cell.value_type.underscore
```
okriuchykhin: We can create separate serializers and call them by cell.value_type. Each data type should have implemented serializers, so there is no need to check if method present

app/assets/javascripts/application.js.erb
aignatov-bio: Fixed

app/helpers/repository_datatable_helper.rb
aignatov-bio: I created serializes for exist  column types

PR #2180, opened 2019-10-28
Update js load order

aignatov-bio:
### What was done
Update js load order


urbanrotnik:
@aignatov-bio please fix this PR

PR #2212, opened 2019-11-14
Merge new filters for global activties 

aignatov-bio:
Merge new filters for global activties


db/schema.rb
```diff
@@ -0,0 +1,1108 @@
+# This file is auto-generated from the current state of the database. Instead
```
urbanrotnik: Should not be here  anymore `schema.rb`

db/structure.sql
```diff
@@ -138,7 +138,7 @@ CREATE TABLE public.activities (
     subject_id bigint,
```
urbanrotnik: I think all changes here belongs to `new-column-types` feature

PR #2217, opened 2019-11-18
Remove Paperclip DB columns and fix experiment templates [SCI-3988]

okriuchykhin:
Jira ticket: [SCI-3988](https://biosistemika.atlassian.net/browse/SCI-3988)

### What was done
Remove Paperclip DB columns and fix experiment templates


db/migrate/20191023162335_remove_paperclip_columns.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+class RemovePaperclipColumns < ActiveRecord::Migration[6.0]
+  def up
+    remove_columns :assets,
+                   :file_file_name,
```
radiokills: It seems we have an index for searching on this column, will this delete the index as well?

db/migrate/20151021085335_add_search_query_indexes.rb:
add_gist_index :assets, :file_file_name

okriuchykhin: "Indexes and table constraints involving the column will be automatically dropped as well"

PR #2220, opened 2019-11-20
DateTime column type (backend part) [SCI-4010]

urbanrotnik:
Jira ticket: [SCI-4010](https://biosistemika.atlassian.net/browse/SCI-4010)

### What was done
Added DateTime column model, controller


db/structure.sql
```diff
@@ -1202,6 +1202,7 @@ CREATE TABLE public.repository_columns (
     data_type integer NOT NULL,
     created_at timestamp without time zone,
     updated_at timestamp without time zone,
+    range boolean,
```
okriuchykhin: It does not belong to here

app/models/repository_column.rb
```diff
@@ -15,12 +15,11 @@ class RepositoryColumn < ApplicationRecord
...
   auto_strip_attributes :name, nullify: false
   validates :name,
-            presence: true,
             length: { maximum: Constants::NAME_MAX_LENGTH },
             uniqueness: { scope: :repository_id, case_sensitive: true }
-  validates :created_by, presence: true
-  validates :repository, presence: true
-  validates :data_type, presence: true
+  validates :name, :data_type, :repository, :created_by, presence: true
+  validates :range, inclusion: { in: [true, false] }, if: :repository_date_time_value?
```
okriuchykhin: It does not belong to here

PR #2223, opened 2019-11-22
Sci 4058 disable inventory delition for normal user

radiokills:
Jira ticket: [SCI-4058 ](https://biosistemika.atlassian.net/browse/SCI-4058)

### What was done
This is a part of the solution for KI: to disable regular users to delete inventory items.

The second part of the solution should be checked here:
https://gitlab.com/biosistemika/scinote/addons/organization_management/merge_requests/54/diffs

### DIFF
--- a/app/controllers/repositories_controller.rb
+++ b/app/controllers/repositories_controller.rb
@@ -29,9 +29,9 @@ class RepositoriesController < ApplicationController
   end
 
   def show
-    @display_edit_button = can_create_repository_rows?(current_user, @repository)
-    @display_delete_button = can_delete_repository_rows?(current_user, @repository)
-    @display_duplicate_button = can_create_repository_rows?(current_user, @repository)
+    @display_edit_button = can_create_repository_rows?(@repository)
+    @display_delete_button = can_delete_repository_rows?(@repository)
+    @display_duplicate_button = can_create_repository_rows?(@repository)
   end
 
   def create_modal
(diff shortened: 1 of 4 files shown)


urbanrotnik:
One comment/question in Gitlab

app/controllers/repositories_controller.rb
```diff
@@ -28,7 +28,11 @@ def index
     render 'repositories/index'
   end
 
-  def show; end
+  def show
+    @display_edit_button = can_create_repository_rows?(current_user, @repository)
```
okriuchykhin: I think you can skip current_user here

app/views/repositories/show.html.erb
```diff
@@ -128,21 +128,30 @@
...
+          <span class="hidden-xs-custom"><%= t("repositories.edit_record") %></span>
+        </button>
+      <% end %>
+      
+      <%if @display_delete_button %>
+        <button type="button" class="btn btn-default"
+          id="deleteRepositoryRecordsButton" onclick="onClickDelete()" disabled>
+          <span class="fas fa-trash"></span>
+          <span class="hidden-xs-custom"><%= t'repositories.delete_record' %></span>
+            <%= submit_tag I18n.t('repositories.delete_record'), :class => "hidden
```
okriuchykhin: Just put it to one line of fix indentation

PR #2255, opened 2019-12-10
Refactor RepositoryRowsController [SCI-4127]

urbanrotnik:
Jira ticket: [SCI-4127](https://biosistemika.atlassian.net/browse/SCI-4127)

### What was done
Add Service for Update and Create repository row, cleanup controller

### DIFF
--- a/app/controllers/api/v1/inventory_cells_controller.rb
+++ b/app/controllers/api/v1/inventory_cells_controller.rb
@@ -69,7 +69,7 @@ module Api
         end
         params.require(:data).require(:attributes).require(:column_id)
         params.require(:data).require(:attributes).require(:value)
-        params[:data][:attributes]
+        params[:data][:attributes].permit!.to_h
       end
 
       def update_inventory_cell_params
--- a/app/controllers/api/v1/inventory_cells_controller.rb
+++ b/app/controllers/api/v1/inventory_cells_controller.rb
@@ -69,7 +69,7 @@ module Api
         end
         params.require(:data).require(:attributes).require(:column_id)
         params.require(:data).require(:attributes).require(:value)
-        params[:data][:attributes].permit!.to_h
+        params[:data][:attributes]
       end
 
       def update_inventory_cell_params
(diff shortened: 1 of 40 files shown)


app/models/repository_time_value.rb
```diff
@@ -1,4 +1,17 @@
 # frozen_string_literal: true
 
 class RepositoryTimeValue < RepositoryDateTimeValueBase
+  def data_changed?(new_data)
+    formatted != formatted(new_date: new_data)
```
okriuchykhin: Why do you compare formatted dates?

app/controllers/api/v1/inventory_cells_controller.rb
```diff
@@ -69,7 +69,7 @@ def inventory_cell_params
         end
         params.require(:data).require(:attributes).require(:column_id)
         params.require(:data).require(:attributes).require(:value)
-        params[:data][:attributes]
+        params[:data][:attributes].permit!.to_h
```
okriuchykhin: Try to avoid this

app/services/repository_rows/update_repository_row_service.rb
```diff
@@ -0,0 +1,76 @@
...
+      return self unless valid?
+
+      ActiveRecord::Base.transaction do
+        # Update invetory row's cells
+        params[:repository_cells]&.each do |column_id, value|
+          column = @repository_row.repository.repository_columns.find_by(id: column_id)
+          next unless column
+
+          cell = @repository_row.repository_cells.find_by(repository_column_id: column.id)
+          unless cell
```
okriuchykhin: You can delete value here if submitted value is empty

app/models/repository_time_value.rb
urbanrotnik: Fixed

app/services/repository_rows/create_repository_row_service.rb
```diff
@@ -0,0 +1,61 @@
...
+      @params = params
+      @errors = {}
+      @repository_row = nil
+    end
+
+    def call
+      return self unless valid?
+
+      ActiveRecord::Base.transaction do
+        @repository_row = RepositoryRow.new(params[:repository_row])
```
okriuchykhin: You can simply call @repository.repository_rows.new(params[:repository_row])

PR #2264, opened 2019-12-12
Add helper for repository tables in docx report [SCI-4157]

aignatov-bio:
Jira ticket: [SCI-4157](https://biosistemika.atlassian.net/browse/SCI-4157)

### What was done
Add helper for repository tables in docx report

### DIFF
--- a/app/services/reports/docx/repository_helper.rb
+++ b/app/services/reports/docx/repository_helper.rb
@@ -12,9 +12,14 @@ module Reports::Docx::RepositoryHelper
       row.push(I18n.l(record.created_at, format: :full))
       row.push(escape_input(record.created_by.full_name))
 
+      cell_values = {}
       custom_cells = record.repository_cells
+      custom_cells.each do |cell|
+        cell_values[cell.repository_column_id] = cell.value.formatted
+      end
+
       repository_data[:data].mappings.each do |column_id, _position|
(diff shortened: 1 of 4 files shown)


app/services/reports/docx/repository_helper.rb
```diff
@@ -0,0 +1,26 @@
...
+    repository_data[:data].repository_rows.each do |record|
+      row = []
+      row.push(record.id)
+      row.push(escape_input(record.name))
+      row.push(I18n.l(record.created_at, format: :full))
+      row.push(escape_input(record.created_by.full_name))
+
+      custom_cells = record.repository_cells
+      repository_data[:data].mappings.each do |column_id, _position|
+        value = custom_cells.find_by(repository_column_id: column_id)&.value&.formatted
```
okriuchykhin: So it loads each cell and value separately?

aignatov-bio: I need check for empty cells and place them in correct order. In `mappings` we have correct column order to match with header.

okriuchykhin: It is not right way to do it as it will generate tons of extra SQL requests

aignatov-bio: Changed

PR #2271, opened 2019-12-18
Refactor checklist column data structure [SCI-4168]

aignatov-bio:
Jira ticket: [SCI-4168](https://biosistemika.atlassian.net/browse/SCI-4168)

### What was done
Refactor checklist column data structure


app/models/repository_checklist_value.rb
```diff
@@ -6,34 +6,43 @@ class RepositoryChecklistValue < ApplicationRecord
...
-    JSON.parse(new_data) != repository_checklist_items
+    JSON.parse(new_data) != repository_checklist_items.pluck(:id)
   end
 
   def update_data!(new_data, user)
-    self.repository_checklist_items = JSON.parse(new_data)
+    repository_cell_values_checklist_items.destroy_all
+    repository_cell.repository_column
+                   .repository_checklist_items.where(id: JSON.parse(new_data)).find_each do |item|
+      repository_cell_values_checklist_items.create(repository_checklist_item: item)
```
okriuchykhin: use create! here

app/models/repository_checklist_value.rb
```diff
@@ -6,34 +6,43 @@ class RepositoryChecklistValue < ApplicationRecord
...
   end
 
   def self.new_with_payload(payload, attributes)
-    value = new(attributes)
-    value.repository_checklist_items = JSON.parse(payload)
+    value = create(attributes)
+    value.repository_cell
+         .repository_column
+         .repository_checklist_items.where(id: JSON.parse(payload)).find_each do |item|
+      value.repository_cell_values_checklist_items.create(repository_checklist_item: item)
```
okriuchykhin: I think it is needed to use 'new' instead of 'create' here

app/models/repository_checklist_value.rb
urbanrotnik: Should we cover this with tests?

app/models/repository_checklist_value.rb
urbanrotnik: Same here?

db/migrate/20191217132401_create_repository_cell_values_checklist_items.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
```
okriuchykhin: Put it into single migration file

db/migrate/20191205133522_create_repository_checklist_column.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+class CreateRepositoryChecklistColumn < ActiveRecord::Migration[6.0]
```
okriuchykhin: Better name it CreateRepositoryChecklists

app/services/repository_datatable_service.rb
```diff
@@ -119,15 +121,18 @@ def sort_rows(column_obj, records)
...
+      sorting_column = RepositoryColumn.find_by(id: id)
       return records unless sorting_column
 
       sorting_data_type = sorting_column.data_type.constantize
 
       cells = RepositoryCell.joins(sorting_data_type::SORTABLE_VALUE_INCLUDE)
                             .where('repository_cells.repository_column_id': sorting_column.id)
                             .select("repository_cells.repository_row_id,
-                                    #{sorting_data_type::SORTABLE_COLUMN_NAME} AS value")
+                                              string_agg(
```
okriuchykhin: Should be STRING_AGG

app/services/repository_datatable_service.rb
```diff
@@ -119,15 +121,18 @@ def sort_rows(column_obj, records)
...
+      sorting_column = RepositoryColumn.find_by(id: id)
       return records unless sorting_column
 
       sorting_data_type = sorting_column.data_type.constantize
 
       cells = RepositoryCell.joins(sorting_data_type::SORTABLE_VALUE_INCLUDE)
                             .where('repository_cells.repository_column_id': sorting_column.id)
                             .select("repository_cells.repository_row_id,
-                                    #{sorting_data_type::SORTABLE_COLUMN_NAME} AS value")
+                                              STRING_AGG(
```
okriuchykhin: It makes select query ~ 2 times slower

aignatov-bio: Now we will use string aggregation only for checklists

PR #2285, opened 2020-01-08
Export/Import for new column types [SCI-4153]

urbanrotnik:
Jira ticket: [SCI-4153](https://biosistemika.atlassian.net/browse/SCI-4153)

### What was done
Tested.

### Note
I found problems when `type` is empty in `DateTimeRangeValues` and `DateTimeValues` tables so I add validations for `type` column. 

Everything is exported/imported as a string (except dropdown). 

Should we implement the same logic also for checklists (and statuses?)?


app/models/repository_checklist_value.rb
```diff
@@ -18,6 +18,10 @@ def formatted
     repository_checklist_items.pluck(:data).join(' | ')
   end
 
+  def export_formatted
```
aignatov-bio: you can  make ` | ` as variable like = `def formatted(separator = ' | ')`
and use `alias export_formatted formatted('\n')`

PR #2286, opened 2020-01-09
Fix for numeric item in SQL array [SCI-4161]

urbanrotnik:
Jira ticket: [SCI-4161](https://biosistemika.atlassian.net/browse/SCI-4161)

### What was done
Fixed bug


urbanrotnik:
@okriuchykhin 
Check fixup commit, probably not change speed, but it's bit more clean

PR #2325, opened 2020-01-16
Fix smart annotation position [SCI-4239]

aignatov-bio:
Jira ticket: [SCI-4239](https://biosistemika.atlassian.net/browse/SCI-4239)

### What was done
Fix smart annotation position


mlorb:
Fix the position so that you can see the input text.
image

PR #2328, opened 2020-01-16
SCI-4260 fix the visibility of manage buttons

radiokills:
Jira ticket: [SCI-4260 ](https://biosistemika.atlassian.net/browse/SCI-4260)

What was done:
- expose the initRowSelection
- call it after reorder or hide/show of the columns


okriuchykhin:
Here is more simple solution:

--- a/app/assets/javascripts/repositories/repository_datatable.js
+++ b/app/assets/javascripts/repositories/repository_datatable.js
@@ -594,7 +594,7 @@ var RepositoryDatatable = (function(global) {
       $(this).parent().find('.repository-row-selector').trigger('click');
     });
 
-    TABLE.on('column-reorder', function() {
+    TABLE.on('column-reorder column-visibility.dt', function() {
       initRowSelection();
     });

PR #2355, opened 2020-01-23
SCI-3403 read openssl-verify-mode from env variable

radiokills:
Jira ticket: [SCI-3403](https://biosistemika.atlassian.net/browse/SCI-3403)


config/secrets.yml
```diff
@@ -41,6 +41,7 @@ common: &common
   mailer_domain: <%= ENV["SMTP_DOMAIN"] %>
   mailer_user_name: <%= ENV["SMTP_USERNAME"] %>
   mailer_password: <%= ENV["SMTP_PASSWORD"] %>
+  mailer_openssl_verify_mode: <%= ENV["SMTP_OPENSSL_VERIFY_MODE"] || nil %>
```
okriuchykhin: Probably it is better to put 'peer' here instead of nil

PR #2383, opened 2020-02-03
Extend inventory import to support new data types [SCI-4220]

okriuchykhin:
Jira ticket: [SCI-4220](https://biosistemika.atlassian.net/browse/SCI-4220)

### What was done
Extend inventory import to support new data types


okriuchykhin:
Because it is scoped to a task and not an inventory

PR #2409, opened 2020-02-12
Move notifications to top navbar [SCI-4235]

aignatov-bio:
Jira ticket: [SCI-4235](https://biosistemika.atlassian.net/browse/SCI-4235)

### What was done
Move notifications to top navbar


mlorb:
Can we move the modals a little lower, so they don't collide with flash notification.
image

PR #2428, opened 2020-02-24
Service for task creating and quick start end point [SCI-4375]

aignatov-bio:
Jira ticket: [SCI-4375](https://biosistemika.atlassian.net/browse/SCI-4375)

### What was done
Service for task creating and quick start end point


app/services/create_my_module.rb
```diff
@@ -0,0 +1,119 @@
...
+      experiment_params = {
+        name: @config[:experiment_name],
+        project: @project,
+        created_by: @user,
+        last_modified_by: @user
+      }
+
+      experiment = @project.experiments.new(experiment_params)
+
+      if experiment.save
```
okriuchykhin: Why do you have save in load method?

app/services/create_my_module.rb
```diff
@@ -0,0 +1,119 @@
...
+        name: @config[:project_name],
+        team: @team,
+        visibility: 'hidden',
+        created_by: @user,
+        last_modified_by: @user
+      }
+
+      project = @team.projects.new(project_params)
+
+      if project.save
```
okriuchykhin: Why do you have save in load method?

app/services/create_my_module.rb
```diff
@@ -0,0 +1,119 @@
...
+        y: 0,
+        name: @config[:my_module_name] || I18n.t('create_task_service.default_task_name')
+      }
+
+      my_module = @experiment.my_modules.new(my_module_params)
+      new_pos = my_module.get_new_position
+      my_module.x = new_pos[:x]
+      my_module.y = new_pos[:y]
+
+      if my_module.save
```
okriuchykhin: Just use save! As it is in transaction, in case of failure all changes will be rolled back.

app/services/create_my_module.rb
```diff
@@ -0,0 +1,119 @@
...
+    @config = config
+    @user = user
+    @team = team
+  end
+
+  def call
+    ActiveRecord::Base.transaction do
+      @project = load_project
+      @experiment = load_experiment
+      raise ActiveRecord::Rollback unless @project && @experiment && can_manage_experiment?(@user, @experiment)
```
okriuchykhin: Why do you check permissions in service and not in controller?

app/services/create_my_module.rb
```diff
@@ -0,0 +1,119 @@
+# frozen_string_literal: true
+
+class CreateMyModule
```
okriuchykhin: Add Service to the name

app/services/create_my_module.rb
```diff
@@ -0,0 +1,119 @@
...
+  include Canaid::Helpers::PermissionsHelper
+
+  # config example: {project_id: 5, experiment_name: 'New Experiment', my_module_name: 'New Task'}
+
+  def initialize(user, team, config)
```
okriuchykhin: Why it is named 'config' and not 'params'?

app/services/create_my_module.rb
```diff
@@ -0,0 +1,119 @@
...
+
+  private
+
+  def load_project
+    if @config[:project_id]
+      @team.projects.find_by(id: @config[:project_id])
+    elsif @config[:project_name]
+      return false unless can_create_projects?(@user, @team)
+
+      project_params = {
```
okriuchykhin: I think it shouldn't be here

app/services/create_my_module.rb
```diff
@@ -0,0 +1,119 @@
...
+      end
+    end
+  end
+
+  def load_experiment
+    return false unless @project
+
+    if @config[:experiment_id]
+      @project.experiments.find_by(id: @config[:experiment_id])
+    elsif @config[:experiment_name]
```
okriuchykhin: I think it shouldn't be here

PR #2466, opened 2020-03-12
Fix bug with date format when importing to inventory [SCI-4450]

mlorb:
Jira ticket: [SCI-4450](https://biosistemika.atlassian.net/browse/SCI-4450)

### What was done
Fix bug with date format when importing to inventory.
Tests were added.

### DIFF
--- a/app/models/repository_date_time_value_base.rb
+++ b/app/models/repository_date_time_value_base.rb
@@ -26,7 +26,7 @@ class RepositoryDateTimeValueBase < ApplicationRecord
   end
 
   def self.import_from_text(text, attributes, options = {})
-    date_format = options.dig(:user, :settings, :date_format) || Constants::DEFAULT_DATE_FORMAT
+    date_format = options[:user].settings[:date_format] || Constants::DEFAULT_DATE_FORMAT
     new(attributes.merge(data: DateTime.strptime(text, date_format)))
   rescue ArgumentError
     nil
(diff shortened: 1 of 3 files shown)


app/utilities/repository_import_parser/importer.rb
```diff
@@ -137,7 +137,7 @@ def import_batch_to_database(full_row_import_batch)
           cell_value = column.data_type.constantize.import_from_text(
             value,
             cell_value_attributes,
-            user: @user.as_json
```
okriuchykhin: It should be serialized

app/models/repository_date_time_value_base.rb
```diff
@@ -26,7 +26,7 @@ def update_data!(new_data, user)
   end
 
   def self.import_from_text(text, attributes, options = {})
-    date_format = options.dig(:user, :settings, :date_format) || Constants::DEFAULT_DATE_FORMAT
+    date_format = options[:user].settings[:date_format] || Constants::DEFAULT_DATE_FORMAT
```
okriuchykhin: What if :user is missing?

app/utilities/repository_import_parser/importer.rb
radiokills: Can you plz explain why it should be serialized.

okriuchykhin: In case we use it in delayed jobs, to preserve user settings for that specific point in time

okriuchykhin: @user.as_json(root: true, only: :settings)

mlorb:
I added another [PR](https://gitlab.com/biosistemika/scinote/addons/audit_trails/-/merge_requests/54)

PR #2503, opened 2020-04-02
Add assigned repositories list [SCI-4507]

aignatov-bio:
Jira ticket: [SCI-4507](https://biosistemika.atlassian.net/browse/SCI-4507)

### What was done
Add assigned repositories list


app/models/my_module.rb
```diff
@@ -205,6 +205,15 @@ def repository_rows_count(repository)
                              .count
   end
 
+  def assigned_repositories_list
```
okriuchykhin, line 208: Does it return collection of Repository models?

aignatov-bio: Updated

app/models/my_module.rb
```diff
@@ -205,6 +205,13 @@ def repository_rows_count(repository)
                              .count
   end
 
+  def assigned_repositories
```
okriuchykhin, line 208: it will do 2 separate SQL queries

app/views/my_modules/protocols.html.erb
```diff
@@ -39,7 +39,7 @@
     </a>
     </div>
     <div class="collapse" id="assigned-items-container">
-      ...
+      <%= render partial: "my_modules/inventories/inventories_list" %>
```
mlorb, line 42: In code we mostly use `repositories`. Can you rename partial.

aignatov-bio: updated

app/models/my_module.rb
aignatov-bio: updated

PR #2521, opened 2020-04-24
Fix items count validation [SCI-4513]

urbanrotnik:
Jira ticket: [SCI-4513](https://biosistemika.atlassian.net/browse/SCI-4513)

### What was done
Not much to say... :D


urbanrotnik:
> Can you fix also backend and checklist limit as well?

Nice catch Mojca, it's fixed. Please check again :)

PR #2525, opened 2020-04-28
Implement background processing of repository snapshots [SCI-4552]

okriuchykhin:
Jira ticket: [SCI-4552](https://biosistemika.atlassian.net/browse/SCI-4552)

### What was done
Implement background processing of repository snapshots


app/controllers/my_module_repository_snapshots_controller.rb
```diff
@@ -30,22 +30,38 @@ def index_dt
...
+      html: render_to_string(partial: 'my_modules/repositories/full_view_version',
+                             locals: { repository_snapshot: @repository_snapshot })
+    }
   end
 
   def destroy
     @repository_snapshot.destroy!
-    @repository_snapshots = @my_module.repository_snapshots.where(original_repository: @repository)
-    render json: { html: render_to_string(partial: 'my_modules/repositories/full_view_versions_sidebar') }
+    render json: {}
```
radiokills, line 64: Would at least ID of deleter record come handy at some point?

app/jobs/application_job.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class ApplicationJob < ActiveJob::Base
+  # Automatically retry jobs that encountered a deadlock
+  retry_on ActiveRecord::Deadlocked
```
radiokills, line 5: :+1:

app/models/repository_snapshot.rb
```diff
@@ -1,10 +1,13 @@
 # frozen_string_literal: true
 
 class RepositorySnapshot < RepositoryBase
+  enum status: { provisioning: 0, ready: 1, failed: 2 }
```
radiokills, line 4: I know it's not used anywhere ATM, but it would be nice to have an information that snapshot generation is in progress, I suggest we add a 'started' state. It would get set before the main transaction. It would help with potential debugging.

db/migrate/20200331183640_add_repository_snapshots.rb
```diff
@@ -3,6 +3,7 @@
 class AddRepositorySnapshots < ActiveRecord::Migration[6.0]
   def up
     add_column :repositories, :parent_id, :bigint, null: true
+    add_column :repositories, :status, :integer, null: true
```
radiokills, line 6: Shouldn't this column have a default state? Especially if you are validating presence in the model.

app/controllers/my_module_repository_snapshots_controller.rb
okriuchykhin: Don't think so

app/models/repository_snapshot.rb
okriuchykhin: It is called 'provisioning'

db/migrate/20200331183640_add_repository_snapshots.rb
okriuchykhin: I think it should be explicitly set by business logic when snapshot is creating

okriuchykhin: But we can set it to 'provisioning' by default

okriuchykhin: Wait, this is STI, so better to have it nullable for simple repositories

PR #2541, opened 2020-05-07
Merge assign and unassign service [SCI-4591]

aignatov-bio:
Jira ticket: [SCI-4591](https://biosistemika.atlassian.net/browse/SCI-4591)

### What was done
Merge assign and unassign service


app/services/repository_rows/my_module_update_rows_service.rb
```diff
@@ -1,17 +1,24 @@
 # frozen_string_literal: true
 
 module RepositoryRows
-  class MyModuleAssigningService
+  class MyModuleUpdateRowsService
```
okriuchykhin, line 4: Rows are not getting updated, they are just assigned/unassigned

app/services/repository_rows/my_module_update_rows_service.rb
```diff
@@ -1,17 +1,24 @@
...
+                :assigned_rows_count,
+                :unassigned_rows_count,
+                :errors
 
     def initialize(my_module:, repository:, user:, params:)
       @my_module = my_module
       @repository = repository
       @user = user
       @params = params
-      @assigned_rows_names = Set[]
```
okriuchykhin, line 14: So it does't return item names anymore?

aignatov-bio: Before we used names for activities and flash messages. Now activities inside service. In flash messages we are showing only count of assigned/unassigned rows.

PR #2557, opened 2020-05-13
Make snapshots selectable [SCI-4596]

okriuchykhin:
Jira ticket: [SCI-4596](https://biosistemika.atlassian.net/browse/SCI-4596)

### What was done
Make snapshots selectable. Also fixes handling of deleted repositories.


app/models/my_module.rb
```diff
@@ -201,7 +200,29 @@ def repository_rows_count(repository)
...
+  def live_and_snapshot_repositories_list
+    snapshots = repository_snapshots.left_outer_joins(:original_repository)
+
+    selected_snapshots = snapshots.where(selected: true)
+    selected_snapshots = selected_snapshots.or(snapshots.where(original_repositories_repositories: { id: nil }))
+    selected_snapshots = selected_snapshots.select('DISTINCT ON ("repositories"."parent_id") "repositories".*')
+                                           .select('COUNT(repository_rows.id) AS assigned_rows_count')
+                                           .joins(:repository_rows)
+                                           .group(:parent_id, :id)
+                                           .order(:parent_id, updated_at: :desc)
```
radiokills, line 219: Why not simply

``` ruby	
selected_snapshots = snapshots.where(selected: true)
                        .or(snapshots.where(original_repositories_repositories: { id: nil }))
                        .select('DISTINCT ON ("repositories"."parent_id") "repositories".*')
                        .select('COUNT(repository_rows.id) AS assigned_rows_count')
                        .joins(:repository_rows)
                        .group(:parent_id, :id)
                        .order(:parent_id, updated_at: :desc)
```

radiokills: Just a styling suggestion, otherwise great work.

PR #2560, opened 2020-05-14
SCI-4601 fix the associations, renaming variable

radiokills:


app/models/repository_status_value.rb
```diff
@@ -31,12 +31,12 @@ def update_data!(new_data, user)
 
   def snapshot!(cell_snapshot)
     value_snapshot = dup
-    list_item = cell_snapshot.repository_column
+    status_item = cell_snapshot.repository_column
                              .repository_status_items
```
okriuchykhin, line 35: align it with previous line

PR #2567, opened 2020-05-15
Show snapshot in report [SCI-4622]

urbanrotnik:
Jira ticket: [SCI-4622](https://biosistemika.atlassian.net/browse/SCI-4622)

### What was done
Show snapshot in the report if snapshot exists and it is selected

### DIFF
--- a/app/controllers/my_module_repository_snapshots_controller.rb
+++ b/app/controllers/my_module_repository_snapshots_controller.rb
@@ -6,6 +6,7 @@ class MyModuleRepositorySnapshotsController < ApplicationController
   before_action :load_repository_snapshot, except: %i(create full_view_sidebar select)
   before_action :check_view_permissions, except: %i(create destroy select)
   before_action :check_manage_permissions, only: %i(create destroy select)
+  before_action :update_report_references, only: :destroy
 
   def index_dt
     @draw = params[:draw].to_i
@@ -98,10 +99,20 @@ class MyModuleRepositorySnapshotsController < ApplicationController
       repository_snapshot.update!(selected: true)
     end
 
+    @my_module.update_report_repository_references(params[:repository_id] || repository_snapshot&.id)
+
     render json: {}
   end
 
   private
+
+  def update_report_references
+    repository_or_snap_id = @repository_snapshot.original_repository&.id || @repository_snapshot.id
+    default_view_candidate =
+      @my_module.active_snapshot_or_live(repository_or_snap_id, exclude_snpashot_ids: [@repository_snapshot.id])
+
+    @my_module.update_report_repository_references(default_view_candidate.id) if default_view_candidate
+  end
 
   def load_my_module
     @my_module = MyModule.find_by(id: params[:my_module_id])
--- a/app/controllers/my_module_repository_snapshots_controller.rb
+++ b/app/controllers/my_module_repository_snapshots_controller.rb
@@ -6,7 +6,6 @@ class MyModuleRepositorySnapshotsController < ApplicationController
   before_action :load_repository_snapshot, except: %i(create full_view_sidebar select)
   before_action :check_view_permissions, except: %i(create destroy select)
   before_action :check_manage_permissions, only: %i(create destroy select)
-  before_action :update_report_references, only: :destroy
 
   def index_dt
     @draw = params[:draw].to_i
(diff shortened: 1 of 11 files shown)


app/models/my_module.rb
```diff
@@ -225,6 +225,20 @@ def live_and_snapshot_repositories_list
     (live_repositories + selected_snapshots).sort_by { |r| r.name.downcase }
   end
 
+  def active_snapshot_or_live(repository_id)
+    r = RepositoryBase.find repository_id
```
okriuchykhin, line 229: trying to save few bytes?

app/views/reports/elements/_my_module_repository_element.html.erb
```diff
@@ -1,45 +1,57 @@
-<% repository ||= Repository.find(element_id) %>
 <% if my_module.blank? and @my_module.present? then my_module = @my_module end %>
-<% rows = my_module.repository_rows.where(repository: repository) %>
+<% repository_snapshot ||= nil %>
+<% repository ||= repository_snapshot %>
+<% repository ||= my_module.active_snapshot_or_live(element_id) || Repository.find_by(id: element_id) %>
```
okriuchykhin, line 4: Can you move it to helper?

app/views/reports/new/modal/_module_contents_inner.html.erb
```diff
@@ -74,6 +74,11 @@
             <%= form.check_box "module_repository_#{repository.id}", label: repository.name.capitalize, data: { id: repository.id } %>
           </li>
         <% end %>
+        <% RepositorySnapshot.with_deleted_parent_by_team(@project.team).each do |repository| %>
```
okriuchykhin, line 77: Please move it to controller or helper

okriuchykhin: And snapshots will be shown separately from live repositories?

app/models/my_module.rb
```diff
@@ -225,6 +225,37 @@ def live_and_snapshot_repositories_list
...
+          else
+            Repository.where(id: rep_or_snap.parent_id).pluck(:id) +
+              RepositorySnapshot.where(parent_id: rep_or_snap.parent_id).pluck(:id)
+          end
+
+    report_elements.where(repository_id: ids).update(repository_id: rep_or_snap.id)
+  end
+
+  def selected_snapshot_for_repo(repository_id, exclude_snpashot_ids: [])
+    repository_snapshots.where(parent_id: repository_id).where.not(id: exclude_snpashot_ids).selected.first
```
okriuchykhin, line 256: Just add 'selected' condition to where. There is no much sense to have scope for this, as long as there can be only one selected snapshot.

app/views/reports/new/modal/_module_contents_inner.html.erb
urbanrotnik: Yes, is this a problem? And the have extra label "(Deleted)", as Sasa suggested.

urbanrotnik: And I would keep it here. Will be hard to find these helpers methods because we render those views in this dynamic magic controller methods..

app/controllers/my_module_repository_snapshots_controller.rb
```diff
@@ -86,12 +87,21 @@ def select
                 .update(selected: nil)
       repository_snapshot.update!(selected: true)
     end
+    @my_module.update_report_references(params[:repository_id] || params[:repository_snapshot_id])
```
okriuchykhin, line 90: Passing unchecked params directly to update method?

app/models/my_module.rb
```diff
@@ -225,6 +225,36 @@ def live_and_snapshot_repositories_list
...
+
+    selected_snapshot_for_repo(parent_id, exclude_snpashot_ids: exclude_snpashot_ids) ||
+      assigned_repositories&.where(id: parent_id)&.first ||
+      repository_snapshots
+        .where(parent_id: parent_id)
+        .where.not(id: exclude_snpashot_ids)
+        .order(updated_at: :desc).first
+  end
+
+  def update_report_references(rep_or_snap_id)
```
okriuchykhin, line 241: update_report_repository_references

app/helpers/reports_helper.rb
```diff
@@ -122,6 +122,13 @@ def font_awesome_cdn_link_tag
     )
   end
 
+  def assign_repository_or_snapshot(my_module, element_id)
+    org_repo = Repository.find_by(id: element_id) if element_id
```
okriuchykhin, line 126: Does it mean organisation repository?

app/controllers/my_module_repository_snapshots_controller.rb
urbanrotnik: Not sure what else should I check here?

app/models/repository_snapshot.rb
```diff
@@ -13,10 +15,23 @@ class RepositorySnapshot < RepositoryBase
   validates :status, presence: true
   validate :only_one_selected_for_my_module, if: ->(obj) { obj.changed.include? :selected }
 
+  scope :with_deleted_parent_by_team, lambda { |team|
```
okriuchykhin, line 18: Is it still used?

okriuchykhin: Ok, seems like

PR #2572, opened 2020-05-15
Prevent the inventory items editing while snapshot is being created [SCI-4629]

okriuchykhin:
Jira ticket: [SCI-4629](https://biosistemika.atlassian.net/browse/SCI-4629)

### What was done
Prevent the inventory items editing while snapshot is being created


radiokills:
Nice cleanup.

PR #2574, opened 2020-05-18
Fix excel header parsing [SCI-4634]

urbanrotnik:
Jira ticket: [SCI-4634](https://biosistemika.atlassian.net/browse/SCI-4634)

### What was done
Skip excel cells formating for header

### DIFF
--- a/app/services/spreadsheet_parser.rb
+++ b/app/services/spreadsheet_parser.rb
@@ -39,20 +39,20 @@ class SpreadsheetParser
     rows = spreadsheet_enumerator(sheet)
     header = []
     columns = []
-    i = 1
-    rows.each do |row_values|
-      # Creek XLSX parser returns Hash of the row, Roo - Array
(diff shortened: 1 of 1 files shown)


app/services/spreadsheet_parser.rb
```diff
@@ -42,7 +42,7 @@ def self.first_two_rows(sheet)
     i = 1
     rows.each do |row_values|
       # Creek XLSX parser returns Hash of the row, Roo - Array
-      row = parse_row(row_values, sheet)
+      row = parse_row(row_values, sheet, header: i.equal?(1))
```
radiokills, line 45: oh god :)

Can you refactor this please, use `each_with_index `and don't forget to compare it with 0 instead of 1

urbanrotnik: Refactored a bit more. Check again please :D

PR #2594, opened 2020-05-22
Fix button visibility [SCI-4663]

urbanrotnik:
Jira ticket: [SCI-4663](https://biosistemika.atlassian.net/browse/SCI-4663)

### DIFF
--- a/app/views/my_modules/repositories/_full_view_sidebar.html.erb
+++ b/app/views/my_modules/repositories/_full_view_sidebar.html.erb
@@ -40,13 +40,18 @@
           </button>
         </div>
       <% end %>
-      <%= render partial: 'my_modules/repositories/full_view_version', collection: @repository_snapshots, as: :repository_snapshot %>
+      <%= render partial: 'my_modules/repositories/full_view_version',
+                 collection: @repository_snapshots,
+                 as: :repository_snapshot,
+                 locals: { can_delete_snapshot: can_manage_module?(@my_module) } %>
     </div>
   </div>
 
-  <div class="list-group-item text-center">
-    <a id="setDefaultVersionButton" class="btn btn-primary default-version-button hidden" data-select-path="<%= my_module_select_default_snapshot_path(@my_module) %>">
-      <%= t('my_modules.repository.snapshots.full_view.set_default_button') %>
-    </a>
-  </div>
+  <% if can_manage_module?(@my_module) %>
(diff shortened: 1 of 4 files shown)


app/views/my_modules/repositories/_full_view_version.html.erb
```diff
@@ -28,9 +28,11 @@
       <% end %>
     </div>
     <div class="col-xs-4">
-      <a class="pull-right btn version-button delete-snapshot-button" href="#" data-action-path="<%= my_module_repository_snapshot_path(@my_module, repository_snapshot) %>">
-        <i class="fas fa-trash"></i>
-      </a>
+      <% if can_manage_module?(@my_module) %>
```
okriuchykhin, line 31: This check will be executed separately for each snapshot? Can you move it out from the loop and use local variable?

urbanrotnik: Updated. Anyway, the value of this check is loaded in the cache and it's not querying each iteration

app/views/my_modules/repositories/_full_view_sidebar.html.erb
```diff
@@ -40,13 +40,18 @@
           </button>
         </div>
       <% end %>
-      <%= render partial: 'my_modules/repositories/full_view_version', collection: @repository_snapshots, as: :repository_snapshot %>
+      <%= render partial: 'my_modules/repositories/full_view_version',
+                 collection: @repository_snapshots,
+                 as: :repository_snapshot,
+                 locals: {can_delete_snapshot: can_manage_module?(@my_module)}%>
```
okriuchykhin, line 46: spaces are missing near { and }

app/views/my_modules/repositories/_full_view_sidebar.html.erb
```diff
@@ -40,13 +40,18 @@
           </button>
         </div>
       <% end %>
-      <%= render partial: 'my_modules/repositories/full_view_version', collection: @repository_snapshots, as: :repository_snapshot %>
+      <%= render partial: 'my_modules/repositories/full_view_version',
+                 collection: @repository_snapshots,
+                 as: :repository_snapshot,
+                 locals: { can_delete_snapshot: can_manage_module?(@my_module) }%>
```
okriuchykhin, line 46: Space is missing after }

PR #2630, opened 2020-06-01
Add repositories list view [SCI-4677]

aignatov-bio:
Jira ticket: [SCI-4677](https://biosistemika.atlassian.net/browse/SCI-4677)

### What was done
Add a repositories list view.

### DIFF
--- a/app/views/repositories/index.html.erb
+++ b/app/views/repositories/index.html.erb
@@ -5,7 +5,7 @@
   <%= render partial: "sidebar", locals: { repositories: @repositories } %>
   <div class="content-pane flexible">
     <div class="content-header">
-      <h1>Inventories</h1>
+      <h1><%= t('libraries.index.head_title') %></h1>
     </div>
     <div class="content-body">
     </div>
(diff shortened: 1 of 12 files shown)


app/views/repositories/index.html.erb
```diff
@@ -3,27 +3,46 @@
...
-  <div class="content-pane">
-    <div class="jumbotron text-center" style="margin-top:12%">
-      <strong><%=t 'libraries.index.no_libraries.text' %></strong>
+  <div class="content-pane flexible">
+    <div class="content-header">
+      <h1><%= t('libraries.index.head_title') %></h1>
+    </div>
+    <div class="content-body">
+    </div>
+    <template id="activeRepositoriesListTable">
```
urbanrotnik, line 12: Should we check how `template` is doin with IE? 
Here is marked as not supported
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template

aignatov-bio: it supported partially, it will be moved out from body. But anyway SciNote not correctly working in IE.

PR #2651, opened 2020-06-09
Archiving & restoring rows [SCI-4706]

urbanrotnik:
Jira ticket: [SCI-4706](https://biosistemika.atlassian.net/browse/SCI-4706)

### What was done
Archiving and restoring inventory rows

### Note 
- View/Archived toggle is missing, so the restore button should be added in the scope of that page and task. JS + backend implemented for both scenarios and it is working. 
- Scopes also needs to be updated. For now, I added `.active` for `repositroy_rows/index` action

### DIFF
--- a/app/controllers/repository_rows_controller.rb
+++ b/app/controllers/repository_rows_controller.rb
@@ -129,7 +129,7 @@ class RepositoryRowsController < ApplicationController
 
   def copy_records
     duplicate_service = RepositoryActions::DuplicateRows.new(
-      current_user, @repository, copy_records_params
+      current_user, @repository, selected_rows_in_repo_params
     )
     duplicate_service.call
     render json: {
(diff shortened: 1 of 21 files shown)


app/controllers/repository_rows_controller.rb
```diff
@@ -170,6 +171,32 @@ def assigned_task_list
     }
   end
 
+  def archive_records
+    service = RepositoryActions::ArchiveRowsService.call(repository: @repository,
+                                                         repository_rows: copy_records_params,
```
mlorb, line 176: Rename `copy_records_params`

config/initializers/extends.rb
```diff
@@ -282,7 +282,9 @@ class Extends
     protocol_description_in_task_edited: 137,
     set_task_start_date: 138,
     change_task_start_date: 139,
-    remove_task_start_date: 140
+    remove_task_start_date: 140,
+    archive_inventory_item: 141,
+    restore_inventory_item: 142
```
mlorb, line 287: 141 is already used in develop branch, use here instead 142 and 143.

PR #2658, opened 2020-06-15
Fix update_canvas [SCI-4762]

urbanrotnik:
Jira ticket: [SCI-4762](https://biosistemika.atlassian.net/browse/SCI-4762)

### What was done
Update order of actions in update_canvas

### DIFF
--- a/app/models/experiment.rb
+++ b/app/models/experiment.rb
@@ -132,14 +132,25 @@ class Experiment < ApplicationRecord
     positions,
     current_user
   )
-    cloned_modules = []
     begin
       with_lock do
-        # First, add new modules
+        # Start with archiving to release positions for new tasks
+        archive_modules(to_archive, current_user) if to_archive.any?
+
+        # Update only existing tasks positions to release positions for new tasks
+        existing_positions = positions
+                             .slice(*positions.keys.map { |k| k unless k.to_s.start_with?('n') }.compact)
+        update_module_positions(existing_positions) if existing_positions.any?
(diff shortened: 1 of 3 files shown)


app/models/experiment.rb
```diff
@@ -325,7 +333,7 @@ def move_modules(to_move, current_user)
     to_move.each do |id, experiment_id|
       my_module = my_modules.find_by_id(id)
       experiment = project.experiments.find_by_id(experiment_id)
-      experiment_org = my_module.experiment
+      experiment_org = my_module&.experiment
```
okriuchykhin, line 336: rename it to experiment_original

okriuchykhin: and you can move it below 'next', without safe nav

app/models/experiment.rb
```diff
@@ -132,14 +132,25 @@ def update_canvas(
...
+        # Start with archiving to release positions for new tasks
+        archive_modules(to_archive, current_user) if to_archive.any?
+
+        # Update only existing tasks positions to release positions for new tasks
+        exisitng_positions = positions
+                             .slice(*positions.keys.map { |k| k unless k.to_s.start_with?('n') }.compact)
+        update_module_positions(exisitng_positions) if exisitng_positions.any?
+
+        # Move only existing tasks to release positions for new tasks
+        exisitng_to_moves = to_move
```
okriuchykhin, line 146: exisitng ?

PR #2660, opened 2020-06-15
Improve user input filtering on user settings page and for task tags [SCI-4707]

okriuchykhin:
Jira ticket: [SCI-4707](https://biosistemika.atlassian.net/browse/SCI-4707)

### What was done
Improve user input filtering on user settings page and for task tags

### DIFF
--- a/app/assets/javascripts/shared/inline_editing.js
+++ b/app/assets/javascripts/shared/inline_editing.js
@@ -77,11 +77,11 @@ var inlineEditing = (function() {
           viewData = result[container.data('response-field')];
         } else {
           // By default we just copy value from input string
-          viewData = inputField(container).val();
+          viewData = $('<div>').text(inputField(container).val()).html();
         }
 
         container.find('.view-mode')
-          .text(viewData)
+          .html(viewData)
           .removeClass('hidden');
         container
           .attr('data-original-name', inputField(container).val())
--- a/app/assets/javascripts/shared/inline_editing.js
+++ b/app/assets/javascripts/shared/inline_editing.js
@@ -63,6 +63,7 @@ var inlineEditing = (function() {
     } else {
       params = { [fieldToUpdate]: inputField(container).val() };
     }
+
     $.ajax({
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/shared/inline_editing.js
```diff
@@ -80,7 +81,7 @@ var inlineEditing = (function() {
         }
 
         container.find('.view-mode')
-          .html(viewData)
+          .text(viewData)
```
aignatov-bio, line 84: For smart annotation fields, we still need to use html. You can check if the field for smart annotation `container.data('smart-annotation')`. As I remember now we use it in comments.

app/controllers/my_module_tags_controller.rb
```diff
@@ -112,18 +114,20 @@ def destroy
...
-    if !(tags.map { |i| i[:label] }.include? params[:query]) && !params[:query].empty?
-      tags = [{ value: 0, label: params[:query], params: { color: nil } }] + tags
+    all_tags = @my_module.experiment.project.tags
+    tags = all_tags.where.not(id: assigned_tags)
+                   .search(current_user, false, params[:query])
+                   .select(:id, :name, :color)
+                   .limit(6)
+
+    tags = tags.map do |tag|
+      { value: tag.id, label: sanitize_input(tag.name), params: { color: sanitize_input(tag.color) } }
```
aignatov-bio, line 124: Colors predefined list in constants

okriuchykhin: And? Please give me a link to the lines where this is enforced?

aignatov-bio: Well, yes someone can force set custom color. But what he can do with 7 characters?

okriuchykhin: And if later this limit is changed, what will happen to this code?

aignatov-bio: In this case yes, it can become issue

PR #2661, opened 2020-06-16
Update default scope [SCI-4740] [SCI-4687][SCI-4746]

urbanrotnik:
Jira ticket: [SCI-4740](https://biosistemika.atlassian.net/browse/SCI-4740)

### What was done
Add default scope and unscope where necessary

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -48,7 +48,8 @@ class Activity < ApplicationRecord
   }
 
   scope :repositories_joins, lambda {
-    joins("LEFT JOIN repositories ON subject_type = 'RepositoryBase' AND subject_id = repositories.id")
+    joins("LEFT JOIN repositories ON subject_type = 'RepositoryBase' AND subject_id = repositories.id " \
+            "AND repositories.archived != TRUE")
   }
 
   scope :reports_joins, lambda {
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -34,6 +34,9 @@ class Repository < RepositoryBase
       .distinct
   }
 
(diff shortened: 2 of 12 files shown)


app/models/repository.rb
```diff
@@ -22,7 +22,8 @@ class Repository < RepositoryBase
             length: { maximum: Constants::NAME_MAX_LENGTH }
 
   scope :accessible_by_teams, lambda { |teams|
-    left_outer_joins(:team_repositories)
+    with_archived
```
aignatov-bio, line 25: Do we need really it here? We are using it to get an active repository scope. Check:
`at_who_controller.rb`
`my_module_repositories_controller.rb`

app/models/activity.rb
```diff
@@ -48,7 +48,8 @@ class Activity < ApplicationRecord
   }
 
   scope :repositories_joins, lambda {
-    joins("LEFT JOIN repositories ON subject_type = 'RepositoryBase' AND subject_id = repositories.id")
+    joins("LEFT JOIN repositories ON subject_type = 'RepositoryBase' AND subject_id = repositories.id " \
+            "AND repositories.archived != TRUE")
```
aignatov-bio, line 52: I think it will join only archived inventories

urbanrotnik: It's `!= TRUE`. So only active inventories

app/models/repository.rb
urbanrotnik: Hmm... You are correct. Based on the name, I would say it should include archived, but it's better to explicitly call `.with_archived` where we need them.

PR #2676, opened 2020-06-22
Revert default_scope for active repositories [SCI-4786]

mlorb:
Jira ticket: [SCI-4786](https://biosistemika.atlassian.net/browse/SCI-4786)

### What was done
default_scope removed

### DIFF
--- a/app/services/dashboard/recent_work_service.rb
+++ b/app/services/dashboard/recent_work_service.rb
@@ -32,6 +32,7 @@ module Dashboard
                       .projects_joins(:from_experiments)
                       .repositories_joins
                       .reports_joins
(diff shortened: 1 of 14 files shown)


app/services/dashboard/recent_work_service.rb
```diff
@@ -32,6 +32,7 @@ def call
                       .projects_joins(:from_experiments)
                       .repositories_joins
                       .reports_joins
+                      .where('repositories.archived IS NOT TRUE')
```
urbanrotnik, line 35: I think we have same in `Activity.rb:51` Check if doing same thing and remove one.

PR #2691, opened 2020-06-26
Improve repository state migration [SCI-4792]

aignatov-bio:
Jira ticket: [SCI-4792](https://biosistemika.atlassian.net/browse/SCI-4792)

### What was done
Improve repository state migration

### DIFF
--- a/db/migrate/20200622155632_repository_table_state_update.rb
+++ b/db/migrate/20200622155632_repository_table_state_update.rb
@@ -18,7 +18,7 @@ class RepositoryTableStateUpdate < ActiveRecord::Migration[6.0]
 
       state['order'][0][0] = order_state.to_i + 2 if order_state.to_i > 5
 
-      if state['columns'].class == Hash
+      if state['columns'].is_a? Hash
         columns_array = []
         state['columns'].each { |k, v| columns_array[k.to_i] = v }
         state['columns'] = columns_array
@@ -36,7 +36,8 @@ class RepositoryTableStateUpdate < ActiveRecord::Migration[6.0]
       end
       table_state.update(state: state)
     rescue StandardError
-      next
+      # Corrupted state will be cleared and default state will be generated on load
+      table_state.update(state: {})
     end
   end
 end
--- a/db/migrate/20200622155632_repository_table_state_update.rb
+++ b/db/migrate/20200622155632_repository_table_state_update.rb
@@ -7,18 +7,36 @@ class RepositoryTableStateUpdate < ActiveRecord::Migration[6.0]
 
   def up
     TempStateModel.find_each do |table_state|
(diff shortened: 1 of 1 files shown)


db/migrate/20200622155632_repository_table_state_update.rb
```diff
@@ -8,18 +8,35 @@ class TempStateModel < ApplicationRecord
...
                                   'regex' => false, 'smart' => true, 'search' => '', 'caseInsensitive' => true
                                 },
                                 'visible' => false)
       end
-      state['ColReorder'] = state['ColReorder'].map { |i| i > 5 ? i + 2 : i } + [6, 7]
+      if state['ColReorder']
+        state['ColReorder'] = state['ColReorder'].map { |i| i.to_i > 5 ? i.to_i + 2 : i.to_i } + [6, 7]
+      end
       table_state.update(state: state)
+    rescue StandardError
```
urbanrotnik, line 38: What kind of errors do we expect here?

okriuchykhin: Shouldn't it reset state to default in case of an error?

db/migrate/20200622155632_repository_table_state_update.rb
```diff
@@ -8,18 +8,35 @@ class TempStateModel < ApplicationRecord
...
+
+      unless order_state
+        order = state.dig('order', '0')
+        state['order'] = [order]
+        order_state = state.dig('order', 0, 0)
+      end
+
+      state['order'][0][0] = order_state.to_i + 2 if order_state.to_i > 5
+
+      if state['columns'].class == Hash
```
okriuchykhin, line 21: .is_a?

db/migrate/20200622155632_repository_table_state_update.rb
```diff
@@ -8,18 +8,35 @@ class TempStateModel < ApplicationRecord
   def up
     TempStateModel.find_each do |table_state|
       state = table_state.state
-      order_state = state['order'][0][0]
-      state['order'][0][0] = order_state + 2 if order_state > 5
+      order_state = state.dig('order', 0, 0)
+
+      unless order_state
+        order = state.dig('order', '0')
```
okriuchykhin, line 14: isn't it already checked in order_state?

aignatov-bio: There is old state format, where some fields use Hash instead of Array

db/migrate/20200622155632_repository_table_state_update.rb
aignatov-bio: I will reset state to empty and then default state will be generated on load

PR #2694, opened 2020-06-30
Show assigned users modal also for view only [SCI-4790]

urbanrotnik:
Jira ticket: [SCI-4790](https://biosistemika.atlassian.net/browse/SCI-4790)

### What was done

Show modal of assigned users also if the user does not have permissions for managing users on my_module. 
Buttons for adding and removing users will be hidden if the user does not have permission. 
+ button is shown only if the user can manage permissions and there is some unassigned users left

Check also [GitLab PR](https://gitlab.com/biosistemika/scinote/addons/electronic_signatures/-/merge_requests/77)

### DIFF
--- a/app/views/user_my_modules/_index.html.erb
+++ b/app/views/user_my_modules/_index.html.erb
@@ -1,35 +1,24 @@
 <% user_my_modules = @my_module.user_my_modules %>
-<% if can_manage_users_in_module?(@my_module) %>
-  <a class="task-assigned-users manage-users-link <%= 'empty' unless user_my_modules.present? %>"
-     data-remote="true"
-     href="<%= my_module_users_edit_path(@my_module, format: :json) %>"
-     data-module-id="<%= @my_module.id %>"
-     data-module-users-url="<%= my_module_user_my_modules_url(@my_module, format: :json) %>">
-    <% if user_my_modules.present? %>
-      <% user_my_modules.each do |user_my_module| %>
-        <%  user = user_my_module.user %>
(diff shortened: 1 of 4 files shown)


app/views/user_my_modules/_index.html.erb
```diff
@@ -1,35 +1,24 @@
...
     <% user_my_modules.each do |user_my_module| %>
       <%  user = user_my_module.user %>
       <span class='global-avatar-container'>
         <%= image_tag avatar_path(user, :icon_small) %>
       </span>
     <% end %>
-  </div>
-<% end %>
+    <% if can_manage_users_in_module?(@my_module) && @my_module.unassigned_users.any? %>
+      <span class='global-avatar-container assign-new-user'>
```
okriuchykhin, line 15: Use " here, to be consistent

app/views/user_my_modules/_index.html.erb
```diff
@@ -1,35 +1,24 @@
...
-      <span class="empty-label">
-        <%= t('my_modules.details.no_assigned_users') %>
-      </span>
-    <% end %>
-  </a>
-<% else %>
-  <div class="task-assigned-users">
+<a class="task-assigned-users manage-users-link <%= 'empty' unless user_my_modules.present? %>"
+   data-remote="true"
+   href="<%= my_module_users_edit_path(@my_module, format: :json) %>"
```
okriuchykhin, line 4: Can we skip 'format' here?

PR #2697, opened 2020-07-01
Add 2fa to user settings page [SCI-4755]

aignatov-bio:
Jira ticket: [SCI-4755](https://biosistemika.atlassian.net/browse/SCI-4755)

### What was done
Add 2fa to user settings page

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -628,9 +628,7 @@ class User < ApplicationRecord
     totp.verify(otp, drift_behind: 10)
   end
 
-  def ensure_2fa_token!
-    return if otp_secret
-
+  def assign_2fa_token!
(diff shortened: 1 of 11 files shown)


app/models/user.rb
```diff
@@ -630,6 +628,21 @@ def valid_otp?(otp)
     totp.verify(otp, drift_behind: 10)
   end
 
+  def ensure_2fa_token!
```
urbanrotnik, line 631: Hmm, I think this naming `ensure.. ` + an `if condition` does not make any sense because we are regenerating this token on each turn on/off. 

I would rename it to something like (re)assign_otp_code. Do you have a better idea?

PR #2714, opened 2020-07-08
Enable Content Security Policy [SCI-4796]

okriuchykhin:
Jira ticket: [SCI-4796](https://biosistemika.atlassian.net/browse/SCI-4796)

### What was done
Enable Content Security Policy

### DIFF
--- a/config/initializers/content_security_policy.rb
+++ b/config/initializers/content_security_policy.rb
@@ -4,22 +4,22 @@
 # For further information see the following documentation
 # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy
 
-# Rails.application.config.content_security_policy do |policy|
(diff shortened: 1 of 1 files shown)


config/initializers/content_security_policy.rb
```diff
@@ -4,22 +4,22 @@
 # For further information see the following documentation
 # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy
```
urbanrotnik, line 6: Won't this prevent `marvinJS` from running?

okriuchykhin: we can whitelist it

urbanrotnik: Cool, we should I guess

PR #2740, opened 2020-07-17
Implement Status labels on a Task card [SCI-4823]

mlorb:
Jira ticket: [SCI-4823](https://biosistemika.atlassian.net/browse/SCI-4823)

### What was done
- Implement Status labels on a Task card
- Restyle task card to fit the status label, but still stays the same size

### Todo
- Replace fake data after backend will be finished.

### DIFF
--- a/app/views/canvas/full_zoom/_my_module.html.erb
+++ b/app/views/canvas/full_zoom/_my_module.html.erb
@@ -37,6 +37,9 @@
     <% else %>
       <%= render partial: "my_modules/card_due_date_label.html.erb", locals: { my_module: my_module, format: :full_date } %>
     <% end %>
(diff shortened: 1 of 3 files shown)


app/views/canvas/full_zoom/_my_module.html.erb
```diff
@@ -37,6 +37,9 @@
     <% else %>
       <%= render partial: "my_modules/card_due_date_label.html.erb", locals: { my_module: my_module, format: :full_date } %>
     <% end %>
+    <div class="status-label" style="--state-color: #2DBE61">
+      Completed
```
okriuchykhin, line 41: What's this?

mlorb: mock data, we don't have backend

PR #2741, opened 2020-07-17
Add export for assigned items [SCI-4748]

aignatov-bio:
Jira ticket: [SCI-4748](https://biosistemika.atlassian.net/browse/SCI-4748)

### What was done
Add export for assigned items

### DIFF
--- a/app/services/repository_zip_export.rb
+++ b/app/services/repository_zip_export.rb
@@ -5,16 +5,25 @@ require 'csv'
 module RepositoryZipExport
   def self.generate_zip(params, repository, current_user)
     # Fetch rows in the same order as in the currently viewed datatable
-    ordered_row_ids = params[:row_ids]
-    id_row_map = RepositoryRow.where(id: ordered_row_ids,
-                                     repository: repository)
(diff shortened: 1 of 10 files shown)


app/services/repository_zip_export.rb
```diff
@@ -5,16 +5,25 @@
 module RepositoryZipExport
   def self.generate_zip(params, repository, current_user)
     # Fetch rows in the same order as in the currently viewed datatable
-    ordered_row_ids = params[:row_ids]
-    id_row_map = RepositoryRow.where(id: ordered_row_ids,
-                                     repository: repository)
-                              .index_by(&:id)
-    ordered_rows = ordered_row_ids.collect { |id| id_row_map[id.to_i] }
+    if params[:my_module_id]
```
urbanrotnik, line 8: Should we write a test for this part with a context `when receive RepositorySnapshot`?

aignatov-bio: We need to refactor factories, for adding snapshots. I think better to make it in scope of the separate task.

PR #2750, opened 2020-07-22
Add modal for status flow [SCI-4835]

aignatov-bio:
Jira ticket: [SCI-4835](https://biosistemika.atlassian.net/browse/SCI-4835)

### What was done
Add modal for status flow


app/views/my_modules/modals/_status_flow_modal_body.html.erb
```diff
@@ -0,0 +1,22 @@
+ <div class="status-flow">
+  <% my_module_statuses.each_with_index do |status, i| %>
```
okriuchykhin, line 2: which ordering is used here?

PR #2757, opened 2020-07-24
Extend asset API upload to support base64 [SCI-4813]

urbanrotnik:
Jira ticket: [SCI-4813](https://biosistemika.atlassian.net/browse/SCI-4813)

### What was done
Refactored update and create actions on: 
- Attachments controller
- Results controller (for File Result) 
Now accepts previous behavior (multipart form) and new base64 encoded file 
- In docs only Base64

### Docs:
Check https://github.com/biosistemika/scinote-api-v1-docs/pull/20

### DIFF
--- a/app/controllers/api/v1/assets_controller.rb
+++ b/app/controllers/api/v1/assets_controller.rb
@@ -28,7 +28,7 @@ module Api
           blob = ActiveStorage::Blob.create_after_upload!(
             io: StringIO.new(Base64.decode64(asset_params[:file_data])),
             filename: asset_params[:file_name],
-            content_type: asset_params[:content_type]
+            content_type: asset_params[:file_type]
           )
           asset = @step.assets.new(file: blob)
         end
@@ -48,7 +48,7 @@ module Api
 
         return params.require(:data).require(:attributes).permit(:file) if @form_multipart_upload
 
-        attr_list = %i(file_data content_type file_name)
+        attr_list = %i(file_data file_type file_name)
         params.require(:data).require(:attributes).require(attr_list)
         params.require(:data).require(:attributes).permit(attr_list)
       end
@@ -59,7 +59,7 @@ module Api
       end
 
       def check_upload_type
-        @form_multipart_upload = true if params.dig(:data, :attributes)[:file]
+        @form_multipart_upload = true if params.dig(:data, :attributes, :file)
       end
     end
   end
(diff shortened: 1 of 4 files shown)


app/controllers/api/v1/assets_controller.rb
```diff
@@ -45,6 +57,10 @@ def load_asset
         @asset = @step.assets.find(params.require(:id))
         raise PermissionError.new(Asset, :read) unless can_read_protocol_in_module?(@asset.step.protocol)
       end
+
+      def check_upload_type
+        @form_multipart_upload = true if params.dig(:data, :attributes)[:file]
```
okriuchykhin, line 62: why [:file] is outside dig?

app/controllers/api/v1/results_controller.rb
```diff
@@ -139,8 +158,14 @@ def result_file_params
...
-        prms.require(:attributes).require(:file)
-        prms.dig(:attributes).permit(:file)
+        if prms.require(:attributes)[:file]
+          @form_multipart_upload = true
+          return prms.dig(:attributes).permit(:file)
+        end
+        attr_list = %i(file_data file_type file_name)
+
+        prms.require(:attributes).require(attr_list)
+        prms.dig(:attributes).permit(attr_list)
```
okriuchykhin, line 168: attributes were already checked by require on previous line

urbanrotnik: Yes, but we want to permit only what is in the attr_list? Or what do you mean? I am not sure. 

Another question. Do you think is ok to send `file_data`, `file_type`, and `file_name`. Or we want to combine data and type in URL base64 string? 
I didn't find what is "more correct" way, but I thin URL base64 string is more for inline images in HTML.

app/controllers/api/v1/assets_controller.rb
urbanrotnik: Updated

app/controllers/api/v1/results_controller.rb
okriuchykhin: I mean that you can simply call prms.permit(attributes: attr_list)

urbanrotnik: Then we have to call `prms.permit(attributes: attr_list)[:attributes]` because we want to return only nested hash in attributes?

PR #2761, opened 2020-07-27
Fix step position adjusting on move and destroy [SCI-4880]

okriuchykhin:
Jira ticket: [SCI-4880](https://biosistemika.atlassian.net/browse/SCI-4880)

### What was done
Fix step position adjusting on move and destroy

### DIFF
--- a/app/models/step.rb
+++ b/app/models/step.rb
@@ -9,7 +9,7 @@ class Step < ApplicationRecord
             presence: true,
             length: { maximum: Constants::NAME_MAX_LENGTH }
   validates :description, length: { maximum: Constants::RICH_TEXT_MAX_LENGTH }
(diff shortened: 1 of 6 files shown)


app/models/step.rb
```diff
@@ -90,6 +90,36 @@ def my_module
...
+
+    step_below = protocol.steps.find_by(position: position + 1)
+
+    return unless step_below
+
+    transaction do
+      step_below.position = position
+      update!(position: -1)
+      step_below.save!
+      update!(position: step_below.position + 1)
```
radiokills, line 119: would this work?:
update([step_below.id, self.id], {position: position}, {position: step_below.position})

radiokills (approved):

PR #2796, opened 2020-08-18
Add factories for new task flows models [SCI-4934]

mlorb:
Jira ticket: [SCI-4934](https://biosistemika.atlassian.net/browse/SCI-4934)

### What was done
Add factories for new task flows models. Create `status_flow` before every creation of `my_module`.

### DIFF
--- a/spec/factories/my_module_status_conditions.rb
+++ b/spec/factories/my_module_status_conditions.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :my_module_status_condition do
+    my_module_status
+  end
+end
--- a/spec/factories/my_module_status_conditions.rb
+++ b/spec/factories/my_module_status_conditions.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 FactoryBot.define do
-  factory :my_module_status_condition do
+  factory :active_condition, class: 'MyModuleStatusConditions::Active' do
     my_module_status
   end
 end
--- a/spec/factories/my_module_status_consequences.rb
+++ b/spec/factories/my_module_status_consequences.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 FactoryBot.define do
-  factory :my_module_status_consequence do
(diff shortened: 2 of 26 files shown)


spec/factories/my_module_status_conditions.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
```
urbanrotnik, line 1: We do not need this factory, this is a superclass and should not be saved.

spec/factories/my_module_status_consequences.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
```
urbanrotnik, line 1: ditto

spec/factories/my_module_status_implications.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
```
urbanrotnik, line 1: ditto

spec/factories/my_module_status_conditions.rb
urbanrotnik: We can implement factory for specific conditions/consequences or implications. Like `read_only.rb`

spec/factories/my_modules.rb
```diff
@@ -7,6 +7,7 @@
     sequence(:x) { |n| n }
     workflow_order { MyModule.where(experiment_id: experiment.id).count + 1 }
     experiment
+    my_module_status
```
urbanrotnik, line 10: I wouldn't add it here. Could be a separate trait, because it is optional.

urbanrotnik (changes requested):
Few thoughts on this: 
- I would not force `my_module` having status always. It is optional. And if we have default status flow, it will be assigned. But in most cases, we do not need it. It will slow down tests if we always generate new status with its parent status flow. Factory can have a trait `:with_status`
-  I would remove status also from the controllers. If we want/need status + status_flows in specific controller, we can generate it in `before` hook with "seeds" generation like `MyModuleStatusFlow.ensure_default`. 
- We do not need factories for superclasses (implications, consequences, and conditions) Should not persist without specific STI type. Currently, we can instance and save it in DB. We can add validation here, and add factories for specific implications...

PR #2798, opened 2020-08-20
Add API endpoints for workflows [SCI-4831]

aignatov-bio:
Jira ticket: [SCI-4831](https://biosistemika.atlassian.net/browse/SCI-4831)

### What was done
Add API endpoints for workflows

### DIFF
--- a/app/controllers/api/v1/tasks_controller.rb
+++ b/app/controllers/api/v1/tasks_controller.rb
@@ -59,7 +59,7 @@ module Api
       def task_params
         raise TypeError unless params.require(:data).require(:type) == 'tasks'
 
-        params.require(:data).require(:attributes).permit(%i(name x y description state))
+        params.require(:data).require(:attributes).permit(%i(name x y description state my_module_status_id))
       end
 
       def load_task_for_managing
--- a/app/controllers/api/v1/tasks_controller.rb
+++ b/app/controllers/api/v1/tasks_controller.rb
@@ -30,7 +30,7 @@ module Api
       def create
         raise PermissionError.new(MyModule, :create) unless can_manage_experiment?(@experiment)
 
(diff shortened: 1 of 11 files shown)


app/serializers/api/v1/task_serializer.rb
```diff
@@ -4,14 +4,39 @@ module Api
...
 
+      def status_id
+        object.my_module_status_id
+      end
+
+      def status_name
+        object.my_module_status.name
+      end
+
+      def prev_status_id
```
urbanrotnik, line 24: You can use belongs_to for previous and next status with its serializer...?

urbanrotnik (changes requested):
Check comments and implement request specs for new controllers. You can add also tests for schema matching

app/serializers/api/v1/task_serializer.rb
aignatov-bio: It does not belong to my_module

app/controllers/api/v1/tasks_controller.rb
```diff
@@ -59,7 +60,7 @@ def activities
       def task_params
         raise TypeError unless params.require(:data).require(:type) == 'tasks'
 
-        params.require(:data).require(:attributes).permit(%i(name x y description state))
```
urbanrotnik, line 62: Can you set not first my_module_status on the `create` method also?  Or will fail with validation for check_status?

urbanrotnik (approved):
Check one comment, if it is covered, merge

app/serializers/api/v1/task_serializer.rb
urbanrotnik: But status `belongs_to`...

PR #2810, opened 2020-09-01
Add error handling for state change [SCI-4968]

aignatov-bio:
Jira ticket: [SCI-4968](https://biosistemika.atlassian.net/browse/SCI-4968)

### What was done
 Add error handling for state change

### DIFF
--- a/app/models/my_module.rb
+++ b/app/models/my_module.rb
@@ -22,6 +22,7 @@ class MyModule < ApplicationRecord
   validate :coordinates_uniqueness_check, if: :active?
   validates :completed_on, presence: true, if: proc { |mm| mm.completed? }
 
+  validate :check_status_order, if: :my_module_status_id_changed?
   validate :check_status_conditions, if: :my_module_status_id_changed?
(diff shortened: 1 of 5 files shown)


app/models/my_module.rb
```diff
@@ -7,8 +7,9 @@ class MyModule < ApplicationRecord
   enum state: Extends::TASKS_STATES
 
   before_create :create_blank_protocol
+  before_create :assign_default_status_flow
   before_validation :set_completed_on, if: :state_changed?
-  before_validation :assign_default_status_flow
+
```
urbanrotnik, line 12: delete this line?

spec/controllers/my_modules_controller_spec.rb
```diff
@@ -153,23 +153,24 @@
       end
     end
 
-    context 'when status not found' do
+    context 'when status not exist' do
```
urbanrotnik, line 156: Hm... I think a better approach to test validation is with the model spec. It is not obvious why you get 422 here.

PR #2816, opened 2020-09-02
Fix css for status flows [SCI-4970]

aignatov-bio:
Jira ticket: [SCI-4960](https://biosistemika.atlassian.net/browse/SCI-4760)

### What was done
Fix css for status flows


urbanrotnik (approved):
Fix tests if fail again

PR #2824, opened 2020-09-04
Refactor smart annotation front-end [SCI-4949]

aignatov-bio:
Jira ticket: [SCI-4949](https://biosistemika.atlassian.net/browse/SCI-4949)

### What was done
Refactor smart annotation front-end


app/controllers/at_who_controller.rb
```diff
@@ -32,7 +32,7 @@ def menu_items
   end
 
   def rep_items
-    repository = Repository.find_by_id(params[:repository_id])
+    repository = Repository.find_by_id(params[:repository_id]) || Repository.active.accessible_by_teams(@team).first
```
urbanrotnik, line 35: Maybe we can call here `Repository.active.accessible_by_teams.find_by(id: params[:repository_id]) || Repository.active.accessible_by_teams(@team).first`
And then we do not need `can_read_repository?` permission check

app/helpers/experiments_helper.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module ExperimentsHelper
+  def grouped_by_prj(experiments)
```
urbanrotnik, line 4: `def grouped_by_project`? :)

urbanrotnik (approved):
Two comments, check it out.

PR #2845, opened 2020-09-15
Improve search for smart annotations [SCI-4789]

urbanrotnik:
Jira ticket: [SCI-4789](https://biosistemika.atlassian.net/browse/SCI-4789)

### What was done
Improved SA search

### DIFF
--- a/app/models/concerns/searchable_by_name_model.rb
+++ b/app/models/concerns/searchable_by_name_model.rb
@@ -12,10 +12,23 @@ module SearchableByNameModel
         .limit(Constants::SEARCH_LIMIT)
     end
 
+    def self.search_by_name_all(user, teams = [], query = nil)
+      return if user.blank? || teams.blank?
+
+      sql_q = viewable_by_user(user, teams)
+      # Remove non-breaking space, &nbsp
+      query_array = query.gsub(/[[:space:]]+/, ' ').split(' ')
+      if query_array.any?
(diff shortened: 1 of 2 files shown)


app/models/concerns/searchable_by_name_model.rb
```diff
@@ -12,10 +12,23 @@ def self.search_by_name(user, teams = [], query = nil, options = {})
...
     end
 
+    def self.search_by_name_all(user, teams = [], query = nil)
+      return if user.blank? || teams.blank?
+
+      sql_q = viewable_by_user(user, teams)
+      # Remove non-breaking space, &nbsp
+      query_array = query.gsub(/[[:space:]]+/, ' ').split(' ')
+      if query_array.any?
+        sql_q = sql_q.where("LOWER(#{table_name}.name) like all (array[?])", query_array.map { |c| "%#{c.downcase}%" })
```
okriuchykhin, line 22: ILIKE is case case insensitive

okriuchykhin: Does it use index?

app/models/concerns/searchable_by_name_model.rb
```diff
@@ -12,10 +12,23 @@ def self.search_by_name(user, teams = [], query = nil, options = {})
         .limit(Constants::SEARCH_LIMIT)
     end
 
+    def self.search_by_name_all(user, teams = [], query = nil)
```
okriuchykhin, line 15: Why not simply add extra option to self.search_by_name?

okriuchykhin (approved):

PR #2862, opened 2020-09-22
Add Endpoins for Project Membership manipulation [SCI-5001]

okriuchykhin:
Jira ticket: [SCI-5001](https://biosistemika.atlassian.net/browse/SCI-5001)

### What was done
Add Endpoins for Project Membership manipulation

### DIFF
--- a/app/controllers/api/v1/user_projects_controller.rb
+++ b/app/controllers/api/v1/user_projects_controller.rb
@@ -6,6 +6,7 @@ module Api
       before_action :load_team
       before_action :load_project
       before_action :load_user_project, only: :show
+      before_action :load_user_project_for_managing, only: %i(update destroy)
 
       def index
         user_projects = @project.user_projects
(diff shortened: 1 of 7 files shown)


app/controllers/api/v1/user_projects_controller.rb
```diff
@@ -23,11 +24,44 @@ def show
...
+        @user_project.role = user_project_params[:role]
+        return render body: nil, status: :no_content unless @user_project.changed?
+
+        @user_project.assigned_by = current_user
+        @user_project.save!
+        render jsonapi: @user_project, serializer: UserProjectSerializer, status: :ok
+      end
+
+      def destroy
+        @user_project.destroy!
```
radiokills, line 45: Should we check here if the user is not unassigning himself?

radiokills: Or at least check if he is owner (or the last user on a project).

radiokills (commented):
Straight to the point! Good work. Should we check when removing the user?

app/controllers/api/v1/user_projects_controller.rb
okriuchykhin: Do we need such checks in an API?

okriuchykhin: I think it is better to move such check to the model itself

radiokills: true

radiokills (approved):

PR #2870, opened 2020-09-24
Fix at long inventory name snapshot in title is not visible [SCI-4841]

mlorb:
Jira ticket: [SCI-4841](https://biosistemika.atlassian.net/browse/SCI-4841)

### What was done
Fix at long inventory name snapshot in title is not visible.


okriuchykhin (changes requested):
Please try to make it more responsive

radiokills (approved):

PR #2874, opened 2020-09-28
Put google fonts in public folder

radiokills:


aignatov-bio (changes requested):
Why not add it to assets pipeline?

PR #2895, opened 2020-10-12
Refactor export all my module generator [SCI-5031]

aignatov-bio:
Jira ticket: [SCI-5031](https://biosistemika.atlassian.net/browse/SCI-5031)

### What was done
Refactor export all my module generator

### DIFF
--- a/app/models/report.rb
+++ b/app/models/report.rb
@@ -92,10 +92,7 @@ class Report < ApplicationRecord
   end
 
   def self.generate_whole_project_report(project, current_user, current_team)
-    report_contents = gen_element_content(project, nil, 'project_header', true)
-    Extends::EXPORT_ALL_PROJECT_ELEMENTS.each do |report_element, children|
-      report_contents += public_send("generate_#{report_element}_element", project, children)
-    end
+    report_contents = gen_element_content(project, Extends::EXPORT_ALL_PROJECT_ELEMENTS)
 
     report = Report.new
     report.name = loop do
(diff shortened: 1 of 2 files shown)


okriuchykhin (commented):
Didn't we discuss tree-like structure?

config/initializers/extends.rb
```diff
@@ -41,6 +41,25 @@ class Extends
                            my_module_repository: 17,
                            my_module_protocol: 18 }
 
+  EXPORT_ALL_PROJECT_ELEMENTS = {
```
okriuchykhin, line 44: So it is not possible to have here all needed arguments for 'gen_element_content' method?

aignatov-bio: Elements not consistent. In some cases, we iterate and pass only one object, in other, we pass parent object and associate objects. And special cases for results and assigned repositories.

config/initializers/extends.rb
```diff
@@ -41,6 +41,93 @@ class Extends
                            my_module_repository: 17,
                            my_module_protocol: 18 }
 
+  EXPORT_ALL_PROJECT_ELEMENTS = [
+    {
+      type_of: 'project_header',
+      id_key: 'project_id',
+      children: []
```
okriuchykhin, line 48: Maybe just add  null check to avoid specifying empty elements here?

app/models/report.rb
```diff
@@ -155,31 +107,29 @@ def self.generate_whole_project_report(project, current_user, current_team)
...
-        'sort_order' => sort_order.present? ? sort_order : nil,
-        'children' => children.present? ? children : []
+    children.each do |element|
+      element_hash = lambda { |object|
+        hash_object = {
+          'type_of' => element[:type_of] || element[:type_of_lambda].call(object),
+          'id' => { element[:id_key] => object.id },
+          'sort_order' => element[:sort_order],
+          'children' => gen_element_content(object, element[:children])
+        }
```
okriuchykhin, line 120: element_hash['children'] = gen_element_content(object, element[:children]) if element[:children].present?

okriuchykhin (approved):

PR #2918, opened 2020-10-26
Refactor projects page [SCI-5134]

urbanrotnik:
Jira ticket: [SCI-5134](https://biosistemika.atlassian.net/browse/SCI-5134)

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -226,7 +226,7 @@ class ProjectsController < ApplicationController
   end
 
   def dt_state_load
-    render json: { state: current_team.current_view_state(current_user).state.dig('projects', 'table') }
+    render json: { state: current_team&.current_view_state(current_user)&.state&.dig('projects', 'table') }
   end
 
   private
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -1,63 +1,42 @@
+# frozen_string_literal: true
+
 class ProjectsController < ApplicationController
   include RenamingUtil
   include TeamsHelper
   include InputSanitizeHelper
 
   before_action :switch_team_with_param, only: :index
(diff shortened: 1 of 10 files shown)


app/views/projects/index/modals/_new.html.erb
```diff
@@ -0,0 +1,30 @@
+
```
okriuchykhin, line 1: Empty lines?

app/controllers/projects_controller.rb
```diff
@@ -258,18 +225,20 @@ def experiment_archive
     current_team_switch(@project.team)
   end
 
+  def dt_state_load
+    render json: { state: current_team.current_view_state(current_user).state.dig('projects', 'table') }
```
okriuchykhin, line 229: What if there is no current_team?

urbanrotnik: You are not able to access that screen if you do not have current_team...

urbanrotnik: Orr.. hm, should I return 403 or something?

app/views/projects/index.html.erb
```diff
@@ -1,169 +1,49 @@
-
 <% provide(:head_title, t("projects.index.head_title")) %>
 
 <% if current_team %>
-  <%= render partial: "shared/sidebar", locals: { page: 'project' } %>
+  <%= render partial: "shared/sidebar", locals: {page: 'project'} %>
```
okriuchykhin, line 4: Why did you remove spaces?

urbanrotnik: Auto formater.. ups

app/controllers/projects_controller.rb
okriuchykhin: Maybe just safe navigation?

okriuchykhin (approved):

PR #2928, opened 2020-10-29
Create database structure for project folders [SCI-5126]

okriuchykhin:
Jira ticket: [SCI-5126](https://biosistemika.atlassian.net/browse/SCI-5126)

### What was done
Create database structure for project folders

We decided to start without root folders first

### DIFF
--- a/app/models/project_folder.rb
+++ b/app/models/project_folder.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+class ProjectFolder < ApplicationRecord
+  include SearchableModel
+  include SearchableByNameModel
+
+  validates :name,
+            length: { minimum: Constants::NAME_MIN_LENGTH,
+                      maximum: Constants::NAME_MAX_LENGTH },
+            uniqueness: { scope: :team_id, case_sensitive: false }
+
+  before_validation :inherit_team_from_parent_folder, if: -> { parent_folder.present? }
+
+  belongs_to :team, inverse_of: :project_folders, touch: true
+  belongs_to :parent_folder, class_name: 'ProjectFolder', optional: true
+  has_many :projects, inverse_of: :project_folder, dependent: :nullify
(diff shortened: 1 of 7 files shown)


app/models/project_folder.rb
```diff
@@ -0,0 +1,26 @@
...
+  include SearchableByNameModel
+
+  validates :name,
+            length: { minimum: Constants::NAME_MIN_LENGTH,
+                      maximum: Constants::NAME_MAX_LENGTH },
+            uniqueness: { scope: :team_id, case_sensitive: false }
+
+  before_validation :inherit_team_from_parent_folder, if: -> { parent_folder.present? }
+
+  belongs_to :team, inverse_of: :project_folders, touch: true
```
urbanrotnik, line 14: Why do we want to touch the team on each project_folder change?

db/migrate/20201028103608_create_project_folders.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class CreateProjectFolders < ActiveRecord::Migration[6.0]
+  include DatabaseHelper
+
+  def up
+    create_table :project_folders do |t|
```
urbanrotnik, line 7: I would call the table and model just `folder(s)`. 
If we introduce folders on more levels, probably it will become STI, and a base class can be `folder`, ProjectFolder, TaskFolder, etc could be subclasses.

urbanrotnik (changes requested):
Can you add `factory` and `model specs` classes?
If we do not add them now, we won't add them later.

app/models/project_folder.rb
okriuchykhin: To invalidate caches

db/migrate/20201028103608_create_project_folders.rb
okriuchykhin: It will be only for projects

urbanrotnik (approved):

PR #2935, opened 2020-11-04
Generate backend for moving the projects and folders [SCI-5135]

okriuchykhin:
Jira ticket: [SCI-5135](https://biosistemika.atlassian.net/browse/SCI-5135)

### What was done
Generate backend for moving the projects and folders

### DIFF
--- a/app/controllers/project_folders_controller.rb
+++ b/app/controllers/project_folders_controller.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+class ProjectFoldersController < ApplicationController
+  before_action :check_manage_permissions, only: %i(move_to)
+
+  def move_to
+    destination_folder = current_team.project_folders.find(move_params[:id])
+    destination_folder.transaction do
+      move_projects(destination_folder)
+      move_folders(destination_folder)
+    end
+    respond_to do |format|
+      format.json { render json: { flash: I18n.t('projects.move.success_flash') } }
+    end
+  rescue StandardError => e
+    Rails.logger.error e.message
+    Rails.logger.error e.backtrace.join("\n")
+    respond_to do |format|
+      format.json { render json: { flash: I18n.t('projects.move.error_flash') }, status: :bad_request }
+    end
+  end
+
+  private
+
+  def check_manage_permissions
+    render_403 unless can_update_team?(current_team)
+  end
+
+  def move_params
+    params.require(:id)
+    params.require(:movables)
+    params.permit(:id, movables: %i(type id))
+  end
+
+  def move_projects(destination_folder)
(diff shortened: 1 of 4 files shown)


config/routes.rb
```diff
@@ -197,6 +197,12 @@ def draw(routes_name)
     get 'projects/sidebar', to: 'projects#sidebar', as: 'projects_sidebar'
     get 'projects/dt_state_load', to: 'projects#dt_state_load', as: 'projects_dt_state_load'
 
+    resources :project_folders, only: [] do
```
urbanrotnik, line 200: I think you have to rebase to the current feature branch and merge with existing routes?

urbanrotnik: Or is this route on different level? 🤔

urbanrotnik (commented):
One comment about routes and I would add simple tests for the `move_to` action.

config/routes.rb
okriuchykhin: It's the same level, I will rebase

app/controllers/project_folders_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+    Rails.logger.error e.backtrace.join("\n")
+    respond_to do |format|
+      format.json { render json: { flash: I18n.t('projects.move.error_flash') }, status: :bad_request }
+    end
+  end
+
+  private
+
+  def check_manage_permissions
+    render_403 unless can_update_team?(current_team)
```
mlorb, line 26: From the specifications: "Moving of projects and folders is only allowed to Team administrators (and Org admins)."

spec/controllers/project_folders_controller_spec.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe ProjectFoldersController, type: :controller do
+  login_user
+  render_views
+
+  let!(:user) { User.first }
```
urbanrotnik, line 9: I think explicit `subject.current_user` would be better

spec/controllers/project_folders_controller_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+  let!(:team) { create :team, created_by: user }
+  let!(:user_team) { create :user_team, team: team, user: user, role: :admin }
+
+  let!(:project_folder_1) do
```
urbanrotnik, line 13: Try to avoid using `let!`. 
When we add additional tests, all those objects will be instanced before each test block. Maybe use before block inside post?

urbanrotnik (changes requested):

spec/controllers/project_folders_controller_spec.rb
okriuchykhin: true

urbanrotnik (approved):

PR #2946, opened 2020-11-09
Implement search endpoint for Projects & folders search [SCI-5136]

okriuchykhin:
Jira ticket: [SCI-5136](https://biosistemika.atlassian.net/browse/SCI-5136)

### What was done
Implement search endpoint for Projects & folders search

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -298,6 +298,7 @@ Rails.application.routes.draw do
 
       collection do
         get 'cards'
+        get ':project_folder_id/cards', to: 'projects#cards'
       end
     end
 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -298,8 +298,11 @@ Rails.application.routes.draw do
 
       collection do
         get 'cards'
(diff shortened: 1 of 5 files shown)


config/routes.rb
```diff
@@ -298,6 +298,7 @@ def draw(routes_name)
 
       collection do
         get 'cards'
+        get ':project_folder_id/cards', to: 'projects#cards'
```
urbanrotnik, line 301: the route will be like: 
`projects/1/cards` what reads like cards of project 1... 
I would add `folders` before `:project_folder_id` like `projects/folders/1/cards`

urbanrotnik (changes requested):
Not sure about the scope of this: 
- Is search input part of this ticket? 
- I talked to Sasa, sorting should work above Folders and Projects. So cards should be mixed and sorted together.

urbanrotnik:
We have to discuss this on a standup meeting, looks like this is a bit outdated task...

urbanrotnik (approved):

PR #2948, opened 2020-11-10
Add project folders to global search [SCI-5180]

okriuchykhin:
Jira ticket: [SCI-5180](https://biosistemika.atlassian.net/browse/SCI-5180)

### What was done
Add project folders to global search


urbanrotnik (changes requested):
Looks good. The only thing, you have to change the icon for Projects in search results. Should be a folder icon with "P". The same as is on the global activities page.

urbanrotnik (approved):

PR #2963, opened 2020-11-18
Add comments sidebar layout [SCI-5186][SCI-5188][SCI-5189]

aignatov-bio:
Jira ticket: [SCI-5186](https://biosistemika.atlassian.net/browse/SCI-5186)
Jira ticket: [SCI-5188](https://biosistemika.atlassian.net/browse/SCI-5188)
Jira ticket: [SCI-5189](https://biosistemika.atlassian.net/browse/SCI-5189)

### What was done
Add comments sidebar layout


app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,106 @@
...
+
+    @commentable = @comment.commentable
+    render_404 and return unless @commentable
+  end
+
+  def comment_params
+    params.permit(:message)
+  end
+
+  def check_view_permissions
```
urbanrotnik, line 73: Should we create comments permissions file and move these checks there?

app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,106 @@
...
+    when Step
+      render_403 and return unless can_read_protocol_in_module?(@commentable.protocol)
+    when Result
+      render_403 and return unless can_read_experiment?(@commentable.my_module.experiment)
+    else
+      render_403 and return
+    end
+  end
+
+  def check_create_permissions
```
urbanrotnik, line 88: Same here

app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,106 @@
...
+    }
+  end
+
+  def create
+    @comment = @commentable.comments.new(
+      message: params[:message],
+      user: current_user,
+      associated_id: @commentable.id
+    )
+    comment_create_helper(@comment, 'comment')
```
urbanrotnik, line 31: This structure is a bit weird for me, without response, render function.  
I would prefer to see here a service call that creates a comment, return comment, and explicit render here based on the result. 

That service could be reused later in API controllers.

app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,106 @@
...
+  end
+
+  def update
+    old_text = @comment.message
+    @comment.message = params[:message]
+    comment_update_helper(@comment, old_text, 'comment')
```
urbanrotnik, line 37: ditto

app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,106 @@
...
+  end
+
+  def destroy
+    comment_destroy_helper(@comment)
```
urbanrotnik, line 41: ditto

app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,106 @@
...
+  end
+
+  private
+
+  def load_object
+    @commentable = case params[:object_type]
```
urbanrotnik, line 47: This can be also a part of the service...

app/helpers/comment_helper.rb
```diff
@@ -96,18 +107,150 @@ def comment_destroy_helper(comment)
...
+        log_my_module_activity(:delete_module_comment, comment)
       end
       render json: {}, status: :ok
     else
       render json: { message: I18n.t('comments.delete_error') },
                  status: :unprocessable_entity
     end
   end
+
+  def result_comment_annotation_notification(comment, old_text = nil)
```
urbanrotnik, line 125: Hmm... 
Can we abstract all these notifications in one CommentNotificationService? 
Or/And implement the `send_notification` method on each Comment SubClass? 

Also, those send notifications can be executed asynchronous...

urbanrotnik (commented):
Great Anton, I think this is a really good idea... 🎉 
Few ideas about moving code, check the comments below...
And maybe we should cover all ruby code with basic unit tests.

urbanrotnik (approved):
Refactor will be made after we remove the old comments code, controllers

PR #2970, opened 2020-11-23
Folders API endpoints [SCI-5185]

urbanrotnik:
Jira ticket: [SCI-5185](https://biosistemika.atlassian.net/browse/SCI-5185)

### What was done
Added CRU endpoints for Folders
Extend Projects endpoints

### Note
We are following JSON API convention, so the folder is `relation` for Projects, which means it should be passed within the `relations` field. 
It becomes a bit nasty to parse JSON API params hash and convert them to rails params. 
I added light `gem` for JSON API parameters. It convert nested fields like Relationships and Includes to rails params. 
We can dry up also existing controllers.


app/serializers/api/v1/project_folder_serializer.rb
```diff
@@ -0,0 +1,13 @@
...
+
+module Api
+  module V1
+    class ProjectFolderSerializer < ActiveModel::Serializer
+      type :project_folders
+      attributes :id, :name
+
+      belongs_to :team, serializer: TeamSerializer
+      belongs_to :parent_folder, serializer: ProjectFolderSerializer
+    end
```
okriuchykhin, line 11: shouldn't it also include projects inside the folder?

app/controllers/api/v1/projects_controller.rb
```diff
@@ -61,13 +62,17 @@ def activities
       def project_params
         raise TypeError unless params.require(:data).require(:type) == 'projects'
 
-        params.require(:data).require(:attributes).permit(:name, :visibility, :archived)
+        params.from_jsonapi.require(:project).permit(:name, :visibility, :archived, :project_folder_id)
```
okriuchykhin, line 65: why do we need it here?

app/controllers/api/v1/projects_controller.rb
```diff
@@ -61,13 +62,17 @@ def activities
...
       end
 
       def load_project_for_managing
         @project = @team.projects.find(params.require(:id))
         raise PermissionError.new(Project, :manage) unless can_manage_project?(@project)
       end
+
+      def check_parent_folder
+        @team.project_folders.find(project_params[:project_folder_id]) if project_params[:project_folder_id]
```
okriuchykhin, line 74: Shouldn't it be done by model validator?

app/controllers/api/v1/project_folders_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+        params.from_jsonapi.require(:project_folder).permit(:name, :parent_folder_id)
+      end
+
+      def update_folder_params
+        raise IDMismatchError unless params.require(:data).require(:id).to_i == params[:id].to_i
+
+        folder_params
+      end
+
+      def check_parent_folder
```
okriuchykhin, line 56: Shouldn't it be done by model validator?

app/controllers/api/v1/project_folders_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+      private
+
+      def load_project_folder
+        @project_folder = @team.project_folders.find(params.require(:id))
+      end
+
+      def folder_params
+        raise TypeError unless params.require(:data).require(:type) == 'project_folders'
+
+        params.from_jsonapi.require(:project_folder).permit(:name, :parent_folder_id)
```
okriuchykhin, line 47: why do we need from_jsonapi here?

app/controllers/api/v1/project_folders_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+        folder = @team.project_folders.create!(folder_params)
+
+        render jsonapi: folder, serializer: ProjectFolderSerializer, status: :created
+      end
+
+      def update
+        @project_folder.attributes = update_folder_params
+
+        if @project_folder.changed? && @project_folder.save!
+          render jsonapi: @project_folder, serializer: ProjectFolderSerializer, status: :ok
```
okriuchykhin, line 32: :ok is default status

app/serializers/api/v1/project_folder_serializer.rb
urbanrotnik: probably yes...

app/controllers/api/v1/projects_controller.rb
urbanrotnik: As I mentioned in a meeting last week, `from_jsonapi` translates the raw json:api parameters into “classic” rails parameters, with a model top-level key and attributes in it’s object. The classic rails parameters are more conventional and easier to work with. As an added bonus, it handles relationships and parses them as `#{relationship_name}_id` or `#{relationship_name}_ids` for has_many relationships.

app/controllers/api/v1/project_folders_controller.rb
urbanrotnik: As I mentioned in a meeting last week, `from_jsonapi` translates the raw json:api parameters into “classic” rails parameters, with a model top-level key and attributes in it’s object. The classic rails parameters are more conventional and easier to work with. As an added bonus, it handles relationships and parses them as `#{relationship_name}_id` or `#{relationship_name}_ids` for has_many relationships.

app/controllers/api/v1/projects_controller.rb
urbanrotnik: I guess yes

app/controllers/api/v1/project_folders_controller.rb
urbanrotnik: I guess yes

app/controllers/api/v1/projects_controller.rb
okriuchykhin: We do not create relationships if specific controller exists for that type

okriuchykhin: def jsonapi_params(type)
  raise TypeError unless params.require(:data).require(:type) == type

  params.require(:data).require(:attributes)
end

urbanrotnik: How would you save the `belongs_to` associations then? 
If you check `json:api` [docs](https://jsonapi.org/format/#document-resource-object-attributes) says:
> Although has-one foreign keys (e.g. author_id) are often stored internally alongside other information to be represented in a resource object, these keys SHOULD NOT appear as attributes.

So we SHOULD NOT use the belongs_to fields in the request `attributes` object and should use the `relationships` object instead. 
It means we need a custom method in each controller, where we are assigning belongs_to attributes.
To translate this
```
{
    "data": {
        "type": "project_folders",
        "id": 1,
        "attributes": {
            
        },
        "relationships": {
            "parent_folder": {
                "data": { 
                    "type": "project_folders",
                    "id": 3
                }
            }
        }
    }
}
```

to this
```
{ project_folder: { name: 'MyUpdatedFolder', parent_folder_id: 2}
``` 

Only for this parent folder, we need to translate parameters with something like: 
```
params.permit(data: { attributes: %i(name) })[:data][:attributes]
          .merge(parent_folder_id:
                   params.permit(data: { relationships: {parent_folder: {data: %i(id)}}})['data']['relationships']['parent_folder']['data']['id'])

```
And a gem is doing exactly that. 

And I am not sure what is considered a trusted source? 
just a few facts about this gem: 
- MIT license
- 100% test coverage
- 34k downloads
- Last commit 20days ago..

app/models/project.rb
```diff
@@ -321,4 +322,12 @@ def generate_teams_export_report_html(
   ensure
     report.destroy if report.present?
   end
+
+  private
+
+  def project_folder_teams
+    return if project_folder.team_id == team_id
+
+    errors.add(:project_folder, 'Project folder and project should belongs to the same team')
```
okriuchykhin, line 331: I18n?

app/models/project_folder.rb
```diff
@@ -87,4 +88,10 @@ def inner_projects
   def inherit_team_from_parent_folder
     self.team = parent_folder.team
   end
+
+  def team_scope
+    return if parent_folder.team_id == team_id
+
+    errors.add(:parent_folder, 'Parent folder and folder should belongs to the same team')
```
okriuchykhin, line 95: I18n?

okriuchykhin (changes requested):
Looks good, just use I18n for error messages

app/models/project.rb
urbanrotnik: sure

okriuchykhin (approved):

PR #2983, opened 2020-11-26
Remove manage permissions check from step toggle [SCI-5252]

urbanrotnik:
Jira ticket: [SCI-5252](https://biosistemika.atlassian.net/browse/SCI-5252)

### What was done
Remove manage permission check


mlorb (approved):
Nice :+1:

PR #2987, opened 2020-11-30
Projects cards and list view with filters  [SCI-5208]

urbanrotnik:
Jira ticket: [SCI-5208](https://biosistemika.atlassian.net/browse/SCI-5208)

### What was done
Adapt filter results to cards and list view

### DIFF
--- a/app/assets/stylesheets/projects.scss
+++ b/app/assets/stylesheets/projects.scss
@@ -667,7 +667,7 @@ li.module-hover {
   .cards-wrapper {
     align-items: center;
     display: grid;
-    grid-auto-rows: 11em;
+    grid-auto-rows: 2em;
     grid-column-gap: 1em;
     grid-row-gap: 1em;
     grid-template-columns: repeat(auto-fit, minmax(270px, 1fr));
@@ -675,6 +675,11 @@ li.module-hover {
 
     .table-header {
       display: none;
+    }
+
+    .projects-group {
+      grid-column: 1/-1;
+      margin: 0;
     }
 
     .card {
@@ -686,6 +691,7 @@ li.module-hover {
       padding: .5em 1em;
       position: relative;
       width: 100%;
+      grid-row: span 4;
 
       .checkbox-cell {
         align-items: center;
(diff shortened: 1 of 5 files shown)


app/assets/stylesheets/projects.scss
```diff
@@ -803,6 +809,24 @@ li.module-hover {
       grid-template-columns: max-content 1fr 1fr 1fr 1fr max-content;
       margin: 40px 0 0 6px;
 
+
+      .projects-group {
+        //border-bottom: 1px solid $color-concrete;
```
okriuchykhin, line 814: Is it left here intensionally?

app/views/projects/index/_breadcrumbs.html.erb
```diff
@@ -4,7 +4,7 @@
               class: 'project-folder-link',
               data: { projects_cards_url: cards_projects_url }) %>
   <span>/</span>
-  <% @current_folder.parent_folders.each do |project_folder| %>
+  <% target_folder&.parent_folders&.each do |project_folder| %>
```
okriuchykhin, line 7: You can write <% (@current_folder || target_folder).parent_folders.each do |project_folder| %>

urbanrotnik: We have to keep safe navigation. When you are rendering breadcrumbs for the root folder, your target_folder/current_folder is nil

okriuchykhin: Like this?
<% (@current_folder || target_folder)&.parent_folders&.each do |project_folder| %>

urbanrotnik: Yes, but whenever the `@current_folder` exists, `target_folder` is there. I mean, I explicitly passing taget_fodler as a `locale` on all places where I use partial.

okriuchykhin: <% (local_assigns[:target_folder] || @current_folder)&.parent_folders&.each do |project_folder| %>

okriuchykhin (approved):

PR #2996, opened 2020-12-07
Implement the 'seen comments' functionality (red dot) for the comments [SCI-5253]

okriuchykhin:
Jira ticket: [SCI-5253](https://biosistemika.atlassian.net/browse/SCI-5253)

### What was done
Implement the 'seen comments' functionality (red dot) for the comments

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -60,4 +60,10 @@ class Comment < ApplicationRecord
         .offset((page - 1) * Constants::SEARCH_LIMIT)
     end
   end
+
+  def self.mark_as_seen_by(user)
+    # rubocop:disable Rails/SkipsModelValidations
+    all.where('? = ANY (unseen_by)', user.id).update_all("unseen_by = array_remove(unseen_by, #{user.id.to_i}::bigint)")
+    # rubocop:enable Rails/SkipsModelValidations
+  end
 end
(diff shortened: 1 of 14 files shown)


app/models/comment.rb
```diff
@@ -60,4 +60,10 @@ def self.search(
         .offset((page - 1) * Constants::SEARCH_LIMIT)
     end
   end
+
+  def self.mark_as_seen_by(user)
```
radiokills, line 64: Is there any way to prevent calling this on the model itself and limit it only to the collection?
We would really like to avoid somebody calling
```Comment.mark_as_seen_by(current_user)```

Comment (okriuchykhin) [reply app/models/comment.rb:64]: That 'somebody' can also call Comment.delete_all

File: app/views/shared/comments/_comments.html.erb [hunk 2]
```diff
@@ -12,15 +12,16 @@
     <% end %>
     <div class="comments-list">
       <%= render partial: 'shared/comments/list.html.erb', locals: { comments: comments} %>
+      <% comments.mark_as_seen_by(current_user) %>
```
radiokills, line 15: Isn't this a bit too much logic for the view?

okriuchykhin: It is directly connected to the view. Comments should be marked only when they are rendering and these changes affect only representation.

PR #3009, opened 2020-12-12
Archiving: update archived TASK screen layout [SCI-5276]

okriuchykhin:
Jira ticket: [SCI-5276](https://biosistemika.atlassian.net/browse/SCI-5276)

### What was done
Archiving: update archived TASK screen layout

### DIFF
--- a/app/helpers/my_modules_helper.rb
+++ b/app/helpers/my_modules_helper.rb
@@ -31,7 +31,7 @@ module MyModulesHelper
 
   def get_task_alert_color(my_module)
     alert = ''
-    unless my_module.completed?
+    if my_module.active? && !my_module.completed?
       alert = ' alert-yellow' if my_module.is_one_day_prior?
       alert = ' alert-red' if my_module.is_overdue?
     end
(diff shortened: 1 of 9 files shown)


app/helpers/my_modules_helper.rb
```diff
@@ -31,7 +31,7 @@ def az_ordered_assets_index(step, asset_id)
 
   def get_task_alert_color(my_module)
     alert = ''
-    unless my_module.completed?
+    unless my_module.archived? || my_module.completed?
```
radiokills, line 34: Isn't there an unwritten rule that unless should be done only when you have one condition, otherwise 'if' should be used?

app/permissions/my_module.rb
```diff
@@ -7,7 +7,8 @@
      assign_sample_to_module
      create_comments_in_module
      create_my_module_repository_snapshot
-     manage_my_module_repository_snapshots)
+     manage_my_module_repository_snapshots
+     change_my_module_flow_status)
```
radiokills, line 11: nice catch

app/helpers/my_modules_helper.rb
okriuchykhin: Linter doesn't say anything about it

radiokills: As said, it's a good practice, changing it with if makes it more readable. Change it please.

PR #3088, opened 2021-01-15
Fix breadcrumbs order for folders [SCI-5395]

urbanrotnik:
Jira ticket: [SCI-5395](https://biosistemika.atlassian.net/browse/SCI-5395)

### DIFF
--- a/app/helpers/project_folders_helper.rb
+++ b/app/helpers/project_folders_helper.rb
@@ -10,4 +10,19 @@ module ProjectFoldersHelper
       I18n.t('projects.index.modal_move_folder.items.projects_and_folders')
     end
   end
+
+  def tree_ordered_parent_folders(current_folder)
+    folders = current_folder&.parent_folders
+    return unless folders
+
+    ordered_folders = [current_folder]
+    folders.each do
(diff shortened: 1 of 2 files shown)


app/models/project_folder.rb
```diff
@@ -83,7 +83,15 @@ def parent_folders
       )
       SELECT id FROM outer_project_folders ORDER BY selected_folders_ids'.gsub(/\n|\t/, ' ').gsub(/ +/, ' ')
 
-    ProjectFolder.where("project_folders.id IN (#{outer_folders_sql})", team.id, id)
+    folders = ProjectFolder.where("project_folders.id IN (#{outer_folders_sql})", team.id, id)
```
okriuchykhin, line 86: Move this special sorting to the helper as this method is used in other places

app/models/project_folder.rb
```diff
@@ -83,7 +83,15 @@ def parent_folders
...
+
+    ordered_folders = [folders.find { |f| f.parent_folder_id.nil? }]
```
okriuchykhin, line 88: isn't it easier to start from current folder and go up towards the root?

okriuchykhin (approved):

PR #3100, opened 2021-01-19
Fix navigation between archived and active projects [SCI-5419]

okriuchykhin:
Jira ticket: [SCI-5419](https://biosistemika.atlassian.net/browse/SCI-5419)

### What was done
Fix navigation between archived and active projects

### DIFF
--- a/app/helpers/projects_helper.rb
+++ b/app/helpers/projects_helper.rb
@@ -2,10 +2,14 @@
 
 module ProjectsHelper
   def projects_view_mode
+    return 'archived' if current_folder&.archived?
+
     params[:view_mode] == 'archived' ? 'archived' : 'active'
   end
 
   def projects_view_mode_archived?
+    return true if current_folder&.archived?
+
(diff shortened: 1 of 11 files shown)


app/helpers/projects_helper.rb
```diff
@@ -2,10 +2,14 @@
 
 module ProjectsHelper
   def projects_view_mode
+    return 'archived' if current_folder&.archived?
+
     params[:view_mode] == 'archived' ? 'archived' : 'active'
   end
 
   def projects_view_mode_archived?
+    return true if current_folder&.archived?
```
urbanrotnik, line 11: Would this work? 
```
def projects_view_mode_archived?
   projects_view_mode == 'archived'
end
```

urbanrotnik (approved):
One comment, you can reduce a line of code :)

app/helpers/projects_helper.rb
okriuchykhin: True

PR #3148, opened 2021-02-11
Store Table/Cards view setting per user [SCI-5450]

urbanrotnik:
Jira ticket: [SCI-5450](https://biosistemika.atlassian.net/browse/SCI-5450)

### What was done
Remember user setting for view_type (cards/table)

### DIFF
--- a/app/helpers/cards_view_helper.rb
+++ b/app/helpers/cards_view_helper.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+module CardsViewHelper
+  def cards_view_type_class(view_type)
+    view_type == 'table' ? 'list' : ''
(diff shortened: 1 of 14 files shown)


app/helpers/cards_view_helper.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module CardsViewHelper
+  def view_type_class(view_type)
```
okriuchykhin, line 4: May be we need more unique name?

urbanrotnik: @okriuchykhin 
Updated

aignatov-bio (approved):

PR #3151, opened 2021-02-11
Fix Move button on Experiment card [SCI-5475]

urbanrotnik:
Jira ticket: [SCI-5475](https://biosistemika.atlassian.net/browse/SCI-5475)

### DIFF
--- a/app/views/projects/show.html.erb
+++ b/app/views/projects/show.html.erb
@@ -32,6 +32,5 @@
   </div>
 </div>
 
-<%= javascript_include_tag("experiments/dropdown_actions") %>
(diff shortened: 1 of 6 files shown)


app/views/projects/show.html.erb
```diff
@@ -32,6 +32,5 @@
   </div>
 </div>
 
-<%= javascript_include_tag("experiments/dropdown_actions") %>
```
urbanrotnik, line 35: @aignatov-bio 
I removed this again. This logic should not be used here at all. 
I updated button for new project (changed it to form)

aignatov-bio (approved):

PR #3188, opened 2021-02-26
Fix sorting on Teams settings page [SCI-4179]

urbanrotnik:
Jira ticket: [SCI-4179](https://biosistemika.atlassian.net/browse/SCI-4179)

### What was done
Fixed.

### DIFF
--- a/app/datatables/team_users_datatable.rb
+++ b/app/datatables/team_users_datatable.rb
@@ -16,9 +16,9 @@ class TeamUsersDatatable < CustomDatatable
     @sortable_columns ||= [
       'User.full_name',
       'User.email',
-      'UserTeam.created_at',
+      'UserTeam.role',
       'User.confirmed_at',
-      'UserTeam.role'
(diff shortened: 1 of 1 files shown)


app/datatables/team_users_datatable.rb
```diff
@@ -16,9 +16,9 @@ def sortable_columns
     @sortable_columns ||= [
       'User.full_name',
       'User.email',
-      'UserTeam.created_at',
+      'UserTeam.role',
+      'User.confirmed_at',
```
aignatov-bio, line 20: Why same values here?

aignatov-bio (changes requested):

app/datatables/team_users_datatable.rb
urbanrotnik: Because last column is "status" and it is actually the value of `confirmed_at.present?`

aignatov-bio (approved):

PR #3192, opened 2021-03-01
Add migration for new user assignments to projects [SCI-5478]

okriuchykhin:
Jira ticket: [SCI-5478](https://biosistemika.atlassian.net/browse/SCI-5478)

### What was done
Add migration for new user assignments to projects


app/permissions/my_module.rb
```diff
@@ -23,57 +23,56 @@
...
   can :restore_module do |user, my_module|
-    user.is_user_or_higher_of_project?(my_module.experiment.project) &&
-      my_module.archived?
+    my_module.archived? && my_module.permission_granted?(user, MyModulePermissions::RESTORE)
   end
 
   # module: move
   can :move_module do |user, my_module|
-    can_manage_experiment?(user, my_module.experiment)
+    my_module.permission_granted?(user, MyModulePermissions::MOVE)
```
ZmagoD, line 43: @okriuchykhin maybe you could use the each loop do define this rules  so when you extend the constants you get the new `can_..` method out of the box. Something like this:

```suggestion
    MyModulePermissions.constants.each do |const|
       can "#{const. downcase}_module" do |user, my_module|
         my_module.permission_granted?(user, "MyModulePermissions::#{const}".constantize)
       end
    end
```

okriuchykhin: Some permission methods contain not only check with permission_granted?, but also other logic

PR #3198, opened 2021-03-02
Improve behavior of the scheduler, add caching to system notifications sync service [SCI-5526]

okriuchykhin:
Jira ticket: [SCI-5526](https://biosistemika.atlassian.net/browse/SCI-5526)

### What was done
Improve behavior of the scheduler, add caching to system notifications sync service

### DIFF
--- a/app/models/system_notification.rb
+++ b/app/models/system_notification.rb
@@ -13,7 +13,7 @@ class SystemNotification < ApplicationRecord
                                            }
   #  ignoring: :accents
 
-  has_many :user_system_notifications
+  has_many :user_system_notifications, dependent: :destroy
   has_many :users, through: :user_system_notifications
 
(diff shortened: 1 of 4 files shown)


app/models/system_notification.rb
```diff
@@ -13,7 +13,7 @@ class SystemNotification < ApplicationRecord
                                            }
   #  ignoring: :accents
 
-  has_many :user_system_notifications
+  has_many :user_system_notifications, dependent: :destroy
```
urbanrotnik, line 16: Hm... Not sure if we have a case where we delete SystemNotification, but if we do and this is on Free Scinote, it will call destroy for each `user_system_notification` record and this can take quite a long time. 
Should we be concerned?

radiokills (approved):

PR #3220, opened 2021-03-18
Fix infinite scroll in current tasks on dashboard [SCI-5574]

aignatov-bio:
Jira ticket: [SCI-5574]](https://biosistemika.atlassian.net/browse/SCI-5574])

### What was done
Fix infinite scroll in current tasks on dashboard

### DIFF
--- a/app/assets/javascripts/sitewide/infinite_scroll.js
+++ b/app/assets/javascripts/sitewide/infinite_scroll.js
@@ -47,8 +47,6 @@ var InfiniteScroll = (function() {
 
     if (config.loadFirstPage) {
       loadData($container, 1);
-    } else if (scrollNotVisible($container)) {
-      loadData($container, $container.data('next-page'));
     }
 
     $container.on('scroll', () => {


app/assets/javascripts/sitewide/infinite_scroll.js
```diff
@@ -47,8 +47,6 @@ var InfiniteScroll = (function() {
 
     if (config.loadFirstPage) {
       loadData($container, 1);
-    } else if (scrollNotVisible($container)) {
```
okriuchykhin, line 50: We don't need to increase default page size?

aignatov-bio: I found, that this part generates a second request. First request correctly loaded elements.

okriuchykhin: Yes, but is size of the first load enough to fill widget for all screen sizes?

aignatov-bio: Yes, max-height for container 400px only 8 row can fit. We loaded 20 rows.

okriuchykhin (approved):

PR #3253, opened 2021-04-10
SCI-5635  Report not generated in case of tasks from more experiments

radiokills:
Jira ticket: [SCI-5635](https://biosistemika.atlassian.net/browse/SCI-5635)

### What was done
This has not much to do with the ticket reported since the cause was hiding elsewhere. 

What was done:
- fix the step API endpoint so it sets last_modified_by_id
- prepared a migration to fix old records 
- minor (not urgent, but still needed) refactor of step model
-

### DIFF
--- a/db/migrate/20210410100006_fix_last_changed_by_on_steps.rb
+++ b/db/migrate/20210410100006_fix_last_changed_by_on_steps.rb
@@ -15,7 +15,7 @@ class FixLastChangedByOnSteps < ActiveRecord::Migration[6.1]
   # because we're not recording the changes to last_modified_by_id
 
   def up
-    Step.where(last_modified_by_id: nil).each do |step|
+    Step.where(last_modified_by_id: nil).find_each do |step|
       # Try to find corresponding activity, if step was changed activity should exist
       # we cannot completely rely on that, so for the fallback we use the step user
       activity = Activity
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -58,7 +58,7 @@ rails-production:
 run:
 	rm -f tmp/pids/server.pid
 	@docker-compose up -d
-	@docker attach scinote_web_development
+	@docker attach scinote_web_testrunner
```
okriuchykhin, line 61: Why it is renamed?

radiokills: wow, this shouldn't be here

db/migrate/20210410100006_fix_last_changed_by_on_steps.rb
```diff
@@ -0,0 +1,33 @@
...
+  # There was an issue with API steps endpoint not settings last_modified_by_id
+  # this migration fix the affected records
+  # At the moment it seems only the complete and uncomplete actions were problematic
+  #
+  # The enterprise build of SciNote also uses the Audit Trails
+  # In this instance we don't neet to worry about these changes
+  # because we're not recording the changes to last_modified_by_id
+
+  def up
+    Step.where(last_modified_by_id: nil).each do |step|
```
okriuchykhin, line 18: use find_each

okriuchykhin (approved):

PR #3257, opened 2021-04-14
SCI-5623 add external id to user

radiokills:
Jira ticket: [SCI-5623](https://biosistemika.atlassian.net/browse/SCI-5623)

### What was done
Added external id as store accessor, added views to profile


app/views/users/registrations/edit_partials/_external_id.erb
```diff
@@ -0,0 +1,12 @@
+<span class="settings-external-id user-settings-block">
```
okriuchykhin, line 1: _external_id.html.erb

okriuchykhin (approved):

PR #3264, opened 2021-04-19
Add styling for template values [SCI-5630]

aignatov-bio:
Jira ticket: [SCI-5630](https://biosistemika.atlassian.net/browse/SCI-5630)

### What was done
Add styling for template values

### DIFF
--- a/app/assets/javascripts/reports/new.js
+++ b/app/assets/javascripts/reports/new.js
@@ -1159,7 +1159,7 @@ function reportHandsonTableConverter() {
       }
     }
 
-    $('.reports-new').on('change', '.report-experiment-checkbox', function() {
+    $('.project-contents-container').on('change', '.report-experiment-checkbox', function() {
       $(this).closest('li').find('.report-my-module-checkbox').prop('checked', this.checked);
       selectAllState();
       hideUnchekedElements($('.hide-unchecked-checkbox').prop('checked'));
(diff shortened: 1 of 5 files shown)


app/components/reports/team_member_input_component.html.erb
```diff
@@ -1,7 +1,7 @@
 <% if @editing %>
```
okriuchykhin, line 1: It was moved to ProjectMembersInputComponent

app/assets/javascripts/reports/new.js
```diff
@@ -1184,6 +1184,15 @@ function reportHandsonTableConverter() {
         let template = $('#templateSelector').val();
         $.get($('#templateSelector').data('valuesEditorPath'), { template: template }, function(result) {
           $('.report-template-values-container').html(result.html);
+
+          $.each($('.report-template-values-container .values-container select'), function(i, select) {
```
okriuchykhin, line 1188: I think it is already initialized

aignatov-bio: It is for dropdowns inside template values (user selection)

aignatov-bio: Ah i see. I will rebase.

okriuchykhin (approved):

PR #3266, opened 2021-04-20
Add warning modal to report edit mode [SCI-5621]

aignatov-bio:
Jira ticket: [SCI-5621](https://biosistemika.atlassian.net/browse/SCI-5621)

### What was done
Add warning modal to report edit mode

### DIFF
--- a/app/assets/javascripts/reports/new.js
+++ b/app/assets/javascripts/reports/new.js
@@ -1251,6 +1251,8 @@ function reportHandsonTableConverter() {
       });
   }
 
+  $('#reportWizardEditWarning').modal('show');
+
   initGenerateButton();
   initReportWizard();
   initDropdowns();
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/reports/new.js
```diff
@@ -1251,6 +1251,8 @@ function reportHandsonTableConverter() {
       });
   }
 
+  $('#reportWizardEditWarning').modal('show');
```
okriuchykhin, line 1254: What if not in edit mode?

app/views/reports/new.html.erb
```diff
@@ -87,4 +87,26 @@
   </div>
 </div>
 
+<% if @edit %>
```
okriuchykhin, line 90: better to move it to separate partial if you want to use it inside such condition

app/assets/javascripts/reports/new.js
aignatov-bio: Modal exist only in edit mode. In new this modal not exist.

okriuchykhin (approved):

PR #3276, opened 2021-04-30
Fix report processing state after create and update [SCI-5678]

aignatov-bio:
Jira ticket: [SCI-5678](https://biosistemika.atlassian.net/browse/SCI-5678)

### What was done
Fix report processing state after create and update

### DIFF
--- a/app/assets/javascripts/reports/reports_datatable.js
+++ b/app/assets/javascripts/reports/reports_datatable.js
@@ -6,6 +6,7 @@
   const RETRY_COUNT = 25;
   const START_POLLING_INTERVAL = 10000;
   var CHECKED_REPORTS = [];
+  var REPORTS_TABLE;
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/reports/reports_datatable.js
```diff
@@ -6,6 +6,7 @@
   const RETRY_COUNT = 25;
   const START_POLLING_INTERVAL = 10000;
   var CHECKED_REPORTS = [];
+  var REPORTS_TABLE;
```
radiokills, line 9: I'm not sure what is the correct naming convention for object variables, but using capitals for this doesn't seem right.

radiokills (commented):
codstyle

okriuchykhin (approved):

PR #3327, opened 2021-05-16
Small fixes in reports templates [SCI-5723]

aignatov-bio:
Jira ticket: [SCI-5723](https://biosistemika.atlassian.net/browse/SCI-5723)

### What was done
Small fixes in reports templates

### DIFF
--- a/app/controllers/reports_controller.rb
+++ b/app/controllers/reports_controller.rb
@@ -56,8 +56,11 @@ class ReportsController < ApplicationController
   end
 
   def new_template_values
-    template = params[:template]
-    return render_404 if template.blank?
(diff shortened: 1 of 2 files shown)


app/controllers/reports_controller.rb
```diff
@@ -56,7 +56,7 @@ def new
   end
 
   def new_template_values
-    template = Extends::REPORT_TEMPLATES[params[:template].to_sym]
+    template = params[:template]
```
okriuchykhin, line 59: Do you want to allow directly inject params to the view load path?

okriuchykhin (approved):

PR #3378, opened 2021-06-10
Remove old reports code [SCI-5650]

okriuchykhin:
Jira ticket: [SCI-5650](https://biosistemika.atlassian.net/browse/SCI-5650)

### What was done
Remove old reports code


app/views/reports/elements/_my_module_repository_element.html.erb
```diff
@@ -1,17 +1,8 @@
-<% my_module ||= @my_module %>
-<% element_id ||= repository&.id %>
+<% repository ||= report_element.repository %>
```
aignatov-bio, line 1: For deleted repositories, `report_element` will be `nil`. We pass only `repository_snapshot`

aignatov-bio (changes requested):
Can you rebase and test how it will work with snapshots from deleted repositories,

app/views/reports/elements/_my_module_repository_element.html.erb
okriuchykhin: What? Why would you even create report element for deleted inventory?

aignatov-bio: I mean if my_module has a snapshot from a deleted repository. Repository itself not exist. But we pass snapshot.

okriuchykhin: Report element should have connection to the repository, even if it is a snapshot

aignatov-bio: Then we need to introduce a new report element - repoistory_snapshot. And create it for each my_module.

app/helpers/reports_helper.rb
```diff
@@ -3,90 +3,24 @@
...
-        children_html.safe_concat render_new_element(true)
-        element.children.active.each do |child|
-          children_html
-            .safe_concat render_report_element(child, provided_locals)
-        end
-      else
-        children_html.safe_concat render_new_element(false)
+    if element.children.active.present?
+      element.children.active.each do |child|
+        children_html.safe_concat render_report_element(child, provided_locals)
```
artoscinote, line 17: `element.children.active&.each do |child|`
would do the same without the if block

artoscinote (approved):

PR #3382, opened 2021-06-11
Fix empty elements in docx rtf table [SCI-5659]

aignatov-bio:
Jira ticket: [SCI-5659](https://biosistemika.atlassian.net/browse/SCI-5659)

### What was done
Fix empty elements in docx rtf table

### DIFF
--- a/app/services/reports/html_to_word_converter.rb
+++ b/app/services/reports/html_to_word_converter.rb
@@ -50,6 +50,8 @@ module Reports
       elements = []
       temp_p = []
       raw_elements.each do |elem|
+        next unless elem
+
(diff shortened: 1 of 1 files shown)


app/services/reports/html_to_word_converter.rb
```diff
@@ -50,6 +50,8 @@ def combine_docx_elements(raw_elements)
       elements = []
       temp_p = []
       raw_elements.each do |elem|
+        next unless elem
```
artoscinote, line 53: We could also use `raw_elements.compact.each do |elem|`.

artoscinote (approved):

PR #3385, opened 2021-06-14
Reload report table after triggering doc generation request [SCI-5788]

artoscinote:
Jira ticket: [SCI-5788](https://biosistemika.atlassian.net/browse/SCI-5788)

### What was done
Added missing class to status cell, so DOCX report status gets refreshed properly after error.


aignatov-bio (changes requested):
Can you make it without table reload?

aignatov-bio (approved):

PR #3387, opened 2021-06-14
Add archived and deleted repositories to reports [SCI-5781]

aignatov-bio:
Jira ticket: [SCI-5781](https://biosistemika.atlassian.net/browse/SCI-5781)

### What was done
Add archived and deleted repositories to reports

### DIFF
--- a/app/views/reports/elements/_my_module_element.html.erb
+++ b/app/views/reports/elements/_my_module_element.html.erb
@@ -99,7 +99,7 @@
       <% end %>
     <% end %>
 
-    <% @settings.dig('deleted_repositories')&.each do |repo| %>
+    <% @settings['deleted_repositories']&.each do |repo| %>
       <%= render partial: 'reports/elements/my_module_repository_element.html.erb', locals: { repository_snapshot: my_module.repository_snapshots.where(parent_id: repo).find_by(selected: true), export_all: export_all, my_module: my_module } %>
     <% end %>
   </div>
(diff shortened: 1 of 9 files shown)


app/views/reports/elements/_my_module_element.html.erb
```diff
@@ -98,6 +98,10 @@
         <%= render partial: 'reports/elements/my_module_result_text_element.html.erb', locals: { result: result, export_all: export_all } %>
       <% end %>
     <% end %>
+
+    <% @settings.dig('deleted_repositories')&.each do |repo| %>
```
artoscinote, line 102: `dig` is not needed here, as there's only one key

app/views/reports/elements/_my_module_repository_element.html.erb
```diff
@@ -1,6 +1,7 @@
 <% my_module ||= @my_module %>
-<% element_id ||= repository&.id %>
+<% repository ||= nil %>
```
artoscinote, line 2: this doesn't look like it does anything, as it will assign `nil` if it's already `nil`

app/views/reports/wizard/_third_step.html.erb
```diff
@@ -73,6 +73,24 @@
                   <span class="sci-checkbox-label"></span>
                 </span>
                 <%= repository.name %>
+                <% unless repository.active? %>
+                  <span class="arhived">
```
artoscinote, line 77: typo in class :P

artoscinote (changes requested):

artoscinote (approved):

PR #3396, opened 2021-06-17
Add webhooks backend [SCI-5800]

artoscinote:
Jira ticket: [SCI-5800](https://biosistemika.atlassian.net/browse/SCI-5800)

### What was done
Implemented activity filter matcher, with which we can find relevant webhooks to exectue, when an activity is created.
Implemented webhook execution service with automatic disabling of faulty webhooks.
Implemented relevant jobs for executing activity related webhooks.
Wrote appropriate specs.


app/jobs/activities/send_webhook_job.rb
```diff
@@ -0,0 +1,11 @@
...
+
+module Activities
+  class SendWebhookJob < ApplicationJob
+    queue_as :high_priority
+
+    def perform(webhook, activity)
+      Activities::ActivityWebhookService.new(webhook, activity).send_webhook
+    end
+  end
+end
```
artoscinote, line 11: This could maybe just be a generalized webhook job, but I made it separate in case we maybe have some lower priority webhooks in the future.

app/services/activities/activity_webhook_service.rb
```diff
@@ -0,0 +1,21 @@
...
+    def initialize(webhook, activity)
+      @webhook = webhook
+      @activity = activity
+    end
+
+    def send_webhook
+      WebhookService.new(@webhook, activity_payload).send_webhook
+    end
+
+    def activity_payload
```
artoscinote, line 14: not sure if this is correct or if we want to define the payload as something else, and if we maybe want to move it into a serializer or some other place

db/migrate/20210617111749_rename_webhook_method_column.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
```
okriuchykhin, line 1: Merge it into one migration

app/services/webhook_service.rb
```diff
@@ -0,0 +1,56 @@
...
+  ensure
+    disable_webhook_if_broken!
+  end
+
+  private
+
+  def log_error!(message)
+    error_count = @webhook.error_count + 1
+
+    @webhook.update(
```
okriuchykhin, line 45: update!

app/services/webhook_service.rb
```diff
@@ -0,0 +1,56 @@
...
+      error_count: error_count,
+      last_error: message
+    )
+  end
+
+  def disable_webhook_if_broken!
+    return if @webhook.error_count < DISABLE_WEBHOOK_ERROR_THRESHOLD
+
+    @webhook.update(active: false)
```
okriuchykhin, line 54: update!

app/jobs/activities/dispatch_webhooks_job.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    def perform(activity)
+      webhooks =
+        Webhook.active.where(
+          activity_filter_id:
+            Activities::ActivityFilterMatchingService.new(activity).activity_filters.select(:id)
+        )
+
+      webhooks.each do |webhook|
+        Activities::SendWebhookJob.perform_later(webhook, activity)
```
okriuchykhin, line 15: is it really needed to schedule another jobs?

app/models/activity.rb
```diff
@@ -63,6 +63,8 @@ class Activity < ApplicationRecord
     breadcrumbs: {}
   )
 
+  after_create :dispatch_webhooks
```
okriuchykhin, line 66: after_create -> (activity) { Activities::DispatchWebhooksJob.perform_later(activity) }

app/jobs/activities/dispatch_webhooks_job.rb
artoscinote: Since some of them can fail, we would need to add additional exception handling in the loop here (something like `rescue nil` at the end of the line to ignore exceptions), if performed inline. It seemed cleaner to just dispatch them as separate jobs and keep the dispatch job simpler, and keep exception handling in one place.

artoscinote: One reason is also webhook retries, which wouldn't work if they were all performed here.

okriuchykhin (approved):

PR #3398, opened 2021-06-18
Expose experiment IDs in interface and reports [SCI-5828]

artoscinote:
Jira ticket: [SCI-5828](https://biosistemika.atlassian.net/browse/SCI-5828)

### What was done
Added experiment ids to card and table view, added to global search, added to exprots.


db/migrate/20210618104853_add_code_index_to_experiments.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddCodeIndexToExperiments < ActiveRecord::Migration[6.1]
+  def change
+    add_index :experiments, "('EX' || id)", name: 'index_experiments_on_experiment_code', unique: true
```
okriuchykhin, line 5: ID is a primary key, why do you need unique constrain here?

artoscinote: if you know the value is unique, it's supposed to improve performance of the index, as it can be better optimized

okriuchykhin: `SET enable_seqscan = OFF;
EXPLAIN ANALYZE SELECT * FROM experiments WHERE ('EX' || id) ILIKE 'EX12%';
                                                         QUERY PLAN                                                         
----------------------------------------------------------------------------------------------------------------------------
 Seq Scan on experiments  (cost=10000000000.00..10000000096.14 rows=9 width=503) (actual time=0.054..0.784 rows=11 loops=1)
   Filter: (('EX'::text || (id)::text) ~~* 'EX12%'::text)
   Rows Removed by Filter: 1096
 Planning Time: 0.645 ms
 Execution Time: 0.822 ms
(5 rows)`

db/migrate/20210622101238_fix_experiment_indices.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+class FixExperimentIndices < ActiveRecord::Migration[6.1]
+  def up
+    remove_index :experiments, name: 'index_experiments_on_name', column: 'name'
+
+    ActiveRecord::Base.connection.execute(
```
okriuchykhin, line 7: There is 'add_gin_index_without_tags' helper for this

db/migrate/20210622101238_fix_experiment_indices.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
```
okriuchykhin, line 1: You can merge it with previous migration

db/migrate/20210618104853_add_code_index_to_experiments.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class AddCodeIndexToExperiments < ActiveRecord::Migration[6.1]
+  def up
+    ActiveRecord::Base.connection.execute(
+      "CREATE INDEX index_experiments_on_experiment_code ON experiments using gin (('EX'::text || id) gin_trgm_ops);"
```
okriuchykhin, line 6: Can you use 'add_gin_index_without_tags' here?

app/models/experiment.rb
```diff
@@ -70,19 +70,19 @@ def self.search(
       new_query =
         Experiment
         .where('experiments.project_id IN (?)', projects_ids)
-        .where_attributes_like([:name, :description], query, options)
+        .where_attributes_like([:name, :description, "('EX' || id)"], query, options)
```
okriuchykhin, line 73: Move it to class constant?

app/models/concerns/searchable_model.rb
```diff
@@ -43,6 +43,8 @@ module SearchableModel
             (attrs.map.with_index do |a, i|
               if %w(repository_rows.id repository_number_values.data).include?(a)
                 "((#{a})::text) #{like} :t#{i} OR "
+              elsif a == "('EX' || id)"
```
okriuchykhin, line 46: why it is needed here?

artoscinote: So it wouldn't run the trim_html_tags to this thing, as it is unnecessary. But I guess if I use the helper it will do it anyway. Should we run the trim function on this also?

db/migrate/20210618104853_add_code_index_to_experiments.rb
artoscinote: It would try and interpolate `('EX'::text || id) ` into the index name, so I guess not, unless I modify the helper. It would also run the trim_html_tags on it, which is unnecessary.

okriuchykhin (approved):

PR #3414, opened 2021-06-30
Add biomolecular toolkit API proxying endpoint [SCI-5843]

okriuchykhin:
Jira ticket: [SCI-5843](https://biosistemika.atlassian.net/browse/SCI-5843)

### What was done
Add biomolecular toolkit API proxying endpoint

### DIFF
--- a/app/controllers/bio_eddie_assets_controller.rb
+++ b/app/controllers/bio_eddie_assets_controller.rb
@@ -3,10 +3,10 @@
 class BioEddieAssetsController < ApplicationController
   include ActiveStorage::SetCurrent
 
-  before_action :load_vars, except: :create
+  before_action :load_vars, except: %i(create bmt_request)
   before_action :load_create_vars, only: :create
 
-  before_action :check_read_permission
+  before_action :check_read_permission, except: %i(update create start_editing bmt_request)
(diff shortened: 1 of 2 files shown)


app/controllers/bio_eddie_assets_controller.rb
```diff
@@ -3,10 +3,10 @@
 class BioEddieAssetsController < ApplicationController
   include ActiveStorage::SetCurrent
 
-  before_action :load_vars, except: :create
+  before_action :load_vars, except: %i(create bmt_request)
   before_action :load_create_vars, only: :create
 
-  before_action :check_read_permission
+  before_action :check_read_permission, except: %i(update create start_editing bmt_request)
```
artoscinote, line 9: Prefer `only:`, especially if there are many exceptions (and `only:` is more explicit).

artoscinote (changes requested):

app/controllers/bio_eddie_assets_controller.rb
okriuchykhin: read permission checks are default, you should explicitly disable it for specific actions

artoscinote: agreed.

artoscinote (approved):

PR #3427, opened 2021-07-10
Add logic to propagate the user assignment on project's experiments and tasks

ZmagoD:
Jira ticket: [SCI-5894](https://biosistemika.atlassian.net/browse/SCI-5894)

### What was done
Added a logic that propagates the user assignment to project's experiments and tasks

#### ToDo:

- [ ] Test with real data

### Note:
_(Optional) Note any other remarks, problems you were facing during the development, suggestion etc_

### DIFF
--- a/db/migrate/20210222123823_migrate_to_new_user_roles.rb
+++ b/db/migrate/20210222123823_migrate_to_new_user_roles.rb
@@ -49,13 +49,20 @@ class MigrateToNewUserRoles < ActiveRecord::Migration[6.1]
 
   def create_user_assignments(user_projects, user_role)
     user_projects.includes(:user, :project).find_in_batches(batch_size: 100) do |user_project_batch|
-      user_assignments = []
       user_project_batch.each do |user_project|
-        user_assignments << UserAssignment.new(user: user_project.user,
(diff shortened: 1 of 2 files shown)


ZmagoD:
@okriuchykhin @radiokills this is ready for code review. Since the test data was not provided I couldn't test the migration but actually it uses the same logic for propagating the assignments through the project tree so it should be fine. 

@okriuchykhin @radiokills what will happen with the enum on `UserProject` mode? Should we open a task for removing this field on the next release after migration?

db/migrate/20210222123823_migrate_to_new_user_roles.rb
```diff
@@ -49,13 +49,20 @@ def change
 
   def create_user_assignments(user_projects, user_role)
     user_projects.includes(:user, :project).find_in_batches(batch_size: 100) do |user_project_batch|
-      user_assignments = []
       user_project_batch.each do |user_project|
-        user_assignments << UserAssignment.new(user: user_project.user,
-                                               assignable: user_project.project,
-                                               user_role: user_role)
+        UserAssignment.create(
```
okriuchykhin, line 53: create!

PR #3439, opened 2021-07-15
Implement Save & Register molecule [SCI-5879]

artoscinote:
Jira ticket: [SCI-5879](https://biosistemika.atlassian.net/browse/SCI-5879)

### What was done
Implemented save & register molecule functionality.


app/models/asset.rb
```diff
@@ -445,6 +445,10 @@ def generate_base64(style)
     return convert_variant_to_base64(medium_preview) if style == :medium
   end
 
+  def my_module
+    (result || step).my_module
```
aignatov-bio, line 449: What if the step does not belong to the task?

aignatov-bio (changes requested):

PR #3441, opened 2021-07-16
Added prefixed IDs to repository rows (items) [SCI-5909]

artoscinote:
Jira ticket: [SCI-5909](https://biosistemika.atlassian.net/browse/SCI-5909)

### What was done
Generalized the way to enable prefixed IDs on models, added prefixed IDs + all the surrounding functionalities to RepositoryRow.

### DIFF
--- a/app/models/my_module.rb
+++ b/app/models/my_module.rb
@@ -313,7 +313,7 @@ class MyModule < ApplicationRecord
     rows = repository.assigned_rows(self).includes(:created_by).order(created_at: order)
     rows.find_each do |row|
       row_json = []
-      row_json << (row.repository.is_a?(RepositorySnapshot) ? row.parent.code : row.code)
+      row_json << (row.parent&.code || row.code)
       row_json << (row.archived ? "#{row.name} [#{I18n.t('general.archived')}]" : row.name)
       row_json << I18n.l(row.created_at, format: :full)
       row_json << row.created_by.full_name
--- a/app/models/my_module.rb
+++ b/app/models/my_module.rb
@@ -313,7 +313,7 @@ class MyModule < ApplicationRecord
     rows = repository.assigned_rows(self).includes(:created_by).order(created_at: order)
     rows.find_each do |row|
       row_json = []
-      row_json << (row.parent&.code || row.code)
+      row_json << row.code
       row_json << (row.archived ? "#{row.name} [#{I18n.t('general.archived')}]" : row.name)
       row_json << I18n.l(row.created_at, format: :full)
       row_json << row.created_by.full_name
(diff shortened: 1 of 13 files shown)


app/models/repository.rb
```diff
@@ -80,8 +80,14 @@ def self.search(
 
     matched_by_user = readable_rows.joins(:created_by).where_attributes_like('users.full_name', query, options)
 
+    prefixed_repository_row_ids = RepositoryRow.where(repository_id: repositories)
+                                               .where_attributes_like([RepositoryRow::PREFIXED_ID_SQL], query, options)
+                                               .pluck(:id)
```
okriuchykhin, line 85: What if you have million rows and search string is 'IT1'?

app/models/my_module.rb
```diff
@@ -313,7 +313,7 @@ def repository_json_hot(repository, order)
     rows = repository.assigned_rows(self).includes(:created_by).order(created_at: order)
     rows.find_each do |row|
       row_json = []
-      row_json << (row.repository.is_a?(RepositorySnapshot) ? row.parent_id : row.id)
+      row_json << (row.repository.is_a?(RepositorySnapshot) ? row.parent.code : row.code)
```
okriuchykhin, line 316: if parent is nil?

artoscinote: I thought it cannot be nil if it's from a snapshot.

app/models/repository.rb
artoscinote: Previously, this was doing a search by "id LIKE ..." which I think was even worse. Should we just remove this from global search? Although I don't think it would ever reach millions of records.

app/models/my_module.rb
okriuchykhin: snapshot of deleted inventory?

app/models/repository.rb
okriuchykhin: and now you simply insert ids as query parameter?

artoscinote: fixed to use subquery

app/models/my_module.rb
artoscinote: fixed

app/models/my_module.rb
```diff
@@ -313,7 +313,7 @@ def repository_json_hot(repository, order)
     rows = repository.assigned_rows(self).includes(:created_by).order(created_at: order)
     rows.find_each do |row|
       row_json = []
-      row_json << (row.repository.is_a?(RepositorySnapshot) ? row.parent_id : row.id)
+      row_json << (row.parent&.code || row.code)
```
okriuchykhin, line 316: So it will show code of the new item in the snapshot for deleted inventory instead of original ID?

okriuchykhin (approved):

PR #3443, opened 2021-07-16
Implement label templates with default ZPL template [SCI-5902]

artoscinote:
Jira ticket: [SCI-5902](https://biosistemika.atlassian.net/browse/SCI-5902)

### What was done
Implemented label templates along with a default ZPL template.


Gemfile
```diff
@@ -110,6 +110,8 @@ gem 'newrelic_rpm'
 # Permission helper Gem
 gem 'canaid', git: 'https://github.com/biosistemika/canaid', branch: 'rails_6'
 
+gem 'ruby-handlebars', require: false
```
okriuchykhin, line 113: Why do we need yet another templating engine??

artoscinote: switched to interpolation

okriuchykhin (approved):

PR #3504, opened 2021-08-23
Amended label template, truncate item name before print [SCI_6025]

artoscinote:
Jira ticket: [SCI-6025](https://biosistemika.atlassian.net/browse/SCI-6025)

### What was done
Amended default label template, now we truncate item name before printing.

### DIFF
--- a/app/controllers/repository_rows_controller.rb
+++ b/app/controllers/repository_rows_controller.rb
@@ -4,6 +4,8 @@ class RepositoryRowsController < ApplicationController
   include ApplicationHelper
   include MyModulesHelper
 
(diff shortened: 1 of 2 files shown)


app/controllers/repository_rows_controller.rb
```diff
@@ -99,7 +99,7 @@ def print
         label_printer,
         LabelTemplate.first.render( # Currently we will only use the default template
           item_id: repository_row.code,
-          item_name: repository_row.name
+          item_name: repository_row.name.truncate(64)
```
okriuchykhin, line 102: Shouldn't it be some sort of constant?

okriuchykhin (approved):

PR #3540, opened 2021-09-10
Fix item import Date type handling, properly ignore empty Excell rows [SCI-6053]

artoscinote:
Jira ticket: [SCI-6053](https://biosistemika.atlassian.net/browse/SCI-6053)

### What was done
Fix item import Date type handling, properly ignore empty Excell rows

### DIFF
--- a/app/models/repository_date_value.rb
+++ b/app/models/repository_date_value.rb
@@ -2,12 +2,16 @@
 
 class RepositoryDateValue < RepositoryDateTimeValueBase
   def data_changed?(new_data)
-    new_time = Time.zone.parse(new_data)
-    new_time.to_date != data.to_date
+    Time.use_zone('UTC') do
+      new_time = Time.zone.parse(new_data)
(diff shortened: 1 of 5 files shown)


app/models/repository_date_value.rb
```diff
@@ -18,7 +18,7 @@ def self.new_with_payload(payload, attributes)
 
   def self.import_from_text(text, attributes, options = {})
     date_format = (options.dig(:user, :settings, :date_format) || Constants::DEFAULT_DATE_FORMAT).gsub(/%-/, '%')
-    new(attributes.merge(data: DateTime.strptime(text, date_format)))
+    new(attributes.merge(data: Date.strptime(text, date_format)))
```
okriuchykhin, line 21: We have DateTime in the DB

okriuchykhin: so I think it will not solve an issue completely

PR #3563, opened 2021-09-29
Add create/destroy endpoints to connections API, add tests [SCI-6087]

artoscinote:
Jira ticket: [SCI-6087](https://biosistemika.atlassian.net/browse/SCI-6087)

### What was done
Added create/destroy endpoints to connections API.

### DIFF
--- a/app/controllers/api/v1/connections_controller.rb
+++ b/app/controllers/api/v1/connections_controller.rb
@@ -8,6 +8,7 @@ module Api
       before_action :load_experiment
       before_action :load_connections
       before_action :load_connection, only: :show
+      before_action :check_manage_permissions, except: %i(index show)
 
       def index
         @connections = @connections.page(params.dig(:page, :number))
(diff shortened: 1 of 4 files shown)


app/controllers/api/v1/connections_controller.rb
```diff
@@ -20,28 +21,48 @@ def index
       def show
         render jsonapi: @connection,
                serializer: ConnectionSerializer,
-               include: %i(input_task output_task)
+               include: %i(to from)
+      end
+
+      def create
+        connection = Connection.create(connection_params)
```
okriuchykhin, line 28: create!

artoscinote: fixed

app/models/experiment.rb
```diff
@@ -49,6 +49,17 @@ class Experiment < ApplicationRecord
     end
   }
 
+  def connections
```
okriuchykhin, line 52: It does not belong to here

okriuchykhin (approved):

PR #3565, opened 2021-09-30
Add read permission check for my_module in report creation [SCI-6108]

okriuchykhin:
Jira ticket: [SCI-6108](https://biosistemika.atlassian.net/browse/SCI-6108)

### What was done
Add read permission check for my_module in report creation


radiokills:
Fixed in another PR by Arto. It's wrong anyway.

PR #3589, opened 2021-10-12
Fixed unassigned user methods, order by name [SCI-6104]

artoscinote:
Jira ticket: [SCI-6104](https://biosistemika.atlassian.net/browse/SCI-6104)

### What was done
Fixed unassigned user methods, order by name in my module designation dropdown

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -196,16 +196,20 @@ class Project < ApplicationRecord
   end
 
   def unassigned_users
-    User
-      .joins('INNER JOIN user_teams ON users.id = user_teams.user_id')
-      .where('user_teams.team_id = ?', team)
-      .where.not(confirmed_at: nil)
-      .where('users.id NOT IN (?)',
-             UserProject.where(project: self).select(:user_id).distinct)
+    User.joins(:user_teams)
+        .joins(
+          "LEFT OUTER JOIN user_assignments ON user_assignments.user_id = users.id "\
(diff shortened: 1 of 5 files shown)


app/views/user_my_modules/_index_edit.html.erb
```diff
@@ -34,13 +34,13 @@
   <% end %>
 <% end %>
 
-<% if can_manage_my_module_users?(@my_module) && @unassigned_users.count > 0 %>
+<% if can_manage_my_module_users?(@my_module) && @undesignated_users.count > 0 %>
```
okriuchykhin, line 37: any?

app/models/project.rb
```diff
@@ -196,16 +196,20 @@ def last_comments(last_id = 1, per_page = Constants::COMMENTS_SEARCH_LIMIT)
...
+        )
+        .where(user_teams: { team_id: team_id })
+        .where(user_assignments: { id: nil })
+        .where.not(confirmed_at: nil)
+        .distinct
   end
 
   def user_role(user)
-    user_assignments.find_by_user_id(user)&.user_role.name
+    UserRole.joins(:user_assignments).find_by(user_assignments: { user: user, assignable: self })&.name
```
okriuchykhin, line 212: why did you change it?

okriuchykhin: The only difference here is join type. I just prefer to use model relation instead of global select

PR #3626, opened 2021-11-02
Fix experiment move service [SCI-6128]

artoscinote:
Jira ticket: [SCI-6128](https://biosistemika.atlassian.net/browse/SCI-6128)

### What was done
Fixed experiment move service.

### DIFF
--- a/app/services/experiments/move_to_project_service.rb
+++ b/app/services/experiments/move_to_project_service.rb
@@ -83,7 +83,9 @@ module Experiments
 
     # recursively move all activities in child associations to new project
     def move_activities!(subject)
-      Activity.where(subject: subject).update(project: @project)
+      Activity.where(subject: subject).find_each do |activity|
+        activity.update!(project: @project)
+      end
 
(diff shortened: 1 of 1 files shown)


app/services/experiments/move_to_project_service.rb
```diff
@@ -83,7 +83,7 @@ def move_tags!(my_module)
 
     # recursively move all activities in child associations to new project
     def move_activities!(subject)
-      Activity.where(subject: subject).update!(project: @project)
```
okriuchykhin, line 86: do you understand difference between update and update! ?

artoscinote: Yes I do. And there is no .update! method for ActiveRecord:Relation.

okriuchykhin: Activity.where(subject: subject).find_each { |activity| activity.update!(project: @project) }

okriuchykhin (approved):

PR #3655, opened 2021-11-11
Fixed permission issues when moving experiment [SCI-6257]

artoscinote:
Jira ticket: [SCI-6257](https://biosistemika.atlassian.net/browse/SCI-6257)

### What was done
Fixed movable_projects method on experiment, added permission check to move service.

### DIFF
--- a/app/services/experiments/move_to_project_service.rb
+++ b/app/services/experiments/move_to_project_service.rb
@@ -18,7 +18,7 @@ module Experiments
     def call
       return self unless valid?
 
-      unless @project.permission_granted?(@user, ProjectPermissions::EXPERIMENTS_CREATE)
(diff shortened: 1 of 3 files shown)


app/services/experiments/move_to_project_service.rb
```diff
@@ -18,6 +18,11 @@ def initialize(experiment_id:, project_id:, user_id:)
     def call
       return self unless valid?
 
+      unless @project.permission_granted?(@user, ProjectPermissions::EXPERIMENTS_CREATE)
```
okriuchykhin, line 21: Why not can_create_project_experiments? ?

okriuchykhin (approved):

PR #3660, opened 2021-11-12
Fixed bug in _invite_users_modal_results.html.erb

Denys-Priadko:
### What was done
Fixed bug in _invite_users_modal_results.html.erb
the bug: when inviting users, after a new user is generated (if needed) and email is sent, an error occurs in the partial responsible for generating the "all is done" message, so that instead of "all is done" one sees "error inviting users", even though the users have been invited successfully. 
The reason for the error is the mismatch between the expected structure of @invite_results and the actual @invite_results.

### DIFF
--- a/app/views/shared/_invite_users_modal_results.html.erb
+++ b/app/views/shared/_invite_users_modal_results.html.erb
@@ -4,6 +4,9 @@
 <hr />
 <div class="results-wrap">
 <% @invite_results.each do |result| %>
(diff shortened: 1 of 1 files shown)


Denys-Priadko:
Update: 
after looking closer at the code, I see that the issue is deeper. 
invitations_controller.rb supports multiple teams per user, but it only supports one status per user. Since the :status depends on both the user and the team, it seems to me that each result in @invite_results should have as many statuses as teams.

PR #3692, opened 2021-11-22
Add endpoint for fetching column data including connected items [SCI-6208]

okriuchykhin:
Jira ticket: [SCI-6208](https://biosistemika.atlassian.net/browse/SCI-6208)

### What was done
Add endpoint for fetching column data including connected items

### DIFF
--- a/app/controllers/repository_columns_controller.rb
+++ b/app/controllers/repository_columns_controller.rb
@@ -34,6 +34,18 @@ class RepositoryColumnsController < ApplicationController
     raise NotImplementedError
   end
 
+  def describe_all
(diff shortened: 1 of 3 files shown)


app/controllers/repository_columns_controller.rb
```diff
@@ -34,6 +34,18 @@ def create
     raise NotImplementedError
   end
 
+  def describe_all
+    response_json = @repository.repository_columns.map do |column|
```
aignatov-bio, line 38: Can we also merge default columns here, it will be the full list of columns.  But not sure which `id` default columns will have.

aignatov-bio (approved):

PR #3716, opened 2021-12-02
[SCI-6322] use correct user when creating templates, create assignmen…

radiokills:
…t on experiments and tasks

Jira ticket: [SCI-6322](https://biosistemika.atlassian.net/browse/SCI-6322)

### What was done
There were two issues with permissions on templates:
1. The user we tried to assign to the imported experiments was still from old `user_projects` relation
2. The permissions were not assigned downstream

### DIFF
--- a/app/services/team_importer.rb
+++ b/app/services/team_importer.rb
@@ -192,9 +192,9 @@ class TeamImporter
         
         # handle the permissions for newly created experiment
         user = User.find(user_id)
-        UserAssignments::GenerateUserAssignmentsJob.perform_later(experiment, user)
+        UserAssignments::GenerateUserAssignmentsJob.perform_now(experiment, user)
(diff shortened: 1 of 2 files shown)


app/services/team_importer.rb
```diff
@@ -189,7 +189,13 @@ def import_experiment_template_from_dir(import_dir, project_id,
         end
 
         update_smart_annotations_in_project(project)
-
+        
+        # handle the permissions for newly created experiment
+        user = User.find(user_id)
+        UserAssignments::GenerateUserAssignmentsJob.perform_now(experiment, user)
+        experiment.my_modules.each do |my_module|
```
artoscinote, line 196: use `find_each`

artoscinote (approved):

PR #3894, opened 2022-02-24
Enable result comment button after canceling result edit [SCI-6537]

rekonder:
Jira ticket: [SCI-6537](https://biosistemika.atlassian.net/browse/SCI-6537)

Enable save button for comments after cancelling editing table or asset result

### DIFF
--- a/app/assets/javascripts/results/result_assets.js
+++ b/app/assets/javascripts/results/result_assets.js
@@ -35,7 +35,7 @@
     }
 
     function applyEditResultAssetCallback() {
-      $('.edit-result-asset').off().on('ajax:success', function(e, data) {
+      $('.edit-result-asset').off('ajax:success ajax:error').on('ajax:success', function(e, data) {
         var $result = $(this).closest('.result');
         var $form = $(data.html);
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/results/result_assets.js
```diff
@@ -35,18 +35,18 @@
     }
 
     function applyEditResultAssetCallback() {
-      $('.edit-result-asset').on('ajax:success', function(e, data) {
+      $('.edit-result-asset').off().on('ajax:success', function(e, data) {
```
aignatov-bio, line 38: It is quite dangerous to use `.off()` it can remove all events. Check this:
```
var foo = function() {
  // Code to handle some kind of event
};
 
// ... Now foo will be called when paragraphs are clicked ...
$( "body" ).on( "click", "p", foo );
 
// ... Foo will no longer be called.
$( "body" ).off( "click", "p", foo );
```

aignatov-bio (changes requested):

aignatov-bio (approved):

PR #3911, opened 2022-03-02
Implement basic logic for Date/DateTime/Stock reminders [SCI-6554]

artoscinote:
Jira ticket: [SCI-6554](https://biosistemika.atlassian.net/browse/SCI-6554)

### What was done
Implement basic logic for Date/DateTime reminders

### DIFF
--- a/app/helpers/repository_datatable_helper.rb
+++ b/app/helpers/repository_datatable_helper.rb
@@ -4,6 +4,8 @@ module RepositoryDatatableHelper
   include InputSanitizeHelper
 
   def prepare_row_columns(repository_rows, repository, columns_mappings, team, options = {})
+    repository_rows_with_active_reminders = repository.repository_rows.with_active_reminders
+
     repository_rows.map do |record|
       default_cells = {
         '1': assigned_row(record),
@@ -18,7 +20,14 @@ module RepositoryDatatableHelper
       row = {
         'DT_RowId': record.id,
         'DT_RowAttr': { 'data-state': row_style(record) },
-        'recordInfoUrl': Rails.application.routes.url_helpers.repository_repository_row_path(repository, record)
+        'recordInfoUrl': Rails.application.routes.url_helpers.repository_repository_row_path(repository, record),
+        'hasActiveReminders': repository_rows_with_active_reminders.include?(record),
+        'rowRemindersUrl':
+          Rails.application.routes.url_helpers
+               .active_reminder_repository_cells_repository_repository_row_url(
+                 repository,
+                 record
+               )
       }.merge(default_cells)
 
       if record.repository.has_stock_management?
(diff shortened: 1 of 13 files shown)


app/controllers/repository_columns/stock_columns_controller.rb
```diff
@@ -47,7 +47,7 @@ def search_params
     def repository_column_params
       params
         .require(:repository_column)
-        .permit(:name, metadata: [:decimals], repository_stock_unit_items_attributes: %i(data))
+        .permit(:name, :low_stock_threshold, metadata: [:decimals], repository_stock_unit_items_attributes: %i(data))
```
aignatov-bio, line 50: Why this here?

app/models/repository_date_time_value_base.rb
```diff
@@ -12,6 +12,14 @@ class RepositoryDateTimeValueBase < ApplicationRecord
 
   validates :repository_cell, :data, :type, presence: true
 
+  scope :with_active_reminder, lambda {
```
aignatov-bio, line 15: We don't have it in time

app/models/repository_row.rb
```diff
@@ -84,6 +85,31 @@ class RepositoryRow < ApplicationRecord
   scope :active, -> { where(archived: false) }
   scope :archived, -> { where(archived: true) }
 
+  scope :in_repository_with_active_reminders, lambda { |repository|
```
aignatov-bio, line 88: Can we just simplify it to a predefined repository?
`@repository.repository_rows.with_active_reminders`

Because I don't get point why we need this query for `repository_cell_ids` if we already will have scoped rows.

And how you will solve duplicate rows? If we have 2 reminders in one row then it will return duplicate rows

aignatov-bio (changes requested):

app/helpers/repository_datatable_helper.rb
```diff
@@ -18,7 +20,14 @@ def prepare_row_columns(repository_rows, repository, columns_mappings, team, opt
       row = {
         'DT_RowId': record.id,
         'DT_RowAttr': { 'data-state': row_style(record) },
-        'recordInfoUrl': Rails.application.routes.url_helpers.repository_repository_row_path(repository, record)
+        'recordInfoUrl': Rails.application.routes.url_helpers.repository_repository_row_path(repository, record),
+        'hasActiveReminders': repository_rows_with_active_reminders.include?(record),
```
aignatov-bio, line 24: Does not trigger a new query on each row?

app/models/repository_row.rb
```diff
@@ -84,6 +85,12 @@ class RepositoryRow < ApplicationRecord
   scope :active, -> { where(archived: false) }
   scope :archived, -> { where(archived: true) }
 
+  scope :with_active_reminders, lambda {
+    joins(:repository_cells).where(
+      repository_cells: { id: RepositoryCell.with_active_reminder.select(:id) }
```
aignatov-bio, line 90: It can be an issue on free Scinote, I think there are millions of cells.

aignatov-bio (changes requested):

PR #3935, opened 2022-03-11
Improve stock ledger records creation [SCI-6419]

okriuchykhin:
Jira ticket: [SCI-6419](https://biosistemika.atlassian.net/browse/SCI-6419)

### What was done
Improve stock ledger records creation

### DIFF
--- a/app/models/my_module_repository_row.rb
+++ b/app/models/my_module_repository_row.rb
@@ -15,9 +15,8 @@ class MyModuleRepositoryRow < ApplicationRecord
 
   validates :repository_row, uniqueness: { scope: :my_module }
 
-  around_save :deduct_stock_balance, if: :stock_consumption_changed?
-
   before_save :nulify_stock_consumption, if: :stock_consumption_changed?
+  around_save :deduct_stock_balance, if: :stock_consumption_changed?
 
   def consume_stock(user, stock_consumption, comment = nil)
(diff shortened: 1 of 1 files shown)


app/models/repository_stock_value.rb
```diff
@@ -87,13 +87,13 @@ def self.import_from_text(text, attributes, _options = {})
...
 
-  def update_stock_with_ledger!(amount, reference, comment)
+  def update_stock_with_ledger!(amount, comment)
     delta = amount.to_d - self.amount.to_d
     repository_ledger_records.create!(
       user: last_modified_by,
       amount: delta,
       balance: amount,
-      reference: reference,
+      reference: repository_cell.repository_column.repository,
```
artoscinote, line 96: I think the reference is supposed to be either the repository (if change came from repository management), or a MyModuleRepositoryRow, if it was done via task stock consumption.

okriuchykhin: is it possible to change stock from task?

artoscinote: Looks like there's separate logic for that, the `deduct_stock_balance` callback method in `app/models/my_module_repository_row.rb`.
Maybe that logic should be unified to use this method, in which case the reference would need to go back in as an argument.

artoscinote (approved):

PR #3938, opened 2022-03-15
Add stock filters [SCI-6547]

aignatov-bio:
Jira ticket: [SCI-6547](https://biosistemika.atlassian.net/browse/SCI-6547)

### What was done
Add stock filters

### DIFF
--- a/app/models/repository_stock_value.rb
+++ b/app/models/repository_stock_value.rb
@@ -22,6 +22,44 @@ class RepositoryStockValue < ApplicationRecord
     amount <= low_stock_threshold
   end
 
+  def self.add_filter_condition(repository_rows, join_alias, filter_element)
+    parameters = filter_element.parameters
+    if filter_element.operator == 'between'
+      return repository_rows if parameters['from'].blank? || parameters['to'].blank?
+    elsif parameters['value'].blank?
+      return repository_rows
+    end
+
+    repository_rows = case parameters['stock_unit']
+                      when 'all'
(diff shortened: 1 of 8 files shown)


app/javascript/vue/repository_filter/container.vue
```diff
@@ -89,6 +89,7 @@
         const index = this.filters.findIndex((f) => f.id === filter.id);
         this.filters[index].data = filter.data;
         this.filters[index].isBlank = filter.isBlank;
+        console.log(this.filters)
```
artoscinote, line 92: should be removed

app/models/repository_stock_value.rb
```diff
@@ -22,6 +22,44 @@ def low_stock?
     amount <= low_stock_threshold
   end
 
+  def self.add_filter_condition(repository_rows, join_alias, filter_element)
+    parameters = filter_element.parameters
+    if filter_element.operator == 'between'
+      return repository_rows if parameters['from'].blank? || parameters['to'].blank?
+    elsif parameters['value'].blank?
+      return repository_rows
+    end
```
artoscinote, line 31: isn't this already handled on the fronted by the isBlank compted property?

artoscinote (changes requested):

app/models/repository_stock_value.rb
aignatov-bio: Well, we have the same number I think this is for some redundancy. Cause otherwise, the query will be not correct.

artoscinote: alright

artoscinote (approved):

PR #3950, opened 2022-03-21
Implement failed snapshot pop-up on task screen [SCI-6612]

artoscinote:
Jira ticket: [SCI-6612](https://biosistemika.atlassian.net/browse/SCI-6612)

### What was done
Added general purpose 'dialog' modal.
Implement failed snapshot pop-up on task screen.

### DIFF
--- a/app/views/shared/_dialog.html.erb
+++ b/app/views/shared/_dialog.html.erb
@@ -1,4 +1,4 @@
-<div id="<%= id %>" class="sci-dialog modal fade" role="dialog">
+<div id="<%= id %>" class="sci-dialog modal fade <%= 'modal-auto-open' if defined?(shown) && shown %>" role="dialog">
   <div class="modal-dialog">
     <div class="modal-content">
       <div class="modal-header">
@@ -30,9 +30,3 @@
     </div>
   </div>
 </div>
-
-<% if defined?(shown) && shown %>
-  <script>
-    $('#<%= id %>').modal('show');
-  </script>
-<% end %>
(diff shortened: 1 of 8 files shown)


app/views/shared/_dialog.html.erb
```diff
@@ -0,0 +1,38 @@
...
+      </div>
+    </div>
+  </div>
+</div>
+
+<% if defined?(shown) && shown %>
+  <script>
+    $('#<%= id %>').modal('show');
+  </script>
+<% end %>
```
aignatov-bio, line 38: If this shared element just move this logic to sitewide

app/views/shared/_dialog.html.erb
```diff
@@ -0,0 +1,38 @@
...
+          <% when 'info' %>
+            <i class="fas fa-info"></i>
+          <% when 'success' %>
+            <i class="fas fa-check-circle"></i>
+          <% else %>
+            <i class="fas fa-info"></i>
+          <% end %>
+        </div>
+        <div class="sci-dialog__body">
+          <%= body.html_safe %>
```
aignatov-bio, line 22: Why html_safe?

artoscinote: Because in this case, the translation includes some bold tags.

aignatov-bio: Hm does `_html` suffix in I18n fix this?

artoscinote: Ah, didn't know that, yes it works without .html_safe. Fix pushed.

PR #3953, opened 2022-03-23
Add alias to cell sort query, to prevent join conflicts [SCI-6649]

artoscinote:
Jira ticket: [SCI-6649](https://biosistemika.atlassian.net/browse/SCI-6649)

### What was done
Added alias to repository_cell join in datatable cell value sort, to prevent issues with other joins.

### DIFF
--- a/app/services/repository_datatable_service.rb
+++ b/app/services/repository_datatable_service.rb
@@ -435,8 +435,10 @@ class RepositoryDatatableService
       sorting_data_type = sorting_column.data_type.constantize
       cells = sorting_data_type.joins(
         "INNER JOIN repository_cells AS repository_sort_cells "\
-        "ON repository_sort_cells.value_id = #{sorting_data_type.table_name}.id"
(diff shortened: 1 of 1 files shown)


app/services/repository_datatable_service.rb
```diff
@@ -433,21 +433,23 @@ def sort_rows(column_obj, records)
       return records unless sorting_column
 
       sorting_data_type = sorting_column.data_type.constantize
-      cells = sorting_data_type.joins(:repository_cell)
-                               .where('repository_cells.repository_column_id': sorting_column.id)
+      cells = sorting_data_type.joins(
+        "INNER JOIN repository_cells AS repository_sort_cells "\
+        "ON repository_sort_cells.value_id = #{sorting_data_type.table_name}.id"
```
okriuchykhin, line 438: May be it is better to add value_type to join condition for better performance?

artoscinote: Added to condition, now works with index.

okriuchykhin (approved):

PR #3964, opened 2022-03-28
Implement stock consumption via the API [SCI-6642]

artoscinote:
Jira ticket: [SCI-6642](https://biosistemika.atlassian.net/browse/SCI-6642)

### What was done
Implement stock consumption via the API, added some specs.


app/serializers/api/v1/inventory_item_serializer.rb
```diff
@@ -5,6 +5,7 @@ module V1
     class InventoryItemSerializer < ActiveModel::Serializer
       type :inventory_items
       attributes :name, :archived
+      attribute :stock_consumption, if: -> { object.repository_stock_cell.present? }
```
okriuchykhin, line 8: Is it needed here?

artoscinote: removed

app/permissions/my_module_repository_row.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+Canaid::Permissions.register_for(MyModuleRepositoryRow) do
+  can :update_stock_my_module_repository_row_stock_consumption do |user, my_module_repository_row|
```
okriuchykhin, line 4: Why do we need it?

artoscinote: Because it's a combination of states/permissions of the module, the row, and the repository. Otherwise we'd need to add the archived conditions at every point where we check the update consumption permission.

app/controllers/my_module_repositories_controller.rb
```diff
@@ -159,20 +159,12 @@ def consume_modal
 
   def update_consumption
     module_repository_row = @my_module.my_module_repository_rows.find_by(id: params[:module_row_id])
-    module_repository_row.with_lock do
+
+    ActiveRecord::Base.transaction do
```
okriuchykhin, line 163: with_lock creates transaction

app/permissions/my_module_repository_row.rb
artoscinote: fixed

app/controllers/my_module_repositories_controller.rb
artoscinote: Added nested transaction support to the consume_stock method (via require_new: true). One transaction is for locked stock update, the one in controller is to make sure activity only gets created when it's successful.

PR #3971, opened 2022-03-30
Disable stock reminders if stock management is disabled [SCI-6654]

artoscinote:
Jira ticket: [SCI-6654](https://biosistemika.atlassian.net/browse/SCI-6654)

### What was done
Disable stock reminders if stock management is disabled.


artoscinote:
Will be done a differnet way.

PR #4034, opened 2022-04-19
Add step orderable elements controller [SCI-6711]

okriuchykhin:
Jira ticket: [SCI-6711](https://biosistemika.atlassian.net/browse/SCI-6711)

### What was done
Add step orderable elements controller [SCI-6711]

[SCI-6711]: https://biosistemika.atlassian.net/browse/SCI-6711?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/step_orderable_elements_controller.rb
+++ b/app/controllers/step_orderable_elements_controller.rb
@@ -3,6 +3,7 @@
 class StepOrderableElementsController < ApplicationController
   before_action :load_vars_nested
   before_action :load_vars, only: :destroy
+  before_action :check_view_permissions, except: %i(create destroy)
   before_action :check_manage_permissions, only: %i(create destroy)
 
   def create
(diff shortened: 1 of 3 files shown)


aignatov-bio (changes requested):
Will we also migrate here TinyMCE images? Now they belong to step. And I think some updates for the TinyMceAssets controller and MarvinJS controller/service, we have all checks with Step as the object.

artoscinote:
Could we call this just StepElement? I think this Orderable part is a bit redundant here.

okriuchykhin:
> Could we call this just StepElement? I think this Orderable part is a bit redundant here.

But it clearly describes purpose of this model

app/controllers/step_orderable_elements_controller.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+class StepOrderableElementsController < ApplicationController
+  before_action :load_vars
+  before_action :check_manage_permissions
+
+  def update
+    @step.with_lock do
+      element_to_swap = @step.step_orderable_elements.find_by(id: params[:new_position])
```
aignatov-bio, line 9: I think we do not swap elements. We just move one element to the new position and all elements between old and new positions must be moved up/down depending on direction.

aignatov-bio (changes requested):

okriuchykhin:
Obsolete

PR #4041, opened 2022-04-20
Update stock value without reloading the table [SCI-6745]

artoscinote:
Jira ticket: [SCI-6745](https://biosistemika.atlassian.net/browse/SCI-6745)

### What was done
Update stock value without reloading the table

### DIFF
--- a/app/controllers/repository_stock_values_controller.rb
+++ b/app/controllers/repository_stock_values_controller.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class RepositoryStockValuesController < ApplicationController
+  include RepositoryDatatableHelper # for use of display_cell_value method on stock update
+
   before_action :load_vars
   before_action :check_manage_permissions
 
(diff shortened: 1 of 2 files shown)


app/controllers/repository_stock_values_controller.rb
```diff
@@ -45,7 +47,10 @@ def create_or_update
       )
     end
 
-    render json: @repository_stock_value
+    render json: {
```
aignatov-bio, line 50: Why just not use serializer?

aignatov-bio (changes requested):

app/controllers/repository_stock_values_controller.rb
artoscinote: Because it needs to be compatible with the DataTable cell to render it properly. These things are not in a serializer, but in the repository datatable service.

aignatov-bio (approved):

PR #4100, opened 2022-05-18
Renaming settings section and remove caching for Settings class [SCI-5625]

rekonder:
Jira ticket: [SCI-5625](https://biosistemika.atlassian.net/browse/SCI-5625)

### What was done
- Settings section renaming
- Disable caching for Settings class, because of logo uploading

### DIFF
--- a/app/models/settings.rb
+++ b/app/models/settings.rb
@@ -2,7 +2,6 @@
 
 class Settings < ApplicationRecord
   def self.instance
-    @instance = first
-    @instance ||= new
+    @instance = first || new
   end
 end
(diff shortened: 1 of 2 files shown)


app/models/settings.rb
```diff
@@ -2,7 +2,7 @@
 
 class Settings < ApplicationRecord
   def self.instance
-    @instance ||= first
+    @instance = first
```
okriuchykhin, line 5: Just:

first || new

app/models/settings.rb
```diff
@@ -2,7 +2,6 @@
 
 class Settings < ApplicationRecord
   def self.instance
-    @instance ||= first
-    @instance ||= new
+    @instance = first || new
```
okriuchykhin, line 5: you don't need to assign a variable here

okriuchykhin (approved):

PR #4115, opened 2022-05-24
Add consume column to assigned item export [SCI-6841]

rekonder:
Jira ticket: [SCI-6841](https://biosistemika.atlassian.net/browse/SCI-6841)

### What was done
Added consume column to assigned item export, and some code refactor

### DIFF
--- a/app/models/repository_row.rb
+++ b/app/models/repository_row.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class RepositoryRow < ApplicationRecord
+
+  include ActionView::Helpers::NumberHelper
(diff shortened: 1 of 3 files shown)


app/models/repository_row.rb
```diff
@@ -149,4 +150,17 @@ def snapshot!(repository_snapshot)
     repository_cells.each { |cell| cell.snapshot!(row_snapshot) }
     row_snapshot
   end
+
+  def row_consumption(stock_consumption)
```
artoscinote, line 154: maybe rename to row_consumption_formatted

PR #4147, opened 2022-06-06
Update tests with new team user assignments [SCI-6843]

okriuchykhin:
Jira ticket: [SCI-6843](https://biosistemika.atlassian.net/browse/SCI-6843)

### What was done
Update tests with new team user assignments


app/services/team_importer.rb
```diff
@@ -642,7 +642,6 @@ def create_experiment(experiment_json, project, user_id = nil)
       user_id || find_user(experiment.last_modified_by_id)
     experiment.archived_by_id = find_user(experiment.archived_by_id)
     experiment.restored_by_id = find_user(experiment.restored_by_id)
-    experiment.skip_user_assignments = true
```
aignatov-bio, line 645: I think then create_user_assignments will fail on line 649. We have validation for assignments uniqness

aignatov-bio: Same for my my_module

aignatov-bio (changes requested):

app/services/team_importer.rb
okriuchykhin: But don't we create a new set of user assignments for a new object?

aignatov-bio: In case of team import, we need to skip it. We export all assignments from the original team, they are part of JSON.

okriuchykhin: ok

aignatov-bio (approved):

PR #4179, opened 2022-06-28
Add step reordering and step element reordering service endpoints to API [SCI-6891][SCI-6892]

artoscinote:
Jira ticket: [SCI-6891](https://biosistemika.atlassian.net/browse/SCI-6891)

### What was done
Add step reordering service endpoint to API

### DIFF
--- a/app/controllers/api/service/base_controller.rb
+++ b/app/controllers/api/service/base_controller.rb
@@ -76,31 +76,6 @@ module Api
         raise PermissionError.new(Team, :read) unless can_read_team?(@team)
       end
 
-      def load_project(key = :project_id)
-        @project = @team.projects.find(params.require(key))
-        raise PermissionError.new(Project, :read) unless can_read_project?(@project)
-      end
-
-      def load_experiment(key = :experiment_id)
-        @experiment = @project.experiments.find(params.require(key))
-        raise PermissionError.new(Experiment, :read) unless can_read_experiment?(@experiment)
-      end
-
-      def load_task(key = :task_id)
-        @task = @experiment.my_modules.find(params.require(key))
-        raise PermissionError.new(MyModule, :read) unless can_read_my_module?(@task)
-      end
-
-      def load_protocol(key = :protocol_id)
-        @protocol = @task.protocols.find(params.require(key))
-        raise PermissionError.new(Protocol, :read) unless can_read_protocol_in_module?(@protocol)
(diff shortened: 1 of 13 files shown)


app/controllers/api/service/base_controller.rb
```diff
@@ -76,6 +76,31 @@ def load_team
         raise PermissionError.new(Team, :read) unless can_read_team?(@team)
       end
 
+      def load_project(key = :project_id)
```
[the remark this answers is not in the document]
artoscinote: I did this due to the permission checks, but I guess we could skip it and have the URLs be direct for each of these resources. I see that you just did permission check directly for the experiment cloning endpoint. I can rework it to match what we did with the experiment service endpoints.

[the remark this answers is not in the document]
okriuchykhin: Yes, it will be easier to call this endpoint. We can just do permission check on parent object

app/controllers/api/service/steps_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+  module Service
+    class StepsController < BaseController
+      include Api::Service::ReorderValidation
+
+      before_action :load_team, :load_project, :load_experiment, :load_task, :load_protocol
+      before_action :load_step, only: :reorder_elements
+      before_action :validate_element_order, only: :reorder_elements
+
+      def reorder_elements
+        ActiveRecord::Base.transaction do
```
okriuchykhin, line 13: Probably it is better to lock step while reordering elements

okriuchykhin, line 13: Probably it is better to lock step while reordering elements

app/controllers/api/service/protocols_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+  module Service
+    class ProtocolsController < BaseController
+      include Api::Service::ReorderValidation
+
+      before_action :load_team, :load_project, :load_experiment, :load_task
+      before_action :load_protocol, only: :reorder_steps
+      before_action :validate_step_order, only: :reorder_steps
+
+      def reorder_steps
+        ActiveRecord::Base.transaction do
```
okriuchykhin, line 13: Probably it is better to lock protocol while reordering steps

okriuchykhin (approved):

PR #4183, opened 2022-06-29
Convert team_repository to polymorphic relation [SCI-6917]

okriuchykhin:
Jira ticket: [SCI-6917](https://biosistemika.atlassian.net/browse/SCI-6917)

### What was done
Convert team_repository to polymorphic relation


config/initializers/extends.rb
```diff
@@ -409,6 +409,12 @@ class Extends
     shared_write: 2
   }.freeze
 
+  SHARED_OBJECTS_PERMISSION_LEVELS = {
+    not_shared: 0,
```
artoscinote, line 413: Why is the not_shared option needed? Can we have shared objects that are not shared?

okriuchykhin: it was originally used in repository model

okriuchykhin: removed

PR #4333, opened 2022-08-09
Hide blank table name if task is locked [SCI-7066]

artoscinote:
Jira ticket: [SCI-7066](https://biosistemika.atlassian.net/browse/SCI-7066)

### What was done
Hide blank table name if task is locked

### DIFF
--- a/app/javascript/vue/protocol/step_elements/table.vue
+++ b/app/javascript/vue/protocol/step_elements/table.vue
@@ -1,11 +1,11 @@
 <template>
-  <div class="step-table-container" :class="{ 'step-element--locked': !element.attributes.orderable.urls.update_url }">
+  <div class="step-table-container" :class="{ 'step-element--locked': locked }">
      <div class="step-element-header" :class="{ 'editing-name': editingName }">
       <div v-if="reorderElementUrl" class="step-element-grip" @click="$emit('reorder')">
         <i class="fas fas-rotated-90 fa-exchange-alt"></i>
       </div>
       <div v-else class="step-element-grip-placeholder"></div>
-      <div class="step-element-name">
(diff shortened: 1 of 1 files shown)


app/javascript/vue/protocol/step_elements/table.vue
```diff
@@ -1,11 +1,11 @@
 <template>
-  <div class="step-table-container" :class="{ 'step-element--locked': !element.attributes.orderable.urls.update_url }">
+  <div class="step-table-container" :class="{ 'step-element--locked': locked }">
      <div class="step-element-header" :class="{ 'editing-name': editingName }">
       <div v-if="reorderElementUrl" class="step-element-grip" @click="$emit('reorder')">
         <i class="fas fas-rotated-90 fa-exchange-alt"></i>
       </div>
       <div v-else class="step-element-grip-placeholder"></div>
-      <div class="step-element-name">
+      <div v-if="!locked || element.attributes.orderable.name" class="step-element-name">
```
aignatov-bio, line 8: If the task is locked you don't show table name?

aignatov-bio (changes requested):

app/javascript/vue/protocol/step_elements/table.vue
artoscinote: We don't show it if it's blank (to avoid the error field). If it's not locked (`!locked`) we do show the blank field with error, otherwise `||` we show it if name is present `element.attributes.orderable.name`.

artoscinote: Shorter, for locked tasks we show it only if name is present.

aignatov-bio (approved):

PR #4336, opened 2022-08-09
Fix result with WOPI file creation [SCI-7073]

okriuchykhin:
Jira ticket: [SCI-7073](https://biosistemika.atlassian.net/browse/SCI-7073)

### What was done
Fix result with WOPI file creation

### DIFF
--- a/app/serializers/asset_serializer.rb
+++ b/app/serializers/asset_serializer.rb
@@ -104,7 +104,6 @@ class AssetSerializer < ActiveModel::Serializer
   end
 
   def urls
-    @user = scope[:user] || @instance_options[:user]
     urls = {
       preview: asset_file_preview_path(object),
       download: rails_blob_path(object.file, disposition: 'attachment'),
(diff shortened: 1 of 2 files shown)


artoscinote (approved):

app/serializers/asset_serializer.rb
```diff
@@ -122,7 +121,7 @@ def urls
         delete: asset_destroy_path(object)
       )
     end
-    if wopi && can_manage_asset?(@user, object)
+    if wopi && can_manage_asset?(scope[:user] || @instance_options[:user], object)
       urls[:wopi_action] = object.get_action_url(@user, 'embedview')
```
aignatov-bio, line 125: @user still used

aignatov-bio (changes requested):

app/serializers/asset_serializer.rb
okriuchykhin: That is better

PR #4346, opened 2022-08-11
Added activity for update protocol name [SCI-7095]

rekonder:
Jira ticket: [SCI-7095](https://biosistemika.atlassian.net/browse/SCI-7095)

### What was done
Added activity for updating protocol name


app/controllers/my_modules_controller.rb
```diff
@@ -277,7 +277,10 @@ def protocol
 
   def update_protocol
     protocol = @my_module.protocol
+    old_protocol_name = protocol.name
     protocol.update!(protocol_params)
+
+    log_activity(:protocol_name_in_task_edited) if old_protocol_name != protocol.name
```
artoscinote, line 283: Should be wrapped in transaction, also you can probably use `protocol.name_changed?`

artoscinote (changes requested):

artoscinote (approved):

PR #4350, opened 2022-08-11
Add client for ZPL preview generation service [SCI-7049]

okriuchykhin:
Jira ticket: [SCI-7049](https://biosistemika.atlassian.net/browse/SCI-7049)

### What was done
Add client for ZPL preview generation service

### DIFF
--- a/app/services/label_templates_preview_service.rb
+++ b/app/services/label_templates_preview_service.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+class LabelTemplatesPreviewService
+  extend Service
+
+  attr_reader :error, :preview
+
+  def initialize(zpl, user, params)
+    @zpl = zpl
+    @user = user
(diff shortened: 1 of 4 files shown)


app/services/label_templates_preview_service.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+class LabelTemplatesPreviewService
+  extend Service
+
+  attr_reader :error, :preview
+
+  def initialize(label, user, params)
```
artoscinote, line 8: Could this be change to accept ZPL directly? As we have preview on the template page, and then also on the actual printing page, so in that case we accept the already rendered ZPL, which is not what's in the content part of a label.

artoscinote (changes requested):

artoscinote (approved):

PR #4373, opened 2022-08-22
Add activities for label template actions [SCI-7086]

okriuchykhin:
Jira ticket: [SCI-7086](https://biosistemika.atlassian.net/browse/SCI-7086)

### What was done
Add activities for label template actions

### DIFF
--- a/app/controllers/label_templates_controller.rb
+++ b/app/controllers/label_templates_controller.rb
@@ -33,17 +33,29 @@ class LabelTemplatesController < ApplicationController
   end
 
   def create
-    label_template = ZebraLabelTemplate.default.save!
-
-    redirect_to label_template_path(label_template, new_label: true)
+    ActiveRecord::Base.transaction do
+      label_template = ZebraLabelTemplate.default
+      label_template.team = current_team
+      label_template.save!
+      log_activity(:label_template_created, label_template)
+      redirect_to label_template_path(label_template, new_label: true)
+    end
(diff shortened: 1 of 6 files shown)


app/controllers/label_templates_controller.rb
```diff
@@ -33,17 +33,28 @@ def show
...
-
-    redirect_to label_template_path(label_template, new_label: true)
+    ActiveRecord::Base.transaction do
+      label_template = ZebraLabelTemplate.default
+      label_template.team = current_team
+      label_template.save!
+      log_activity(:label_template_created, label_template)
+      redirect_to label_template_path(label_template, new_label: true)
+    end
+  rescue StandardError => e
```
artoscinote, line 43: Should we just rescue ActiveRecord::RecordInvalid here?

artoscinote (changes requested):

app/controllers/label_templates_controller.rb
okriuchykhin: Nope, this is for general error

artoscinote: Would it make sense to at least include the stack trace then? As this can be difficult to debug if just looking at server logs.

artoscinote (approved):

app/controllers/label_templates_controller.rb
okriuchykhin: Done

PR #4388, opened 2022-08-29
Add syncing service and schedule for Fluics templates sync [SCI-7047]

okriuchykhin:
Jira ticket: [SCI-7047](https://biosistemika.atlassian.net/browse/SCI-7047)

### What was done
Add syncing service and schedule for Fluics templates sync

### DIFF
--- a/app/controllers/label_templates_controller.rb
+++ b/app/controllers/label_templates_controller.rb
@@ -100,6 +100,15 @@ class LabelTemplatesController < ApplicationController
     end
   end
 
+  def sync_fluics_templates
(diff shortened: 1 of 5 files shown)


app/controllers/label_templates_controller.rb
```diff
@@ -100,6 +100,15 @@ def zpl_preview
     end
   end
 
+  def sync_fluics_templates
```
artoscinote, line 103: Is there supposed to be a route for this, or is it part of another ticket?

okriuchykhin: It can be added with a button, later on

artoscinote (approved):

PR #4408, opened 2022-09-05
fix archived repositories show button not showing [SCI-7162]

G-Chubinidze:
Jira ticket: [SCI-7162](https://biosistemika.atlassian.net/browse/SCI-7162)

### What was done
_show repositories button fixed while user has maxed out their available repository limit and has archived them all_

### DIFF
--- a/app/views/repositories/empty_index.html.erb
+++ b/app/views/repositories/empty_index.html.erb
@@ -16,26 +16,7 @@
   </div>
 <% end %>
 
-<div class="content-pane flexible <%= params[:archived] ? :archived : :active %> repositories-index"
-      data-readonly="<%= !current_user.is_admin_of_team?(current_team) %>">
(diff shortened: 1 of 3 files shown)


app/views/repositories/empty_index.html.erb
```diff
@@ -16,6 +16,27 @@
...
+        <li class="view-switch-active" data-view-mode="archived">
+          <i class="fas fa-rocket button-icon"></i>
+          <%= t('libraries.index.switch_view.active') %>
+        </li>
+      <% end %>
+    </div>
+  </div>
+  <div class="content-body">
+  </div>
+</div>
```
artoscinote, line 38: If this was copy-pasted from somewhere, it would be better to make a partial and render the partial at both locations.

artoscinote (changes requested):

PR #4440, opened 2022-09-16
Migration fixes [SCI-7219]

artoscinote:
Jira ticket: [SCI-7219](https://biosistemika.atlassian.net/browse/SCI-7219)

### What was done
Migration fixes

### DIFF
--- a/db/migrate/20220516111152_add_team_level_permissions.rb
+++ b/db/migrate/20220516111152_add_team_level_permissions.rb
@@ -1,6 +1,18 @@
 # frozen_string_literal: true
 
 class AddTeamLevelPermissions < ActiveRecord::Migration[6.1]
+  class UserAssignment < ApplicationRecord
+    self.table_name = 'user_assignments'
+    belongs_to :assignable, polymorphic: true, touch: true
+    belongs_to :user_role
+    belongs_to :user
+    belongs_to :assigned_by, class_name: 'User', optional: true
+
+    enum assigned: { automatically: 0, manually: 1 }, _suffix: true
+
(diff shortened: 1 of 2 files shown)


db/migrate/20220516111152_add_team_level_permissions.rb
```diff
@@ -1,6 +1,18 @@
 # frozen_string_literal: true
 
 class AddTeamLevelPermissions < ActiveRecord::Migration[6.1]
+  class UserAssignment < ApplicationRecord
```
okriuchykhin, line 4: Why it is required to define this model here again?

artoscinote: Because we later added team to UserAssignments, so this migration will fail with current model, as the team_id validation will fail.

okriuchykhin: This validation?

  validates :user, uniqueness: { scope: %i(assignable team_id) }

artoscinote: This also, but `belongs_to :team` also generates a presence validator for team.

okriuchykhin: We can make uniqueness validation conditional, and always set team for belongs_to

artoscinote: We can't, because at the point of this migration, UserAssignment doesn't have team_id yet. It was added in a later migration (20220621153016_add_team_reference_to_user_assignments.rb).

okriuchykhin (approved):

PR #4471, opened 2022-09-27
Remove icon from the status repository column at label print time [SCI-7267]

rekonder:
Jira ticket: [SCI-7267](https://biosistemika.atlassian.net/browse/SCI-7267)

### What was done
Remove icon from the status column at label print time

### DIFF
--- a/app/models/repository_status_value.rb
+++ b/app/models/repository_status_value.rb
@@ -20,8 +20,6 @@ class RepositoryStatusValue < ApplicationRecord
   end
 
   def formatted_status
-    return nil unless repository_status_item
-
     repository_status_item&.status
   end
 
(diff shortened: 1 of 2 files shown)


app/models/repository_status_value.rb
```diff
@@ -19,6 +19,12 @@ def formatted
     data
   end
 
+  def formatted_status
+    return nil unless repository_status_item
+
+    repository_status_item.status
```
artoscinote, line 25: use safe navigation instead of guard clause:
`repository_status_item&.status`

artoscinote (changes requested):

artoscinote (approved):

PR #4513, opened 2022-10-10
Added titles to results page after title fix [SCI-7307]

G-Chubinidze:
Jira ticket: [SCI-7307](https://biosistemika.atlassian.net/browse/SCI-7307)

### What was done
_added titles to results page after title fix, some steps skipped such as tinymce, marvin editors..._

### DIFF
--- a/app/views/result_tables/_edit.html.erb
+++ b/app/views/result_tables/_edit.html.erb
@@ -17,7 +17,7 @@
 			</button>
 			<%= f.button t("general.save"),
 									 class: 'btn btn-primary save-result',
-									 title: 'save edited result table',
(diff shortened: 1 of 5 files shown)


app/views/result_tables/_edit.html.erb
```diff
@@ -17,7 +17,7 @@
 			</button>
 			<%= f.button t("general.save"),
 									 class: 'btn btn-primary save-result',
-									 title: 'save edited result table',
+									 title: "save edited result table #{@result.id}",
```
artoscinote, line 20: use .name instead of .id

artoscinote (changes requested):

PR #4573, opened 2022-10-25
Add ENV setting for max step text length [SCI-7377]

artoscinote:
Jira ticket: [SCI-7377](https://biosistemika.atlassian.net/browse/SCI-7377)

### What was done
Add ENV setting for max RTE text length


radiokills (commented):
this does not belong here :)

Even with a hurry, we're in, I think the logic for determination of the actual limit does not belong into model. Please check how we handle the max file size for example:

https://github.com/scinote-eln/scinote-web/blob/develop/config/application.rb#L59

radiokills (approved):

PR #4603, opened 2022-11-08
update display of the current tasks with IDs [SCI-7205]

G-Chubinidze:
Jira ticket: [SCI-7205](https://biosistemika.atlassian.net/browse/SCI-7205)

### What was done
_updated current tasks on overview page, added task code function as it was reverted_


app/models/my_module.rb
```diff
@@ -10,6 +10,9 @@ class MyModule < ApplicationRecord
   include PermissionCheckableModel
   include Assignable
 
+  ID_PREFIX = 'TA'
+  include PrefixedIdModel
```
artoscinote, line 14: The SEARCHABLE_ATTRIBUTES constant needs to be moved below this, and `PREFIXED_ID_SQL` added to it. This is necessary for global search to work properly with prefixed IDs. For reference check `app/models/experiment.rb`.

artoscinote (changes requested):

artoscinote (approved):

PR #4618, opened 2022-11-14
Add archive action to experiment table [SCI-7384]

aignatov-bio:
Jira ticket: [SCI-7384](https://biosistemika.atlassian.net/browse/SCI-7384)

### What was done
Add archive action to experiment table

### DIFF
--- a/app/assets/javascripts/experiments/table.js
+++ b/app/assets/javascripts/experiments/table.js
@@ -1,4 +1,4 @@
-/* global I18n GLOBAL_CONSTANTS InfiniteScroll filterDropdown dropdownSelector */
+/* global I18n GLOBAL_CONSTANTS InfiniteScroll filterDropdown dropdownSelector HelperModule */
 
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/experiments/table.js
```diff
@@ -42,6 +42,16 @@ var ExperimnetTable = {
       $(`<div class="table-row" data-id="${id}">${row}</div>`).appendTo(`${this.table} .table-body`);
     });
   },
+  initArchiveMyModules: function() {
+    $('#archiveTask').on('click', (e) => {
```
okriuchykhin, line 46: Shouldn't we use an event delegation for such actions?

aignatov-bio: Toolbar static, we render it on load once. I think we can assign an action to the button directly.

okriuchykhin (approved):

PR #4653, opened 2022-11-25
Implement task cloning in experiments table [SCI-7382]

artoscinote:
Jira ticket: [SCI-7382](https://scinote.atlassian.net/browse/SCI-7382)

### What was done
Implement task cloning in experiments table

### DIFF
--- a/app/assets/javascripts/experiments/table.js
+++ b/app/assets/javascripts/experiments/table.js
@@ -21,14 +21,28 @@ var ExperimnetTable = {
   },
   appendRows: function(result) {
     $.each(result, (id, data) => {
-      // Checkbox selector
-      let row = `
-        <div class="table-body-cell">
-          <div class="sci-checkbox-container">
-            <input type="checkbox" class="sci-checkbox my-module-selector" data-my-module="${id}">
-            <span class="sci-checkbox-label"></span>
-          </div>
-        </div>`;
+      let row;
+      let idColumnData = data.find((column) => column.column_type === 'id').data;
+      let provisioningStatus = idColumnData.provisioning_status;
+      let provisioningStatusUrl = idColumnData.urls.provisioning_status;
+
+      // Checkbox selector or provisioning spinner
+      if (provisioningStatus === 'in_progress') {
+        row = `
+              <div class="table-body-cell">
+                <div class="sci-checkbox-container my-module-provisioning-spinner">
+                  <div class="loading-overlay"></div>
+                </div>
+              </div>`;
+      } else {
+        row = `
+              <div class="table-body-cell">
+                <div class="sci-checkbox-container">
+                  <input type="checkbox" class="sci-checkbox my-module-selector" data-my-module="${id}">
+                  <span class="sci-checkbox-label"></span>
+                </div>
+              </div>`;
+      }
       // Task columns
       $.each(data.columns, (_i, cell) => {
         let hidden = '';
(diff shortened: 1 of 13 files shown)


app/assets/javascripts/experiments/table.js
```diff
@@ -285,15 +306,51 @@ var ExperimnetTable = {
           return { ...params, ...{ filters: this.activeFilters } };
         }
       });
+
+      this.initProvisioningStatusPolling();
     });
   },
+  initProvisioningStatusPolling: function() {
+    let provisioningStatusUrls = $('[data-urls]').toArray()
```
aignatov-bio, line 314: Maybe get provisioning rows and only URLs for this modules - `this.getUrls().provisioning_status

app/assets/javascripts/experiments/table.js
```diff
@@ -285,15 +306,51 @@ var ExperimnetTable = {
...
+          if (data.provisioning_status === 'in_progress') remainingUrls.push(url);
+        },
+        async: false
+      });
+    });
+
+    if (remainingUrls.length > 0) {
+      setTimeout(() => {
+        this.pollProvisioningStatuses(remainingUrls);
+      }, 5000);
```
aignatov-bio, line 337: Maybe more 10 seconds?

app/models/concerns/cloneable.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Cloneable
```
aignatov-bio, line 3: Did you check this helper - `module RenamingUtil`

app/assets/javascripts/experiments/table.js
```diff
@@ -21,14 +21,26 @@ var ExperimnetTable = {
...
-          <div class="sci-checkbox-container">
-            <input type="checkbox" class="sci-checkbox my-module-selector" data-my-module="${id}">
-            <span class="sci-checkbox-label"></span>
-          </div>
-        </div>`;
+      let row;
+      let provisioningStatus = data.provisioning_status;
+
+      // Checkbox selector or provisioning spinner
+      if (provisioningStatus === 'in_progress') {
```
aignatov-bio, line 28: Can we just use class to row during provisioning to hide/show required information?

aignatov-bio (changes requested):

aignatov-bio (commented):
Do we need update my_module permissions, to prevent mange during cloning?

app/models/concerns/cloneable.rb
artoscinote: That doesn't work as specified (specification is it should work the same as experiments). Also, I think that code should be either refactored, or scrapped.

app/models/concerns/cloneable.rb
```diff
@@ -0,0 +1,18 @@
...
+  extend ActiveSupport::Concern
+
+  def next_clone_name
+    raise NotImplementedError, "Cloneable model must implement the '.parent' method!" unless respond_to?(:parent)
+
+    last_clone_number =
+      parent.public_send(self.class.table_name)
+            .select("substring(#{self.class.table_name}.name, '(?:^Clone )(\\d+)')::int AS clone_number")
```
okriuchykhin, line 11: Maybe it makes sense to move "Clone" to I18N?

aignatov-bio (approved):

PR #4661, opened 2022-11-29
Add suffix for checklist and table duplication [SCI-7424]

aignatov-bio:
Jira ticket: [SCI-7424](https://scinote.atlassian.net/browse/SCI-7424)

### What was done
Add a suffix for checklist and table duplication

### DIFF
--- a/app/controllers/step_elements/checklists_controller.rb
+++ b/app/controllers/step_elements/checklists_controller.rb
@@ -43,6 +43,7 @@ module StepElements
         @step.step_orderable_elements.where('position > ?', position).order(position: :desc).each do |element|
           element.update(position: element.position + 1)
         end
+        @checklist.name += ' (1)'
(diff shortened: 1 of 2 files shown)


app/controllers/step_elements/checklists_controller.rb
```diff
@@ -43,6 +43,7 @@ def duplicate
         @step.step_orderable_elements.where('position > ?', position).order(position: :desc).each do |element|
           element.update(position: element.position + 1)
         end
+        @checklist.name += ' (1)'
```
artoscinote, line 46: This will just name the checklist the same if it's copied multiple times, so you'll have two 'Checklist (1)' checklists, named the same. Is this how it's intended to work? Shouldn't the number change to (1), (2), (3), etc. ?

artoscinote (changes requested):

artoscinote (approved):

PR #4686, opened 2022-12-05
The View flyout in the Archived tasks view [SCI-7496]

G-Chubinidze:
Jira ticket: [SCI-7496](https://scinote.atlassian.net/browse/SCI-7496)

### What was done
_added canvas view for archived tasks, removed option for going back to active tasks if theres no active task left while in archived view, updated sidebar for archived table view_

### DIFF
--- a/app/controllers/experiments_controller.rb
+++ b/app/controllers/experiments_controller.rb
@@ -95,6 +95,7 @@ class ExperimentsController < ApplicationController
     @current_sort = view_state.state.dig('my_modules', view_mode, 'sort') || 'atoz'
 
     @project = @experiment.project
+    @archived_modules = @experiment.my_modules.archived.order(:name)
     @active_modules = @experiment.my_modules.active.order(:name)
   end
 
--- a/app/controllers/experiments_controller.rb
+++ b/app/controllers/experiments_controller.rb
@@ -94,8 +94,11 @@ class ExperimentsController < ApplicationController
     @current_sort = view_state.state.dig('my_modules', view_mode, 'sort') || 'atoz'
 
     @project = @experiment.project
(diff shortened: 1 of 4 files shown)


app/controllers/experiments_controller.rb
```diff
@@ -95,6 +95,7 @@ def table
     @current_sort = view_state.state.dig('my_modules', view_mode, 'sort') || 'atoz'
 
     @project = @experiment.project
+    @archived_modules = @experiment.my_modules.archived.order(:name)
```
aignatov-bio, line 98: Why do tasks load twice?

aignatov-bio (changes requested):

app/controllers/experiments_controller.rb
G-Chubinidze: i needed access to only archived tasks in the table view, since it was displaying only active tasks even in the archived state

aignatov-bio (changes requested):

app/controllers/experiments_controller.rb
G-Chubinidze: i dont, it just was there by default, whenever we changed to archive mode, it displayed active tasks instead of archived ones, so i added archive modules variable, it gets used when were in archive mode, active_modules gets used when we're in active view mode

artoscinote: Just have a `@my_modules` variable, and load the appropriate collection based on parameter, something like:
`@my_modules = @experiment.my_modules.public_send(params[:view_mode] == "archived" ? :archived : :active).order(:name)`

aignatov-bio (approved):

PR #4687, opened 2022-12-06
BE: Column display switch (Task data display) [SCI-7442]

G-Chubinidze:
Jira ticket: [SCI-7442](https://scinote.atlassian.net/browse/SCI-7442)

### What was done
_backend for saving column visibility states for my_module table_

### DIFF
--- a/app/controllers/experiments_controller.rb
+++ b/app/controllers/experiments_controller.rb
@@ -96,7 +96,7 @@ class ExperimentsController < ApplicationController
 
     @project = @experiment.project
     @active_modules = @experiment.my_modules.active.order(:name)
-    @my_module_visible_table_columns = current_user.settings['visible_my_module_table_columns'].presence || []
+    @my_module_visible_table_columns = current_user.my_module_visible_table_columns
   end
 
   def load_table
--- a/app/controllers/experiments_controller.rb
+++ b/app/controllers/experiments_controller.rb
@@ -92,7 +92,7 @@ class ExperimentsController < ApplicationController
     redirect_to module_archive_experiment_path(@experiment) if @experiment.archived_branch?
     @project = @experiment.project
     @active_modules = @experiment.my_modules.active.order(:name)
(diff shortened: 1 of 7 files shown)


app/views/experiments/table.html.erb
```diff
@@ -23,7 +23,7 @@
           </div>
         </div>
         <% Experiments::TableViewService::COLUMNS.each do |col| %>
-          <div class="table-header-cell <%= col %>-column">
+          <div class="table-header-cell <%= col %>-column <%= 'hidden' unless col.to_s.in?(@my_module_visible_table_columns) || col == :task_name %>">
```
aignatov-bio, line 26: You just need to refresh with javascript column state.

aignatov-bio (changes requested):

app/views/experiments/table.html.erb
artoscinote: Doesn't that already happens when you switch columns on and off? It seems to work OK, this is just the initial state of the table.

app/controllers/experiments_controller.rb
```diff
@@ -96,6 +96,7 @@ def table
 
     @project = @experiment.project
     @active_modules = @experiment.my_modules.active.order(:name)
+    @my_module_visible_table_columns = current_user.settings['visible_my_module_table_columns'].presence || []
```
aignatov-bio, line 99: You meed to define default visible state, not empty array

G-Chubinidze: it has to be empty if all the columns are hidden, so the logic afterwards will hide all the columns in the backend too, only visible columns are added inside that array, if all the presence is nil, it just means that every column is hidden, not shown

aignatov-bio: So when the user first time comes to the table, will all columns be hidden?

G-Chubinidze: aha, i just checked and you're right, all columns are hidden, i will fix it

aignatov-bio (approved):

PR #4701, opened 2022-12-09
Add versions and access columns to protocols table, update protocol model [SCI-7516]

okriuchykhin:
Jira ticket: [SCI-7516](https://scinote.atlassian.net/browse/SCI-7516)

### What was done
Add versions and access columns to protocols table, update protocol model


app/datatables/protocols_datatable.rb
```diff
@@ -77,88 +80,84 @@ def new_search_condition(column, value)
...
+        DT_RowId: record.id,
+        DT_CanClone: can_clone_protocol_in_repository?(record),
+        DT_CloneUrl: if can_clone_protocol_in_repository?(record)
+                       clone_protocol_path(record, team: @team, type: @type)
+                     end,
+        DT_CanMakePrivate: can_manage_protocol_in_repository?(record),
+        DT_CanPublish: can_manage_protocol_in_repository?(record),
+        DT_CanArchive: can_manage_protocol_in_repository?(record),
+        DT_CanRestore: can_restore_protocol_in_repository?(record),
+        DT_CanExport: can_read_protocol_in_repository?(record),
```
aignatov-bio, line 94: I think we don't need it anymore, I made permissions the same as in other places. When we load permissions on row selection.

aignatov-bio (changes requested):

app/datatables/protocols_datatable.rb
okriuchykhin: OK

aignatov-bio (approved):

PR #4710, opened 2022-12-12
Text block on step without title in rearrange modal [SCI-7195]

G-Chubinidze:
Jira ticket: [SCI-7195](https://scinote.atlassian.net/browse/SCI-7195)

### What was done
_improved step rearrange modal so elements have names_

[SCI-7195]: https://scinote.atlassian.net/browse/SCI-7195?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/javascript/vue/protocol/modals/reorderable_items_modal.vue
+++ b/app/javascript/vue/protocol/modals/reorderable_items_modal.vue
@@ -23,8 +23,7 @@
               <div class="step-element-name">
                 <strong v-if="includeNumbers" class="step-element-number">{{ index + 1 }}</strong>
                 <i v-if="item.attributes.icon" class="fas" :class="item.attributes.icon"></i>
-                <span :title="item.attributes.name" v-if="item.attributes.name">{{ item.attributes.name }}</span>
-                <span :title="getTitle(item)" v-else-if="item.attributes.name == ''">{{ getTitle(item) }}</span>
+                <span :title="nameWithFallbacks(item)" v-if="nameWithFallbacks(item)">{{ nameWithFallbacks(item) }}</span>
                 <span :title="item.attributes.placeholder" v-else class="step-element-name-placeholder">{{ item.attributes.placeholder }}</span>
               </div>
             </div>
@@ -80,6 +79,9 @@
         $item_html.remove('table, img');
         let str = $item_html.text().trim();
         return str.length > 56 ? str.slice(0, 56) + "..." : str;
+      },
+      nameWithFallbacks(item) {
+        return item.attributes.name || this.getTitle(item);
       }
     }
   }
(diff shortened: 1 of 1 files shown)


app/javascript/vue/protocol/modals/reorderable_items_modal.vue
```diff
@@ -24,7 +24,7 @@
                 <strong v-if="includeNumbers" class="step-element-number">{{ index + 1 }}</strong>
                 <i v-if="item.attributes.icon" class="fas" :class="item.attributes.icon"></i>
                 <span :title="item.attributes.name" v-if="item.attributes.name">{{ item.attributes.name }}</span>
-                <span v-else class="step-element-name-placeholder">{{ item.attributes.placeholder }}</span>
+                <span v-else :title="getTitle(item)">{{ getTitle(item) }}</span>
```
aignatov-bio, line 27: We need a placeholder when the text block is empty. The name must be fixed in the serializer.

aignatov-bio (changes requested):

app/javascript/vue/protocol/modals/reorderable_items_modal.vue
artoscinote: I think we should just leave placeholder as is, and move the getTitle(item) thing to the 'name' line above it. If, after removing images and tables, the title is empty, the placeholder will kick in.

app/javascript/vue/protocol/modals/reorderable_items_modal.vue
```diff
@@ -24,7 +24,8 @@
                 <strong v-if="includeNumbers" class="step-element-number">{{ index + 1 }}</strong>
                 <i v-if="item.attributes.icon" class="fas" :class="item.attributes.icon"></i>
                 <span :title="item.attributes.name" v-if="item.attributes.name">{{ item.attributes.name }}</span>
-                <span v-else class="step-element-name-placeholder">{{ item.attributes.placeholder }}</span>
+                <span :title="getTitle(item)" v-else-if="item.attributes.name == ''">{{ getTitle(item) }}</span>
+                <span :title="item.attributes.placeholder" v-else class="step-element-name-placeholder">{{ item.attributes.placeholder }}</span>
```
artoscinote, line 28: This should be done using an expression, not as if-else statements in the template, so something like:
`:title="item.attributes.name || getTitle(item) || item.attributes.placeholder"`

artoscinote (changes requested):

app/javascript/vue/protocol/modals/reorderable_items_modal.vue
artoscinote: Maybe in a method called nameWithFallbacks, or something like that.

artoscinote (approved):

PR #4822, opened 2023-01-13
Archiving task from experiment table - connections are lost [SCI-7732]

sboursen-scinote:
Jira ticket: [SCI-7732](https://scinote.atlassian.net/browse/SCI-7732)

### What was done
I prevented connection loss after an intermediary `task` (`my_module`) is archived. Now the connection is maintained between the `src` tasks and the `dest` tasks following the same logic in `Canvas` (two-by-two connections between src tasks and dest tasks)

[SCI-7732]: https://scinote.atlassian.net/browse/SCI-7732?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/controllers/experiments_controller.rb
```diff
@@ -630,4 +633,22 @@ def sort_my_modules(records, sort)
       records
     end
   end
+
+  def generate_two_by_two_connections(src_tasks, dest_tasks)
+    return if src_tasks.empty? || dest_tasks.empty?
+
+    src_tasks.each do |src|
```
artoscinote, line 640: The MyModule model already has these relation methods defined:
```
  has_many :my_modules, through: :outputs, source: :to, class_name: 'MyModule'
  has_many :my_module_antecessors, through: :inputs, source: :from, class_name: 'MyModule'
```

so we can just use these here

app/controllers/experiments_controller.rb
```diff
@@ -517,7 +520,7 @@ def view_type_params
   def check_read_permissions
     current_team_switch(@experiment.project.team) if current_team != @experiment.project.team
     render_403 unless can_read_experiment?(@experiment) ||
-                      @experiment.archived? && can_read_archived_experiment?(@experiment)
+                      (@experiment.archived? && can_read_archived_experiment?(@experiment))
```
artoscinote, line 523: was this change intentional?

app/controllers/experiments_controller.rb
```diff
@@ -630,4 +633,22 @@ def sort_my_modules(records, sort)
       records
     end
   end
+
+  def generate_two_by_two_connections(src_tasks, dest_tasks)
```
artoscinote, line 637: should maybe be more descriptive, maybe something like `reconnect_my_modules_after_archive`

app/controllers/experiments_controller.rb
```diff
@@ -630,4 +633,22 @@ def sort_my_modules(records, sort)
...
+    return if src_tasks.empty? || dest_tasks.empty?
+
+    src_tasks.each do |src|
+      dest_tasks.each do |dest|
+        Connection.create!(input_id: dest.id, output_id: src.id)
+      end
+    end
+  end
+
+  def get_dest_my_modules(my_module)
```
artoscinote, line 647: These methods are not needed (as we already have these defined on MyModule), but as a general rule, in Ruby we don't use `get_` prefixes, as they are redundant. In this case, a more proper name would just be `destination_my_modules`.

artoscinote (changes requested):

app/controllers/experiments_controller.rb
sboursen-scinote: Hi Martin,

Thank you for mentioning that.
Will work on it right away.

app/controllers/experiments_controller.rb
sboursen-scinote: No, I think rubocop recommended it!
Should I change it back to how it was?

artoscinote (approved):

PR #4828, opened 2023-01-14
Fix: Edit task from experiment table view - Enter causes error [SCI-7745]

sboursen-scinote:
Jira ticket: [SCI-7745](https://scinote.atlassian.net/browse/SCI-7745)

### What was done

- [x] While the edit module modal is open, pressing `Enter` no longer results in an error but exhibits a behavior similar to pressing the `Confirm` button.

[SCI-7745]: https://scinote.atlassian.net/browse/SCI-7745?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/assets/javascripts/experiments/table.js
+++ b/app/assets/javascripts/experiments/table.js
@@ -464,7 +464,7 @@ var ExperimnetTable = {
   },
   clearRowTaskSelection: function() {
     this.selectedMyModules = [];
-    $('.select-all-checkboxes, .sci-checkbox').prop('checked', false);
+    $('.select-all-checkboxes .sci-checkbox').prop('checked', false);
     this.updateExperimentToolbar();
   },
   initNewTaskModal: function(table) {
--- a/app/assets/javascripts/experiments/table.js
+++ b/app/assets/javascripts/experiments/table.js
@@ -230,12 +230,12 @@ var ExperimnetTable = {
       return true;
     };
 
-    $('#modal-edit-module').on('click', 'button[data-action="confirm"]', handleRenameModal);
-
-    $('#modal-edit-module').find('form').submit((e) => {
-      e.preventDefault();
-      handleRenameModal();
-    });
+    $('#modal-edit-module')
+      .on('click', 'button[data-action="confirm"]', handleRenameModal)
+      .on('submit', 'form', (e) => {
+        e.preventDefault();
+        handleRenameModal();
+      });
   },
   initManageUsersDropdown: function() {
     $(this.table).on('show.bs.dropdown', '.assign-users-dropdown', (e) => {
--- a/app/assets/javascripts/experiments/table.js
+++ b/app/assets/javascripts/experiments/table.js
@@ -227,6 +227,10 @@ var ExperimnetTable = {
         }
       });
 
+      if ($(`.my-module-selector[data-my-module="${id}"]`).prop('checked')) {
+        $(`.my-module-selector[data-my-module="${id}"]`).click();
+      }
+
       return true;
     };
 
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/experiments/table.js
```diff
@@ -226,7 +227,18 @@ var ExperimnetTable = {
...
+      if ($(`.my-module-selector[data-my-module="${id}"]`).prop('checked')) {
+        $(`.my-module-selector[data-my-module="${id}"]`).click();
+      }
+
       return true;
+    };
+
+    $('#modal-edit-module').on('click', 'button[data-action="confirm"]', handleRenameModal);
+
+    $('#modal-edit-module').find('form').submit((e) => {
```
aignatov-bio, line 239: Can you please use delegated event and `.on` syntax? It is more readable. And you can chain it to line 237.

app/assets/javascripts/experiments/table.js
```diff
@@ -226,7 +227,18 @@ var ExperimnetTable = {
         }
       });
 
+      if ($(`.my-module-selector[data-my-module="${id}"]`).prop('checked')) {
+        $(`.my-module-selector[data-my-module="${id}"]`).click();
```
aignatov-bio, line 231: if you want clear row selection - `this.clearRowTaskSelection();`

aignatov-bio (changes requested):

app/assets/javascripts/experiments/table.js
sboursen-scinote: Hi @aignatov-bio 

Sure, good suggestion!

app/assets/javascripts/experiments/table.js
sboursen-scinote: Yeah, because the behavior of the checkboxes is different depending on whether we want to archive from the header or the task row itself.

Thanks for the feedback.

app/assets/javascripts/experiments/table.js
```diff
@@ -196,7 +196,8 @@ var ExperimnetTable = {
       $('#modal-edit-module').data('id', this.selectedMyModules[0]);
       $('#edit-module-name-input').val($(`#taskName${$('#modal-edit-module').data('id')}`).data('full-name'));
     });
-    $('#modal-edit-module').on('click', 'button[data-action="confirm"]', () => {
+
+    const handleRenameModal = () => {
       let id = $('#modal-edit-module').data('id');
       let newValue = $('#edit-module-name-input').val();
```
sboursen-scinote, line 203: Actually, I just mimicked what was done here.
After closer inspection, it seems that the `this.clearRowTaskSelection()` was not working properly, because the CSS was missing a `,`.

app/assets/javascripts/experiments/table.js
```diff
@@ -456,7 +464,7 @@ var ExperimnetTable = {
   },
   clearRowTaskSelection: function() {
     this.selectedMyModules = [];
-    $('.select-all-checkboxes .sci-checkbox').prop('checked', false);
+    $('.select-all-checkboxes, .sci-checkbox').prop('checked', false);
```
aignatov-bio, line 467: well, this line before was correct because it is only for removing the check from the checkbox at the header.
I think you need to add another line to be sure that you uncheck only the checkbox in the table. Not all checkboxes are on the page.

aignatov-bio (changes requested):

app/assets/javascripts/experiments/table.js
sboursen-scinote: Hi @aignatov-bio,
My bad!
From the naming, I thought it would clear only one row.

I suggest something like this:
- Rename the method to `clearSelectAllCheckboxes`
- Bring the selector back to ` $('.select-all-checkboxes .sci-checkbox').prop('checked', false);`
- Add another method that unchecks and updates a specific checkbox with the id `clearMyModuleSelector(myModuleId)`

What do you think?

aignatov-bio (approved):

PR #4881, opened 2023-01-24
Throw error when authenticating with revoked tokens [SCI-7631]

artoscinote:
Jira ticket: [SCI-7631](https://scinote.atlassian.net/browse/SCI-7631)

### What was done
Throw error when authenticating with revoked tokens
Added specific token revoke endpoint

[SCI-7631]: https://scinote.atlassian.net/browse/SCI-7631?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/concerns/token_authentication.rb
+++ b/app/controllers/concerns/token_authentication.rb
@@ -17,7 +17,7 @@ module TokenAuthentication
     @token = request.headers['Authorization']&.sub('Bearer ', '')
     raise JWT::VerificationError, I18n.t('api.core.missing_token') unless @token
 
-    verify_token!
+    check_token_revocation!
 
     @token_iss = Api::CoreJwt.read_iss(@token)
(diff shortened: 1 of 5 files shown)


app/controllers/concerns/token_authentication.rb
```diff
@@ -34,4 +36,10 @@ def authenticate_request!
     @current_user = User.find_by(id: payload['sub'])
     raise JWT::InvalidPayload, I18n.t('api.core.no_user_mapping') unless current_user
   end
+
+  def verify_token!
```
okriuchykhin, line 40: Maybe better call it 'check_token_revocation!' ? Because actual verification happens during token decoding.

PR #4912, opened 2023-01-31
Text block on step without title in rearrange modal [SCI-7195]

G-Chubinidze:
Jira ticket: [SCI-7195](https://scinote.atlassian.net/browse/SCI-7195)

### What was done
_simple logic update for optimization_

[SCI-7195]: https://scinote.atlassian.net/browse/SCI-7195?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/javascript/vue/protocol/modals/reorderable_items_modal.vue
+++ b/app/javascript/vue/protocol/modals/reorderable_items_modal.vue
@@ -81,7 +81,7 @@
         return str.length > 56 ? str.slice(0, 56) + "..." : str;
       },
       nameWithFallbacks(item) {
-        return item.attributes.name || this.getTitle(item);
+        return this.getTitle(item) || item.attributes.name;
       }
     }
   }


app/javascript/vue/protocol/modals/reorderable_items_modal.vue
```diff
@@ -81,7 +81,7 @@
         return str.length > 56 ? str.slice(0, 56) + "..." : str;
       },
       nameWithFallbacks(item) {
-        return item.attributes.name || this.getTitle(item);
+        return this.getTitle(item) || item.attributes.name;
```
artoscinote, line 84: This seems incorrect, as the fallback title will always get called, even if `item.attributes.name` is available.

artoscinote (changes requested):

app/javascript/vue/protocol/modals/reorderable_items_modal.vue
G-Chubinidze: it will get called, but will return false, since if item.attribute.name exists, the function returns empty string, hence the logic goes to item.attribute.name itself

artoscinote (approved):

PR #4920, opened 2023-02-01
FE manage devices (table view) + revoking the device [SCI-7633]

G-Chubinidze:
Jira ticket: [SCI-7633](https://scinote.atlassian.net/browse/SCI-7633)

### What was done
_frontend for logged devices and their revokation modal, note: we dont save the logged devices data yet so i created a dummy variable, also, the revokation functionality isn't implemented yet. just FE_

[SCI-7633]: https://scinote.atlassian.net/browse/SCI-7633?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/assets/javascripts/sitewide/device_revocation.js
+++ b/app/assets/javascripts/sitewide/device_revocation.js
@@ -1,0 +1,5 @@
+$(document).on('turbolinks:load', function() {
+  $(document).on('click', '#revokeDeviceBtn', function() {
+    $('#deviceRevocationModal').modal('show');
+  });
+});
--- a/app/assets/javascripts/sitewide/device_revocation.js
+++ b/app/assets/javascripts/sitewide/device_revocation.js
@@ -1,5 +1,0 @@
-$(document).on('turbolinks:load', function() {
-  $(document).on('click', '#revokeDeviceBtn', function() {
-    $('#deviceRevocationModal').modal('show');
-  });
-});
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/sitewide/device_revocation.js
```diff
@@ -0,0 +1,5 @@
+$(document).on('turbolinks:load', function() {
+  $(document).on('click', '#revokeDeviceBtn', function() {
+    $('#deviceRevocationModal').modal('show');
+  });
+});
```
artoscinote, line 5: Do we need this? I think for just opening the modal, we can use:
`data-toggle="modal" data-target="#revokeDeviceBtn"`
on the button.

app/views/users/registrations/edit.html.erb
```diff
@@ -70,6 +75,8 @@
 </div>
 
 <%= render partial: '2fa_modal' %>
+<%= render partial: 'device_revocation_modal' %>
```
artoscinote, line 78: I would move this partial inside the manage_devices partial, because it's required for it to work.

artoscinote (changes requested):

app/views/users/registrations/_manage_devices.html.erb
```diff
@@ -0,0 +1,46 @@
+<%# mobile_app_devices = [['Lenovo ProBook', '07/12/2022'], ['iPhone', '06/12/2022'], ['iMac', '02/12/2022'], ['Acer Air Pro', '01/12/2022']]%>
+<% mobile_app_devices = [] %>
```
rekonder, line 2: Fetch real data

rekonder (changes requested):

PR #4928, opened 2023-02-03
Step timestamp display [SCI-7837]

G-Chubinidze:
Jira ticket: [SCI-7837](https://scinote.atlassian.net/browse/SCI-7837)

### What was done
_added step timestamp_

[SCI-7837]: https://scinote.atlassian.net/browse/SCI-7837?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/serializers/step_serializer.rb
+++ b/app/serializers/step_serializer.rb
@@ -101,7 +101,7 @@ class StepSerializer < ActiveModel::Serializer
     urls_list
   end
 
-  def created_on
+  def created_at
     object.created_at.strftime('%B %d, %Y at %H:%M')
   end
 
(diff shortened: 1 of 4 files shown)


app/serializers/step_serializer.rb
```diff
@@ -99,4 +100,12 @@ def urls
 
     urls_list
   end
+
+  def created_on
```
artoscinote, line 104: Please stick to the conventions of on/at -> created_**on** is a date, here you are printing a datetime - which should be called created_**at**. Not sure about the text in the translation file, but it would also be appropriate to have it say `"Created at %{datetime} by %{user}"`. I know there are already some inconsistencies with regards to this in our database, but I wouldn't want to add more :)

artoscinote (changes requested):

artoscinote (approved):

PR #4950, opened 2023-02-08
Missing error message, when a user tries to save a protocol with an existing name [SCI-7839]

sboursen-scinote:
Jira ticket: [SCI-7839](https://scinote.atlassian.net/browse/SCI-7839)

### What was done
Implemented the error effect and message functionality when trying to rename a protocol with an existing name.
The protocol names for which the errors are shown are dependent on the `team`, `curent_user`, and the protocol `type`.

[SCI-7839]: https://scinote.atlassian.net/browse/SCI-7839?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/protocols_controller.rb
+++ b/app/controllers/protocols_controller.rb
@@ -12,6 +12,7 @@ class ProtocolsController < ApplicationController
   before_action :check_create_permissions, only: %i(
     create
   )
+  before_action :load_team_and_type, only: :show
   before_action :check_clone_permissions, only: [:clone]
   before_action :check_view_permissions, only: %i(
     show
@@ -164,6 +165,10 @@ class ProtocolsController < ApplicationController
   end
 
   def show
+    @names = Protocol.where(
+      team: @current_team,
+      protocol_type: Protocol.protocol_types[@type == :public ? :in_repository_public : :in_repository_private]
+    ).map(&:name)
     respond_to do |format|
       format.json { render json: @protocol, serializer: ProtocolSerializer, user: current_user }
       format.html
--- a/app/controllers/protocols_controller.rb
+++ b/app/controllers/protocols_controller.rb
@@ -1251,5 +1251,4 @@ class ProtocolsController < ApplicationController
       protocol_type: Protocol.protocol_types[type]
     ).map(&:name)
   end
-
 end
--- a/app/controllers/protocols_controller.rb
+++ b/app/controllers/protocols_controller.rb
@@ -12,7 +12,6 @@ class ProtocolsController < ApplicationController
   before_action :check_create_permissions, only: %i(
     create
   )
-  before_action :load_team_and_type, only: :show
   before_action :check_clone_permissions, only: [:clone]
   before_action :check_view_permissions, only: %i(
     show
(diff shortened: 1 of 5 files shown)


app/controllers/protocols_controller.rb
```diff
@@ -1241,4 +1242,13 @@ def protocol_annotation_notification(old_text)
...
     )
   end
+
+  def existing_protocol_names(protocol)
+    team = protocol.team
+    type = protocol.protocol_type.to_sym == :in_repository_private ? :in_repository_private : :in_repository_public
+    Protocol.where(
+      team: team,
+      protocol_type: Protocol.protocol_types[type]
+    ).map(&:name)
```
artoscinote, line 1252: In cases like this, use `.pluck(:name)`

artoscinote:
It would be better if this was implemented as an endpoint to check name availability (e.g. protocols/name_available), rather than always including all protocol names in the template. Could get heavy with clients that have a lot of protocols in a single team.

app/controllers/protocols_controller.rb
```diff
@@ -1241,4 +1242,13 @@ def protocol_annotation_notification(old_text)
                  protocol: link_to(@protocol.name, protocol_url(@protocol)))
     )
   end
+
+  def existing_protocol_names(protocol)
+    team = protocol.team
+    type = protocol.protocol_type.to_sym == :in_repository_private ? :in_repository_private : :in_repository_public
+    Protocol.where(
```
artoscinote, line 1249: shorter would be team.protocols.where(protocol_type...)

artoscinote (changes requested):
Please refactor to use an endpoint for protocol name availability.

aignatov-bio (changes requested):
We don't need this at all, render a validation error. Why do we need an additional method? What if we have 1000 protocols, you will always load all names?!

sboursen-scinote:
> We don't need this at all, render a validation error. Why do we need an additional method? What if we have 1000 protocols, you will always load all names?!

Hi @aignatov-bio ,
Normally that would suffice. But given the current functionality for empty names, that validation happens dynamically on the client side. That means that the error message is rendered even before the user hits "Save".

aignatov-bio (changes requested):
And meanwhile, why are we doing this ticket now? I talked with Nejc last week and was told him that this thing was replaced in a new protocol screen and already working.
Screenshot 2023-02-09 at 19 35 42

aignatov-bio:
> > We don't need this at all, render a validation error. Why do we need an additional method? What if we have 1000 protocols, you will always load all names?!
> 
> Hi @aignatov-bio , Normally that would suffice. But given the current functionality for empty names, that validation happens dynamically on the client side. That means that the error message is rendered even before the user hits "Save".

The current implementation is not reliable and just slows down the overall protocol loading speed.

sboursen-scinote:
> > > We don't need this at all, render a validation error. Why do we need an additional method? What if we have 1000 protocols, you will always load all names?!
> > 
> > 
> > Hi @aignatov-bio , Normally that would suffice. But given the current functionality for empty names, that validation happens dynamically on the client side. That means that the error message is rendered even before the user hits "Save".
> 
> The current implementation is not reliable and just slows down the overall protocol loading speed.

I mean, it depends on the UX that we would like to have.
Let's discuss this with the Product team Tomorrow :)

PR #4970, opened 2023-02-14
Archived Protocol templates - a link between task and protocol should stay unchanged (backend) [SCI-7637]

G-Chubinidze:
Jira ticket: [SCI-7637](https://scinote.atlassian.net/browse/SCI-7637)

### What was done
_changed: archiving protocols dont break link between tasks_

[SCI-7637]: https://scinote.atlassian.net/browse/SCI-7637?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


artoscinote (approved):

app/models/protocol.rb
```diff
@@ -420,6 +420,13 @@ def archive(user)
     # Update all module protocols that had
     # parent set to this protocol
     if result
+      Protocol.where(parent: self).find_each do |p|
```
aignatov-bio, line 423: Why you do this?

aignatov-bio (changes requested):

app/models/protocol.rb
G-Chubinidze: this was done on develop branch but then we changed target branch to protocol_versioning feature, need to update to match it's updated logic

PR #4993, opened 2023-02-20
Add office file to task step - reload file option shows plain html page [SCI-7673]

G-Chubinidze:
Jira ticket: [SCI-7673](https://scinote.atlassian.net/browse/SCI-7673)

### What was done
_click to reload file no longer goes to the plain html page, it was caused by remote="true" in vue.js,_

### Note:
_i added page refreshing after office step is added just in case, should that be removed? also this feature isn't tested so im not sure if it refreshes the file correctly since i cant test it on my dev instance_

[SCI-7673]: https://scinote.atlassian.net/browse/SCI-7673?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/javascript/vue/protocol/step_attachments/inline.vue
+++ b/app/javascript/vue/protocol/step_attachments/inline.vue
@@ -12,8 +12,8 @@
           :data-gallery-view-id="stepId"
           :data-preview-url="attachment.attributes.urls.preview"
         >
-          <span data-toggle="tooltip" 
-               data-placement="bottom" 
+          <span data-toggle="tooltip"
+               data-placement="bottom"
                :title="`${ attachment.attributes.file_name }`">
             {{ attachment.attributes.file_name }}
           </span>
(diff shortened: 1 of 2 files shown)


app/javascript/vue/protocol/step_attachments/inline.vue
```diff
@@ -72,6 +71,7 @@
 </template>
 
 <script>
+  import axios from 'axios'
```
aignatov-bio, line 74: Why you need this?

app/javascript/vue/protocol/step_attachments/mixins/wopi_file_modal.js
```diff
@@ -16,6 +16,7 @@ export default {
             $wopiModal.modal('hide');
             window.open(data.edit_url, '_blank');
             window.focus();
+            window.location.reload();
```
aignatov-bio, line 19: Again why this?

aignatov-bio (changes requested):

app/javascript/vue/protocol/step_attachments/mixins/wopi_file_modal.js
G-Chubinidze: in a way it removes the need of the "reload file" need, but it also negates the whole ideology of not having the need of refreshing in vue.js.
will remove it

artoscinote (approved):

PR #5009, opened 2023-02-21
Update protocol templates update action on task [SCI-7903]

rekonder:
Jira ticket: [SCI-7903](https://scinote.atlassian.net/browse/SCI-7903)

### What was done
Update protocol templates update action on task


app/models/protocol.rb
```diff
@@ -505,21 +505,28 @@ def update_parent(current_user)
     save!
   end
 
-  def update_from_parent(current_user)
+  def update_from_parent(current_user, latest_version: false)
```
aignatov-bio, line 508: Simplify it, by just pass protocol object from controller

aignatov-bio (changes requested):

aignatov-bio (approved):

PR #5035, opened 2023-02-27
Fix protocols repository table CSS [SCI-7996]

sboursen-scinote:
Jira ticket: [SCI-7996](https://scinote.atlassian.net/browse/SCI-7996)

### What was done
Updated the protocol repository index with the new styles.

[SCI-7996]: https://scinote.atlassian.net/browse/SCI-7996?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/assets/stylesheets/shared/filter_dropdown.scss
+++ b/app/assets/stylesheets/shared/filter_dropdown.scss
@@ -29,23 +29,11 @@
       position: relative;
       width: 100%;
 
-      &.folders {
-        align-items: center;
-        column-gap: 8px;
-        display: flex;
-      }
-
       label {
         @include font-small;
         font-weight: bold;
         margin-bottom: .3em;
         user-select: none;
-      }
-
-      .input-field:focus-within,
-      .datetime-picker-container:focus-within {
-        border: 1px solid $brand-focus;
-        border-radius: 4px;
       }
     }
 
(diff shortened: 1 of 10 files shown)


app/assets/stylesheets/shared/filter_dropdown.scss
```diff
@@ -29,12 +29,24 @@
       position: relative;
       width: 100%;
 
+      &.folders {
```
aignatov-bio, line 32: can you rename this class in protocol partial? And add to only protocol CSS, because this checkbox is unique for this view.

aignatov-bio: Screenshot 2023-02-28 at 15 56 42
So it will not `folder`, but `has-draft`. I forgot to update.

aignatov-bio (changes requested):

aignatov-bio (changes requested):
I think you missed that text in cells must be in a single line.
And for long names, it must be hidden with dots - `very very long nam...`

sboursen-scinote:
> I think you missed that text in cells must be in a single line. And for long names, it must be hidden with dots - `very very long nam...`
@aignatov-bio 
I don't think this is part of the current ticket.
Let me check with QA in case someone else has implemented it.

aignatov-bio (approved):

PR #5040, opened 2023-02-28
Fix version number for draft created from old versions [SCI-8023]

aignatov-bio:
Jira ticket: [SCI-8023](https://scinote.atlassian.net/browse/SCI-8023)

### What was done
Fix version number for draft created from old versions

[SCI-8023]: https://scinote.atlassian.net/browse/SCI-8023?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/protocol.rb
+++ b/app/models/protocol.rb
@@ -7,6 +7,7 @@ class Protocol < ApplicationRecord
   include PrefixedIdModel
   SEARCHABLE_ATTRIBUTES = ['protocols.name', 'protocols.description',
                            'protocols.authors', 'protocol_keywords.name', PREFIXED_ID_SQL].freeze
+  REPOSITORY_TYPES = %i(in_repository_published_original in_repository_draft in_repository_published_version).freeze
 
   include SearchableModel
   include RenamingUtil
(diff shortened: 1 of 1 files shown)


app/models/protocol.rb
```diff
@@ -323,6 +323,10 @@ def in_repository?
     in_repository_published? || in_repository_draft?
   end
 
+  def in_repository_types
+    %i(in_repository_published_original in_repository_draft in_repository_published_version)
```
artoscinote, line 327: This should probably be moved to a constant and frozen.

app/models/protocol.rb
```diff
@@ -61,6 +60,7 @@ class Protocol < ApplicationRecord
     validates :added_by, presence: true
     validates :my_module, absence: true
     validates :parent_updated_at, absence: true
+    validate :version_number_constrain
```
artoscinote, line 63: Should be spelled _version_number_constraint_.

artoscinote (changes requested):

artoscinote (approved):

PR #5053, opened 2023-03-01
Fix protocol name field editability [SCI-8033]

rekonder:
Jira ticket: [SCI-8033](https://scinote.atlassian.net/browse/SCI-8033)

### What was done
Fix protocol name field editability 

[SCI-8033]: https://scinote.atlassian.net/browse/SCI-8033?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/protocols_controller.rb
+++ b/app/controllers/protocols_controller.rb
@@ -1116,6 +1116,7 @@ class ProtocolsController < ApplicationController
   end
 
   def set_inline_name_editing
+    return unless @protocol.initial_draft?
     return unless can_manage_protocol_in_repository?(@protocol)
 
     @inline_editable_title_config = {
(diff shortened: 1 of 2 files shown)


app/controllers/protocols_controller.rb
```diff
@@ -1116,7 +1116,7 @@ def move_protocol(action)
   end
 
   def set_inline_name_editing
-    return unless can_manage_protocol_in_repository?(@protocol)
+    return unless can_manage_protocol_name_in_repository?(@protocol)
```
okriuchykhin, line 1119: I wouldn't create a separate permission helper just for this one case.
Better create a method in protocol:
  def initial_draft?
    in_repository_draft? && parent.blank?
  end

and than add it before can_manage_protocol_in_repository? check

okriuchykhin (approved):

PR #5070, opened 2023-03-03
Fix load protocol to task [SCI-8072]

rekonder:
Jira ticket: [SCI-8072](https://scinote.atlassian.net/browse/SCI-8072)

### What was done
Fix load protocol to task

[SCI-8072]: https://scinote.atlassian.net/browse/SCI-8072?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/datatables/protocols_datatable.rb
+++ b/app/datatables/protocols_datatable.rb
@@ -144,6 +144,7 @@ class ProtocolsDatatable < CustomDatatable
                                      .select(:id)
     published_versions = @team.protocols
                               .where(protocol_type: Protocol.protocol_types[:in_repository_published_version])
+                              .order(:parent_id, version_number: :desc)
                               .select('DISTINCT ON (parent_id) id')
     new_drafts = @team.protocols
                       .where(protocol_type: Protocol.protocol_types[:in_repository_draft], parent_id: nil)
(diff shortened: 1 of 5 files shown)


app/datatables/protocols_datatable.rb
```diff
@@ -144,6 +144,7 @@ def get_raw_records_base
                                      .select(:id)
     published_versions = @team.protocols
                               .where(protocol_type: Protocol.protocol_types[:in_repository_published_version])
+                              .order(:parent_id, version_number: :desc)
```
okriuchykhin, line 147: why do you need order here?

rekonder: To show the latest published version on the screen. In other cases, we can show incorrect data connected to selected records (eg. keywords,...). As I understand, DISTINCT ON selects the first row, so if we do not order before, we can not guarantee that we choose the row of the last published protocol.

okriuchykhin: Right, missed that

okriuchykhin (approved):

PR #5267, opened 2023-04-06
Bulk editing will display inaccurate changes after saving [SCI-8219]

sboursen-scinote:
Jira ticket: [SCI-8219](https://scinote.atlassian.net/browse/SCI-8219)

### What was done
Fixed race condition when submitting multiple forms `row forms` at the same time.

[SCI-8219]: https://scinote.atlassian.net/browse/SCI-8219?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/assets/javascripts/repositories/row_editor.js
+++ b/app/assets/javascripts/repositories/row_editor.js
@@ -29,34 +29,6 @@ var RepositoryDatatableRowEditor = (function() {
     let $files = $table.find('input[type=file]');
     $row.find('.has-error').removeClass('has-error').find('span').remove();
 
-    // register submits event for row forms
-    let formsCount = $forms.length;
-    $forms.each(function() {
-      const form = $(this);
-      form.off().on('submit', function(event) {
-        event.preventDefault();
-        $.ajax({
-          url: form.attr('action'),
-          type: form.attr('method'),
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/repositories/row_editor.js
```diff
@@ -22,13 +22,41 @@ var RepositoryDatatableRowEditor = (function() {
...
     let valid = true;
     let directUrl = $table.data('direct-upload-url');
     let $files = $table.find('input[type=file]');
     $row.find('.has-error').removeClass('has-error').find('span').remove();
 
+    // register submits event for row forms
+    let formsCount = $forms.length;
+    $forms.each(function() {
+      const form = $(this);
+      form.off().on('submit', function(event) {
```
okriuchykhin, line 36: any way to avoid using 'off' ?

sboursen-scinote: Not sure. We need it if we want to change data in case some forms failed in submitting.
But we actually need to unregister only the `submit` event :+1:

PR #5367, opened 2023-05-05
Implement backend for inventory item assignment to tasks [SCI-8248]

sboursen-scinote:
Jira ticket: [SCI-8248](https://scinote.atlassian.net/browse/SCI-8248)

### What was done
Created a 3-level set of filters for projects/experiments/tasks, to select a task for inventory item assignment

[SCI-8248]: https://scinote.atlassian.net/browse/SCI-8248?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/experiments_controller.rb
+++ b/app/controllers/experiments_controller.rb
@@ -9,7 +9,8 @@ class ExperimentsController < ApplicationController
 
   before_action :load_project, only: %i(new create archive_group restore_group)
   before_action :load_experiment, except: %i(new create archive_group restore_group experiment_filter)
-  before_action :check_read_permissions, except: %i(edit archive clone move new create archive_group restore_group experiment_filter)
+  before_action :check_read_permissions, except: %i(edit archive clone move new create
+                                                    archive_group restore_group experiment_filter)
   before_action :check_canvas_read_permissions, only: %i(canvas)
   before_action :check_create_permissions, only: %i(new create)
   before_action :check_manage_permissions, only: %i(edit batch_clone_my_modules)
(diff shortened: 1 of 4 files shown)


app/controllers/experiments_controller.rb
```diff
@@ -430,6 +431,18 @@ def sidebar
     end
   end
 
+  def experiment_filter
+    project = Project.readable_by_user(current_user).find_by(id: params[:project_id])
+    return render plain: [].to_json if project.blank?
```
okriuchykhin, line 436: return 404

app/controllers/my_modules_controller.rb
```diff
@@ -442,6 +442,18 @@ def provisioning_status
     render json: { provisioning_status: @my_module.provisioning_status }
   end
 
+  def my_module_filter
+    experiment = Experiment.readable_by_user(current_user).find_by(id: params[:experiment_id])
+    return render plain: [].to_json if experiment.blank?
```
okriuchykhin, line 447: return 404

app/controllers/projects_controller.rb
```diff
@@ -107,6 +107,16 @@ def sidebar
     }
   end
 
+  def project_filter
+    projects = Project.readable_by_user(current_user)
+                      .search(current_user, false, params[:query], 1, current_team)
+                      .pluck(:id, :name)
+
+    return render plain: [].to_json if projects.blank?
```
okriuchykhin, line 115: return 404

okriuchykhin (changes requested):
it has inconsistent responses, 2 endpoints return readable objects and another - managable

app/controllers/projects_controller.rb
Sboursen: I think it will be better to return and empty array in this case because that's simply the result of the search.
Same logic for seaching experiments and tasks.

okriuchykhin (approved):

PR #5368, opened 2023-05-05
Add html support to disclaimer body and enable during registration [SCI-8410][SCI-8407]

aignatov-bio:
### What was done
Add html support to disclaimer body and enable during registration


app/views/users/sessions/_login_disclaimer.html.erb
```diff
@@ -9,7 +9,7 @@
           </h4>
         </div>
         <div class="modal-body">
-          <%= login_disclaimer["body"]  %>
+          <%= login_disclaimer["body"]&.html_safe  %>
```
okriuchykhin, line 12: and what about title?

aignatov-bio: updated

okriuchykhin (approved):

PR #5388, opened 2023-05-10
Add metadata to table API endpoints [SCI-8376]

rekonder:
Jira ticket: [SCI-8376](https://scinote.atlassian.net/browse/SCI-8376)

### What was done
Add metadata to table API endpoints

[SCI-8376]: https://scinote.atlassian.net/browse/SCI-8376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/api/v1/base_controller.rb
+++ b/app/controllers/api/v1/base_controller.rb
@@ -59,6 +59,12 @@ module Api
                      :bad_request)
       end
 
+      rescue_from ActionController::BadRequest do |e|
+        render_error(
+          I18n.t('api.core.errors.parameter_incorrect.title'), e.message, :bad_request
+        )
+      end
+
       rescue_from ActionController::ParameterMissing do |e|
         render_error(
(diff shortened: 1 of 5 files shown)


app/models/table.rb
```diff
@@ -26,6 +26,7 @@ class Table < ApplicationRecord
   has_one :result, through: :result_table
   has_many :report_elements, inverse_of: :table, dependent: :destroy
 
+  before_save :convert_metadata_plate_template
```
okriuchykhin, line 29: why do you need this conversion?

rekonder: We need this conversion so we handle correctly "true" and true values for plateTemplates in case someone creates like this, so we can handle rendering from here because we expect to have this field boolean value.

okriuchykhin: why not handle this specific case in the controller?

okriuchykhin (approved):

PR #5457, opened 2023-05-22
Reports bottom action toolbar position update [SCI-8302]

G-Chubinidze:
Jira ticket: [SCI-8302](https://scinote.atlassian.net/browse/SCI-8302)

### What was done
_changed position of actions toolbar_


aignatov-bio (changes requested):
You also need add paddings to body of table otherwise action bar will cover lines, check in other table there is one line.

aignatov-bio (approved):

PR #5534, opened 2023-06-01
Fix extra large image preview for results [SCI-8585]

rekonder:
Jira ticket: [SCI-8585](https://scinote.atlassian.net/browse/SCI-8585)

### What was done
Fix extra large image preview for results

### DIFF
--- a/app/assets/javascripts/assets/asset_inline.js
+++ b/app/assets/javascripts/assets/asset_inline.js
@@ -1,7 +1,7 @@
 /* global ActiveStoragePreviews */
 
 (function() {
-  $('.attachment-preview .asset-inline-image')
+  $('.attachment-preview, .asset-inline-image')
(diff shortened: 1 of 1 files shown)


artoscinote (approved):

peteranmu:
Hello developers,

Recently we tried to build production mode on local server. After starting the server, we found that the preview of attachment images in projects was not shown (always in loading), as well as the avatar on the top right corner of web page. So we used Chrome to check the elements, and it seemed that the address of the preview image wasn't resolved correctly:

<div class="attachment-preview processing">

</div>

Should the "localhost" in "image src" be real IP+port(default3000)? We tried copying the address in browser and changing "localhost" into our server IP+port, and then the image showed.

I'm not sure whether it's a bug or I missed some config when building...

Best regards,
Peter

PR #5598, opened 2023-06-14
Activities update for grant project access to all team members [SCI-8383]

sboursen-scinote:
Jira ticket: [SCI-8383](https://scinote.atlassian.net/browse/SCI-8383)

### What was done
Added new activity types for project access and updated the filters.

[SCI-8383]: https://scinote.atlassian.net/browse/SCI-8383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


config/initializers/extends.rb
```diff
@@ -441,11 +441,13 @@ class Extends
...
     protocol_template_access_revoked_all_team_members: 240,
-    project_access_changed_all_team_members: 241
+    project_access_changed_all_team_members: 241,
+    project_grant_access_to_all_team_members: 242,
+    project_remove_access_from_all_team_members: 243
   }
 
   ACTIVITY_GROUPS = {
-    projects: [*0..7, 32, 33, 34, 95, 108, 65, 109, *158..162, 241],
+    projects: [0, 1, *3..7, 32, 33, 34, 95, 108, 65, 109, *158..162, 241, 242, 243],
```
okriuchykhin, line 450: why did you remove 2?

app/controllers/access_permissions/projects_controller.rb
```diff
@@ -68,7 +68,10 @@ def create
 
           if user_assignment_params[:user_id] == 'all'
             @project.update!(visibility: :visible, default_public_user_role_id: user_assignment_params[:user_role_id])
-            log_activity(:change_project_visibility, { visibility: t('projects.activity.visibility_visible') })
+            log_activity(:project_grant_access_to_all_team_members,
+                         { visibility: t('projects.activity.visibility_visible'),
+                           role: UserRole.find(@project.default_public_user_role_id).name,
```
okriuchykhin, line 73: Why don't you use default_public_user_role relation on project?

app/controllers/projects_controller.rb
```diff
@@ -201,15 +201,25 @@ def update
                        end
 
     default_public_user_name = nil
-    if @project.visibility_changed? && @project.default_public_user_role_id_changed?
-      default_public_user_name = UserRole.find(project_params[:default_public_user_role_id])&.name
+    if !@project.visibility_changed? && @project.default_public_user_role_id_changed?
+      default_public_user_name = UserRole.find(project_params[:default_public_user_role_id]).name
```
okriuchykhin, line 205: wrong naming, should be default_public_user_role_name

okriuchykhin (changes requested):
I think you don't cover case when project was already public and just user role was changed

app/controllers/access_permissions/projects_controller.rb
```diff
@@ -142,7 +145,10 @@ def update_default_public_user_role
           # revoke all team members access
           @project.visibility = :hidden
           @project.save!
-          log_activity(:change_project_visibility, { visibility: t('projects.activity.visibility_hidden') })
+          log_activity(:project_remove_access_from_all_team_members,
+                       { visibility: t('projects.activity.visibility_hidden'),
+                         role: UserRole.find(@project.default_public_user_role_id).name,
```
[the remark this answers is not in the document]
sboursen-scinote: The reason it works is that we are just setting visibility to `:hidden` without changing the `default_public_user_role_id`

config/initializers/extends.rb
sboursen-scinote: Because we need to 
> Remove the activity type “Project visibility changed” 
` ACTIVITY_GROUPS` is only being used to generate filters for activities.

okriuchykhin (approved):

PR #5609, opened 2023-06-16
Breadcrumbs on label/protocol template page [SCI-8476]

G-Chubinidze:
Jira ticket: [SCI-8476](https://scinote.atlassian.net/browse/SCI-8476)

### What was done
_changed breadcrumb logic for templates_

[SCI-8476]: https://scinote.atlassian.net/browse/SCI-8476?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/views/shared/navigation/_breadcrumbs.html.erb
+++ b/app/views/shared/navigation/_breadcrumbs.html.erb
@@ -44,7 +44,13 @@
       <%= last_item[:label] %>
     </span>
   <% else %>
-    <% last_item = @breadcrumbs_items.pop %>
+    <% template_paths = [/^\/protocols\/\d+$/, /^\/label_templates\/\d+$/] %>
+    <% if template_paths.any? { |pattern| request.path.match?(pattern) }%>
+     <% last_breadcrumb_items = @breadcrumbs_items.pop(2) %>
+    <% else %>
+      <% last_item = @breadcrumbs_items.pop %>
+    <% end %>
+
     <% @breadcrumbs_items.each do |item| %>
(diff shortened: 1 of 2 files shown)


artoscinote (approved):

app/views/shared/navigation/_breadcrumbs.html.erb
```diff
@@ -44,7 +44,13 @@
...
     </span>
   <% else %>
-    <% last_item = @breadcrumbs_items.pop %>
+    <% template_paths = [/^\/protocols\/\d+$/, /^\/label_templates\/\d+$/] %>
+    <% if template_paths.any? { |pattern| request.path.match?(pattern) }%>
+     <% last_breadcrumb_items = @breadcrumbs_items.pop(2) %>
+    <% else %>
+      <% last_item = @breadcrumbs_items.pop %>
+    <% end %>
+
```
sboursen-scinote, line 53: Hi
I think we can check if `item` has a `url` or not. if not we just render an simple `span` like the last item.
In that case we also add a CSS class to make it look disabled
Something like this maybe?
```
      <% if item[:url] %>
        <%= link_to(item[:label], item[:url],
                    class: "breadcrumbs-link",
                    title: item[:label]) %>
      <% else %>
        <%= content_tag(:span, item[:label],
                        class: "breadcrumbs-link",
                        title: item[:label]) %>
      <% end %>
 ```

PR #5664, opened 2023-06-23
Breadcrumbs on label/protocol template page [SCI-8476]

G-Chubinidze:
Jira ticket: [SCI-8476](https://scinote.atlassian.net/browse/SCI-8476)

### What was done
_changed logic to accomodate templates unique design_

### DIFF
--- a/app/views/shared/navigation/_breadcrumbs.html.erb
+++ b/app/views/shared/navigation/_breadcrumbs.html.erb
@@ -44,37 +44,42 @@
       <%= last_item[:label] %>
     </span>
   <% else %>
+    <% last_item = @breadcrumbs_items.pop %>
     <% template_paths = [/^\/protocols\/\d+$/, /^\/label_templates\/\d+$/] %>
-    <% if template_paths.any? { |pattern| request.path.match?(pattern) }%>
-     <% last_breadcrumb_items = @breadcrumbs_items.pop(2) %>
+
+    <% if template_paths.any? { |pattern| request.path.match?(pattern) } || ["/protocols", "/label_templates"].include?(request.path) %>
+      <span class="breadcrumbs-link" style="color: var(--sn-grey);" title="<%= @breadcrumbs_items.first[:label] %>">
+        <%= @breadcrumbs_items.first[:label] %>
+      </span>
+      <span class="delimiter">
+        <img src="<%= asset_path "icon_small/navigate_next.svg"%>"
+            alt="navigate next"
+            class="navigate_next">
+      </span>
+      <% @breadcrumbs_items.each do |item| %>
+      <%= next if @breadcrumbs_items.first == item%>
+        <%= link_to(item[:label], item[:url],
+                    class: "breadcrumbs-link",
+                    title: item[:label]) %>
+        <span class="delimiter">
+          <img src="<%= asset_path "icon_small/navigate_next.svg"%>"
+              alt="navigate next"
+              class="navigate_next">
+        </span>
+      <% end %>
     <% else %>
-      <% last_item = @breadcrumbs_items.pop %>
+      <% @breadcrumbs_items.each do |item| %>
+        <%= link_to(item[:label], item[:url],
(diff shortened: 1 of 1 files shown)


app/views/shared/navigation/_breadcrumbs.html.erb
```diff
@@ -44,37 +44,42 @@
...
         <img src="<%= asset_path "icon_small/navigate_next.svg"%>"
-             alt="navigate next"
-             class="navigate_next">
-      </span>
-      <span class="breadcrumbs-link" style="color: var(--sn-grey);" title="<%= last_breadcrumb_items.last[:label] %>">
-        <%= last_breadcrumb_items.last[:label] %>
+            alt="navigate next"
+            class="navigate_next">
       </span>
+      <% @breadcrumbs_items.each do |item| %>
```
artoscinote, line 59: fix indents

app/views/shared/navigation/_breadcrumbs.html.erb
```diff
@@ -44,37 +44,42 @@
...
+      <%= next if @breadcrumbs_items.first == item%>
```
artoscinote, line 60: missing space at the end, no need for `<%=` because we're not printing anything out, just use `<%`.

app/views/shared/navigation/_breadcrumbs.html.erb
```diff
@@ -44,37 +44,42 @@
...
-        <img src="<%= asset_path "icon_small/navigate_next.svg"%>"
-             alt="navigate next"
-             class="navigate_next">
-      </span>
-    <% end %>
-
-    <% if last_breadcrumb_items %>
-      <span class="breadcrumbs-link" style="color: var(--sn-grey);" title="<%= last_breadcrumb_items.first[:label] %>">
-        <%= last_breadcrumb_items.first[:label] %>
+    <% if template_paths.any? { |pattern| request.path.match?(pattern) } || ["/protocols", "/label_templates"].include?(request.path) %>
```
artoscinote, line 50: would it be possible to rather check params[:controller] for these cases?

artoscinote (changes requested):

app/views/shared/navigation/_breadcrumbs.html.erb
G-Chubinidze: done

app/views/shared/navigation/_breadcrumbs.html.erb
```diff
@@ -44,37 +44,41 @@
       <%= last_item[:label] %>
     </span>
   <% else %>
-    <% template_paths = [/^\/protocols\/\d+$/, /^\/label_templates\/\d+$/] %>
-    <% if template_paths.any? { |pattern| request.path.match?(pattern) }%>
-     <% last_breadcrumb_items = @breadcrumbs_items.pop(2) %>
-    <% else %>
-      <% last_item = @breadcrumbs_items.pop %>
-    <% end %>
+    <% last_item = @breadcrumbs_items.pop %>
```
sboursen-scinote, line 47: Hi
I think we can check if `item` has a `url` or not. if not we just render an simple `span` like the last item.
In that case we also add a CSS class to make it look disabled
Something like this maybe?
```
      <% if item[:url] %>
        <%= link_to(item[:label], item[:url],
                    class: "breadcrumbs-link",
                    title: item[:label]) %>
      <% else %>
        <%= content_tag(:span, item[:label],
                        class: "breadcrumbs-link",
                        title: item[:label]) %>
      <% end %>
 ```

sboursen-scinote: We also need to do the same in case the breadcrumbs are shortened `root > ... > link1 > link 2 > link 3` (with the dropdown).
[this line](https://github.com/scinote-eln/scinote-web/pull/5664/files#diff-00b0b88581e8475f6f80b9c63d977059f47b616197e601222b1b71143312c582R5)

PR #5733, opened 2023-07-07
Project card permission with not assigned team admin [SCI-8727]

G-Chubinidze:
Jira ticket: [SCI-8727](https://scinote.atlassian.net/browse/SCI-8727)

### What was done
_if user isnt assigned to project, they'll only see access button_

[SCI-8727]: https://scinote.atlassian.net/browse/SCI-8727?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/toolbars/projects_service.rb
+++ b/app/services/toolbars/projects_service.rb
@@ -13,7 +13,7 @@ module Toolbars
       @project_folders = current_user.current_team.project_folders.where(id: project_folder_ids)
 
       @items = @projects + @project_folders
-      @not_assigned = @projects.any? { |project| !project.assigned_users.include?(current_user) }
+      @not_assigned = current_user.current_team.projects.not_assigned_to_user(current_user).any?
 
       @single = @items.length == 1
 
--- a/app/services/toolbars/projects_service.rb
+++ b/app/services/toolbars/projects_service.rb
@@ -17,6 +17,9 @@ module Toolbars
 
       @single = @items.length == 1
 
+      @team_owner = @current_user.current_team.user_assignments.find { |ua| ua.user == current_user }&.user_role.owner?
+      @unassigned_team_owner = @team_owner && !can_read_project?(@items.first)
+
       @item_type = if project_ids.blank? && project_folder_ids.blank?
                      :none
                    elsif project_ids.present? && project_folder_ids.present?
@@ -30,7 +33,7 @@ module Toolbars
 
     def actions
       return [] if @item_type == :none
-      return [access_action] if @not_assigned
+      return [access_action] if @unassigned_team_owner
 
       [
         restore_action,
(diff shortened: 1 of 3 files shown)


app/services/toolbars/projects_service.rb
```diff
@@ -13,6 +13,7 @@ def initialize(current_user, project_ids: [], project_folder_ids: [])
       @project_folders = current_user.current_team.project_folders.where(id: project_folder_ids)
 
       @items = @projects + @project_folders
+      @not_assigned = @projects.any? { |project| !project.assigned_users.include?(current_user) }
```
artoscinote, line 16: This should be done as a single query with a join, since this loop will run a bunch of queries unnecessarily, something like:
```
current_user.current_team.projects.joins(
  "JOIN user_assignments ON user_assignments.assignable_type = 'Project' " \
  "AND user_assignments.assignable_id = projects.id " \
  "AND user_assignments.user_id = #{current_user.id}"
).exists?(user_assignments: { id: nil })
```

Not sure if there's a prettier or a more readable way to do it (maybe with a subquery?).

artoscinote (changes requested):

app/services/toolbars/projects_service.rb
artoscinote: Also maybe extract this to a scope on the Project model

app/models/project.rb:
```
  scope :not_assigned_to_user, lambda(user) {
    joins(
      "LEFT JOIN user_assignments ON user_assignments.assignable_type = 'Project' " \
      "AND user_assignments.assignable_id = projects.id " \
      "AND user_assignments.user_id = #{user.id}"
    ).exists?(user_assignments: { id: nil })
  }
```

app/services/toolbars/projects_service.rb
```diff
@@ -13,9 +13,13 @@ def initialize(current_user, project_ids: [], project_folder_ids: [])
       @project_folders = current_user.current_team.project_folders.where(id: project_folder_ids)
 
       @items = @projects + @project_folders
+      @not_assigned = current_user.current_team.projects.not_assigned_to_user(current_user).any?
```
artoscinote, line 16: is this still needed?

app/services/toolbars/projects_service.rb
```diff
@@ -13,9 +13,13 @@ def initialize(current_user, project_ids: [], project_folder_ids: [])
...
 
       @single = @items.length == 1
 
+      @team_owner = @current_user.current_team.user_assignments.find { |ua| ua.user == current_user }&.user_role.owner?
```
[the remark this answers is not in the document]
artoscinote: Make this a single query, not a loop through all team assignments. You can try looking through the project if we already have some team owner check somewhere, but it should be something like:
`@current_user.current_team.user_assignments.exists?(user: @current_user, role: UserRole.find_predefined_owner_role)`

PR #5763, opened 2023-07-13
Refactor html partial rendering in json responses [SCI-8811]

wandji20:
Jira ticket: [SCI-8811](https://scinote.atlassian.net/browse/SCI-8811)

### What was done
Remove redundant use of respond_to in controller actions with just one responder format and add formats to render_ro_string call with no formats

#### ToDo:
N/A

### Note:
N/A

[SCI-8811]: https://scinote.atlassian.net/browse/SCI-8811?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/my_module_tags_controller.rb
+++ b/app/controllers/my_module_tags_controller.rb
@@ -63,10 +63,8 @@ class MyModuleTagsController < ApplicationController
               my_module: my_module.id,
               tag: @mt.tag.id
             })
-    # rubocop:disable Rails/HttpStatus
     redirect_to my_module_tags_edit_path(format: :json), turbolinks: false,
-                status: 303
-    # rubocop:enable Rails/HttpStatus
+                status: :see_other
   end
 
   def destroy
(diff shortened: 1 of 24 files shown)


app/controllers/my_module_tags_controller.rb
```diff
@@ -63,13 +63,10 @@ def create
...
-
-    respond_to do |format|
-      format.json do
-        redirect_to my_module_tags_edit_path(format: :json), turbolinks: false,
-                    status: 303
-      end
-    end
+    # rubocop:disable Rails/HttpStatus
+    redirect_to my_module_tags_edit_path(format: :json), turbolinks: false,
+                status: 303
```
rekonder, line 68: Why just not use :see_other here?

app/controllers/my_module_tags_controller.rb
```diff
@@ -89,13 +86,10 @@ def destroy
...
```
rekonder, line 91: Why just not use :see_other here?

rekonder (changes requested):

app/controllers/my_module_tags_controller.rb
wandji20: @rekonder Can you please help with what **:see_other** means?

rekonder (approved):

PR #5811, opened 2023-07-19
Fix result table rendering in reports [SCI-8897]

rekonder:
Jira ticket: [SCI-8897](https://scinote.atlassian.net/browse/SCI-8897)

### What was done
Fix result table rendering in reports

[SCI-8897]: https://scinote.atlassian.net/browse/SCI-8897?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/services/reports/docx/draw_result_table.rb
+++ b/app/services/reports/docx/draw_result_table.rb
@@ -14,7 +14,7 @@ module Reports::Docx::DrawResultTable
       cell_style cols[0], bold: true, background: color[:concrete]
 
       if table.metadata.present?
-        JSON.parse(table.metadata)['cells']&.each do |cell|
+        table.metadata['cells']&.each do |cell|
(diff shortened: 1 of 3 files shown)


app/services/reports/docx/draw_result_table.rb
```diff
@@ -14,7 +14,7 @@ def draw_result_table(result)
       cell_style cols[0], bold: true, background: color[:concrete]
 
       if table.metadata.present?
-        JSON.parse(table.metadata)['cells']&.each do |cell|
+        table.metadata['cells']&.each do |cell|
```
artoscinote, line 17: I think this will fail if metadata is nil

artoscinote (changes requested):

artoscinote (approved):

PR #5850, opened 2023-07-24
Update illustrations [SCI-8842]

sboursen-scinote:
Jira ticket: [SCI-8842](https://scinote.atlassian.net/browse/SCI-8842)

### What was done
Update illustrations in Scinote.


rekonder (changes requested):
images should be in public/images folder

rekonder (approved):

PR #5886, opened 2023-07-29
Add unique database index to checklist items position

wandji20:
Jira ticket: [SCI-8841](https://scinote.atlassian.net/browse/SCI-8841)

### What was done
_Add unique database index to checklist items position and render flash message when a checklist item validation fails_

#### ToDo:
_N/A_

### Note:
_N/A_

[SCI-8841]: https://scinote.atlassian.net/browse/SCI-8841?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/api/v1/checklist_items_controller.rb
+++ b/app/controllers/api/v1/checklist_items_controller.rb
@@ -30,7 +30,7 @@ module Api
       end
 
       def update
-        @checklist_item.assign_attributes(checklist_item_params)
+        @checklist_item.assign_attributes(checklist_item_params.except(:position))
 
         if @checklist_item.changed? && @checklist_item.save!
           if @checklist_item.saved_change_to_attribute?(:checked)
--- a/app/controllers/api/v1/checklist_items_controller.rb
+++ b/app/controllers/api/v1/checklist_items_controller.rb
@@ -30,7 +30,7 @@ module Api
       end
 
       def update
-        @checklist_item.assign_attributes(checklist_item_params.except(:position))
+        @checklist_item.assign_attributes(checklist_item_params)
 
         if @checklist_item.changed? && @checklist_item.save!
           if @checklist_item.saved_change_to_attribute?(:checked)
--- a/app/javascript/vue/shared/content/checklist.vue
+++ b/app/javascript/vue/shared/content/checklist.vue
@@ -172,8 +172,8 @@
           );
 
           if(callback) callback();
-        }).error(() => {
-          HelperModule.flashAlertMsg(this.i18n.t('errors.general'), 'danger');
+        }).error((xhr) => {
+          this.setFlashErrors(xhr.responseText)
         });
 
         this.update();
@@ -190,7 +190,7 @@
               updatedItem.attributes.id = item.attributes.id
               this.$set(this.checklistItems, item.attributes.position, updatedItem)
             },
-            error: () => HelperModule.flashAlertMsg(this.i18n.t('errors.general'), 'danger')
+            error: (xhr) => this.setFlashErrors(xhr.responseText)
           });
         } else {
           // create item, then append next one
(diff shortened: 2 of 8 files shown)


db/migrate/20230728104825_add_index_to_checklist_items.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class
+ AddIndexToChecklistItems < ActiveRecord::Migration[7.0]
+  def change
+    add_index :checklist_items, :position
```
okriuchykhin, line 6: it should be unique for each checklist

app/javascript/vue/shared/content/checklist.vue
```diff
@@ -172,8 +172,8 @@
           );
 
           if(callback) callback();
-        }).error(() => {
-          HelperModule.flashAlertMsg(this.i18n.t('errors.general'), 'danger');
+        }).error((xhr) => {
+          this.setFlashErrors(xhr.responseText)
```
okriuchykhin, line 176: maybe xhr.responseJSON.errors? and then you don't need to parse it

db/migrate/20230728104825_add_index_to_checklist_items.rb
wandji20: The database uniqueness constraint for the checklist (checklist_id) is already added to the checklist table.
```
  create_table "checklist_items", force: :cascade do |t|
    t.string "text", null: false
    t.boolean "checked", default: false, null: false
    t.bigint "checklist_id", null: false
    t.datetime "created_at", precision: nil, null: false
    t.datetime "updated_at", precision: nil, null: false
    t.bigint "created_by_id"
    t.bigint "last_modified_by_id"
    t.integer "position"
    t.index "trim_html_tags((text)::text) gin_trgm_ops", name: "index_checklist_items_on_text", using: :gin
    t.index ["checklist_id"], name: "index_checklist_items_on_checklist_id"
    t.index ["created_by_id"], name: "index_checklist_items_on_created_by_id"
    t.index ["last_modified_by_id"], name: "index_checklist_items_on_last_modified_by_id"
  end
  ```

Comment (okriuchykhin) [reply db/migrate/20230728104825_add_index_to_checklist_items.rb:6]: Where? I don't see it

Comment (wandji20) [reply db/migrate/20230728104825_add_index_to_checklist_items.rb:6]: How about this
    `t.index ["checklist_id"], name: "index_checklist_items_on_checklist_id"`

Comment (okriuchykhin) [reply db/migrate/20230728104825_add_index_to_checklist_items.rb:6]: should be something like this:

t.index ["checklist_id", "position"], name: "index_checklist_items_on_checklist_id_and_position", unique: true

Comment (wandji20) [reply db/migrate/20230728104825_add_index_to_checklist_items.rb]: The unique: true will make it impossible to reorder item positions as it would raise an error (uniqueness) when reassigning positions during reordering. This could occur when reordering positions or deleting an item up the list that will trigger the rest to be reordered.

File: db/migrate/20230728104825_add_index_to_checklist_items.rb [hunk 3]
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class
+ AddIndexToChecklistItems < ActiveRecord::Migration[7.0]
+  def change
+    add_index :checklist_items, [:checklist_id, :position]
```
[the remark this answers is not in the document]
okriuchykhin: Fix this.
And you missed unique: true

app/controllers/api/v1/checklist_items_controller.rb
```diff
@@ -30,7 +30,7 @@ def create
       end
 
       def update
-        @checklist_item.assign_attributes(checklist_item_params)
+        @checklist_item.assign_attributes(checklist_item_params.except(:position))
```
okriuchykhin, line 33: no need to do this

okriuchykhin (approved):

PR #5888, opened 2023-07-31
Disable create step text using description attribute when creating a step via the API [SCI-8907]

wandji20:
Jira ticket: [SCI-8907](https://scinote.atlassian.net/browse/SCI-8907)

### What was done
Disable create step text using the params description attribute when creating a step via the API

#### ToDo:
_N/A_

### Note:
_N/A_


app/controllers/api/v1/steps_controller.rb
```diff
@@ -30,15 +30,23 @@ def show
...
-        step = @protocol.steps.create!(
+        step = @protocol.steps.build(
           step_params.except(:description)
                      .merge!(completed: false,
                              user: current_user,
                              position: @protocol.number_of_steps,
                              last_modified_by_id: current_user.id)
         )
-        step.step_texts.create!(text: step_params[:description]) if step_params[:description]
+        ActiveRecord::Base.transaction do
```
okriuchykhin, line 40: you can use @protocol.transaction

app/controllers/api/v1/steps_controller.rb
```diff
@@ -30,15 +30,23 @@ def show
 
       def create
         raise PermissionError.new(Protocol, :create) unless can_manage_protocol_in_module?(@protocol)
-
-        step = @protocol.steps.create!(
+        step = @protocol.steps.build(
```
okriuchykhin, line 33: just use create! inside the transaction block

okriuchykhin (approved):

PR #5901, opened 2023-08-02
Group of all users from team assigned to a task are not removed when their access is removed [SCI-8587]

G-Chubinidze:
Jira ticket: [SCI-8587](https://scinote.atlassian.net/browse/SCI-8587)

### What was done
_fixed access view when access is removed on task_

[SCI-8587]: https://scinote.atlassian.net/browse/SCI-8587?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/views/experiments/_assigned_users.html.erb
+++ b/app/views/experiments/_assigned_users.html.erb
@@ -1,4 +1,4 @@
-<% users = my_module.designated_users.order(:full_name) %>
+<% users = my_module.assigned_users.order(:full_name) %>
 <% if users.empty? && !can_manage_my_module_designated_users?(user, my_module) %>
   <%= I18n.t('experiments.table.not_set') %>
 <% else %>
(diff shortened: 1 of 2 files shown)


app/views/experiments/_assigned_users.html.erb
```diff
@@ -1,4 +1,4 @@
-<% users = my_module.designated_users.order(:full_name) %>
+<% users = my_module.assigned_users.order(:full_name) %>
```
artoscinote, line 1: This is not correct. Designated users is a different thing from assigned users.
What needs to be done here is to also clear the user from designated users along with the assignments.

artoscinote (changes requested):

PR #6081, opened 2023-08-25
Fix OVE CSS issues [SCI-9114]

sboursen-scinote:
Jira ticket: [SCI-9114](https://scinote.atlassian.net/browse/SCI-9114)

### What was done
- Fix broken CSS import.
- Hide body scroll when the editor is open.
- Fix editor header design.


app/assets/javascripts/sitewide/iframe_modal.js
```diff
@@ -24,4 +24,12 @@ $(document).on('turbolinks:load', function() {
...
     iFrameModal.dispatchEvent(new Event('hidden'));
   });
+
+  iFrameModal.addEventListener('shown', () => {
+    document.body.style.overflow = 'hidden';
+  });
+
+  iFrameModal.addEventListener('hidden', () => {
+    document.body.style.overflow = 'auto';
+  });
```
artoscinote, line 34: maybe it would be better to add a class to the body in this case, but this is also ok

artoscinote (approved):

PR #6101, opened 2023-08-29
Hide from annotations projects where user is not assigned [SCI-9148]

okriuchykhin:
Jira ticket: [SCI-9148](https://scinote.atlassian.net/browse/SCI-9148)

### What was done
Hide from annotations projects where user is not assigned

[SCI-9148]: https://scinote.atlassian.net/browse/SCI-9148?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -64,8 +64,12 @@ class Project < ApplicationRecord
                                 reject_if: :all_blank
 
   scope :visible_to, (lambda do |user, team|
-                        unless team.permission_granted?(user, TeamPermissions::MANAGE)
-                          left_outer_joins(user_assignments: :user_role)
+                        # Team owners see all projects in the team
+                        if team.permission_granted?(user, TeamPermissions::MANAGE)
+                          where(team: team)
+                        else
+                          where(team: team)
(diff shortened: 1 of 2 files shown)


app/models/project.rb
```diff
@@ -64,6 +64,7 @@ class Project < ApplicationRecord
                                 reject_if: :all_blank
 
   scope :visible_to, (lambda do |user, team|
+                        # Team owners see all projects in the team
```
artoscinote, line 67: In this scope, should we call `team.projects`? If we call just `Project.visible_to(user, team)`, this will return all projects, regardless of team, if the user is an owner of the team we're passing to the scope.

artoscinote (changes requested):

app/models/project.rb
okriuchykhin: Nope, that will brake chaining of other conditions

artoscinote (approved):

PR #6120, opened 2023-09-01
Placeholder text in the "Move to different result" modal dropdown [SCI-9185]

G-Chubinidze:
Jira ticket: [SCI-9185](https://scinote.atlassian.net/browse/SCI-9185)

### What was done
_included unnamed results in move modal logic_

[SCI-9185]: https://scinote.atlassian.net/browse/SCI-9185?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/javascript/vue/shared/content/modal/move.vue
+++ b/app/javascript/vue/shared/content/modal/move.vue
@@ -15,7 +15,7 @@
           <div class="w-full">
             <Select
               :value="target"
-              :options="targets"
+              :options="getOptions(targets)"
               v-bind:disabled="false"
               @change="setTarget"
             ></Select>
@@ -77,6 +77,16 @@
       },
       cancel() {
         $(this.$refs.modal).modal('hide');
+      },
+      getOptions(targets) {
+        const updatedTargets = targets.map(target => {
+          if (target[1] === null) {
+            target[2] = this.i18n.t('protocols.steps.modals.move_element.result.untitled_result');
+          }
+          return target;
+        });
+
+        return updatedTargets;
       }
     }
   }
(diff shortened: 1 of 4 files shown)


app/javascript/vue/shared/content/modal/move.vue
```diff
@@ -77,6 +77,16 @@
...
+      },
+      getOptions(targets) {
+        const updatedTargets = targets.map(target => {
+          if (target[1] === null) {
+            target[2] = this.i18n.t('protocols.steps.modals.move_element.result.untitled_result');
+          }
+          return target;
+        });
+
+        return updatedTargets;
```
artoscinote, line 89: Move this to a computed property called `targetOptions`. You can also just directly return the result of the map `return targets.map(...`.

artoscinote (changes requested):

app/javascript/vue/shared/content/modal/move.vue
```diff
@@ -77,6 +77,16 @@
       },
       cancel() {
         $(this.$refs.modal).modal('hide');
+      },
+      getOptions(targets) {
+        const updatedTargets = targets.map(target => {
+          if (target[1] === null) {
+            target[2] = this.i18n.t('protocols.steps.modals.move_element.result.untitled_result');
```
artoscinote, line 84: Don't mutate target, make a new array:

```
targets.map(target => {
  return [target[0], target[1] || this.i18n.t('protocols.steps.modals.move_element.result.untitled_result')]
});
```

PR #6272, opened 2023-09-21
Refactor checklist [SCI-9340]

aignatov-bio:
Jira ticket: [SCI-9340](https://scinote.atlassian.net/browse/SCI-9340)

### What was done
Refactor checklist

[SCI-9340]: https://scinote.atlassian.net/browse/SCI-9340?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/step_elements/checklist_items_controller.rb
+++ b/app/controllers/step_elements/checklist_items_controller.rb
@@ -45,23 +45,12 @@ module StepElements
       ActiveRecord::Base.transaction do
         new_items = @checklist_item.save_multiline!
         new_items.each_with_index do |item, i|
-          if i.zero?
-            log_activity(
-              "#{@step.protocol.in_module? ? :task : :protocol}_step_checklist_item_edited",
-              checklist_item: item.text,
-              checklist_name: @checklist.name
-            )
-            checklist_item_annotation(@step, item, old_text)
-          else
-            log_activity(
-              "#{@step.protocol.in_module? ? :task : :protocol}_step_checklist_item_added",
-              {
-                checklist_item: item.text,
-                checklist_name: @checklist.name
-              }
-            )
-            checklist_item_annotation(@step, item)
-          end
+          log_activity(
+            "#{@step.protocol.in_module? ? :task : :protocol}_step_checklist_item_#{i.zero? ? 'edited' : 'added'}",
+            checklist_item: item.text,
+            checklist_name: @checklist.name
+          )
+          checklist_item_annotation(@step, item, old_text)
         end
       end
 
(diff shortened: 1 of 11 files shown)


db/migrate/20230921142200_add_with_paragraphs_to_checklist_item.rb
```diff
@@ -0,0 +1,7 @@
+class AddWithParagraphsToChecklistItem < ActiveRecord::Migration[7.0]
+  def change
+    add_column :checklist_items, :with_paragraphs, :boolean, default: false
+
+    ChecklistItem.where("text ~ '^.*\\n.*$'").update_all(with_paragraphs: true)
```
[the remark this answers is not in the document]
artoscinote: This column is probably not needed, can't we just calculate this property from the text of the item? If it has newlines, then a method can return true, something like:

```
def has_paragraphs?
  text.include?("\n")
end
```
On the frontend we could still just set the flag when pressing Shift + Enter, to allow adding new lines.

app/models/checklist_item.rb
```diff
@@ -34,4 +73,8 @@ def touch_checklist
     checklist.touch
     # rubocop:enable Rails/SkipsModelValidations
   end
+
+  def split_multiline
```
artoscinote, line 77: maybe not needed in this case (it's pretty short)

app/controllers/step_elements/checklist_items_controller.rb
```diff
@@ -35,17 +41,31 @@ def update
...
+            checklist_item_annotation(@step, item, old_text)
+          else
+            log_activity(
+              "#{@step.protocol.in_module? ? :task : :protocol}_step_checklist_item_added",
+              {
+                checklist_item: item.text,
+                checklist_name: @checklist.name
+              }
+            )
+            checklist_item_annotation(@step, item)
```
artoscinote, line 63: Instead of the if block we could just do:
`"#{@step.protocol.in_module? ? :task : :protocol}_step_checklist_item_#{i.zero? 'edited' : 'added'}",`
and then
` checklist_item_annotation(@step, item) unless i.zero?`

app/javascript/vue/shared/content/checklist.vue
```diff
@@ -117,36 +116,37 @@
...
     },
     computed: {
       orderedChecklistItems() {
-        return this.checklistItems.map((item, index) => {
-          return { attributes: {...item.attributes, position: index } }
-        });
-      },
-      pastingMultiline() {
-        return this.linesToPaste > 0;
+        return this.checklistItems.sort((a, b) => a.attributes.position - b.attributes.position || b.id - a.id)
```
artoscinote, line 138: Not sure I understand this sort

artoscinote (changes requested):

app/javascript/vue/shared/content/checklist.vue
aignatov-bio: This for insert new element between other. new element inherit position,  this `id` sorting will verify so element will be after element. This logic to do not move other elements each time.

artoscinote (approved):

PR #6274, opened 2023-09-22
Dot divider between steps/results [SCI-9341]

G-Chubinidze:
Jira ticket: [SCI-9341](https://scinote.atlassian.net/browse/SCI-9341)

### What was done
_changed separator style_

[SCI-9341]: https://scinote.atlassian.net/browse/SCI-9341?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/assets/stylesheets/my_modules/protocol.scss
+++ b/app/assets/stylesheets/my_modules/protocol.scss
@@ -43,7 +43,12 @@
 
     &::before,
     &::after {
-      border-bottom: 1px dotted var(--sn-sleepy-grey);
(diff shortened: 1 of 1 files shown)


artoscinote:
This will need to be done using a background image, as we want to have the divider exactly as in figma, which is not achievable with only the standard CSS border.

artoscinote (changes requested):
This will need to be done using a background image, as we want to have the divider exactly as in figma, which is not achievable with only the standard CSS border.

artoscinote (approved):

PR #6364, opened 2023-10-05
Add qr code and print label to item card [SCI-9471

wandji20:
Jira ticket: [SCI-9471](https://scinote.atlassian.net/browse/SCI-9471)

### What was done
_Add qr code and print label to item card_

#### ToDo:
_N/A_

### Note:
_N/A_

[SCI-9471]: https://scinote.atlassian.net/browse/SCI-9471?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/javascript/vue/repository_item_sidebar/RepositoryItemSidebar.vue
+++ b/app/javascript/vue/repository_item_sidebar/RepositoryItemSidebar.vue
@@ -152,7 +152,7 @@
               <div class="font-inter text-base font-semibold leading-7 mb-4 mt-0">{{ i18n.t('repositories.item_card.section.qr') }}</div>
               <div class="bar-code-container">
                 <canvas id="bar-code-canvas" class="hidden"></canvas>
-                <img id="bar-code-image" />
+                <img :src="barCodeSrc" />
               </div>
             </section>
           </div>
@@ -230,6 +230,7 @@ export default {
       customColumns: null,
       assignedModules: null,
       isShowing: false,
+      barCodeSrc: null,
       permissions: null
     }
   },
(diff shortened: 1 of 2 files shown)


app/javascript/vue/repository_item_sidebar/RepositoryItemSidebar.vue
```diff
@@ -273,14 +276,13 @@ export default {
...
-      if (!text) return
-      // generate the QR code
+      if(!text) return;
       const barCodeCanvas = bwipjs.toCanvas('bar-code-canvas', {
         bcid: 'qrcode',
         text,
         scale: 3
       });
-      this.barCodeSrc = barCodeCanvas.toDataURL('image/png')
+      $('#repository-item-sidebar #bar-code-image').attr('src', barCodeCanvas.toDataURL('image/png'));
```
artoscinote, line 285: We shouldn't use jQuery if possible.
In this case, the previous implementation (``) was better. Was there a specific reason for the switch to setting the attribute via jQuery?

artoscinote (changes requested):

app/javascript/vue/repository_item_sidebar/RepositoryItemSidebar.vue
wandji20: @artoscinote 
My thought here was to avoid a rerender of the view component and use jQuery instead!
And yea, the previous implementation is better if we should avoid the use of jQuery

artoscinote (approved):

okriuchykhin (approved):

PR #6416, opened 2023-10-11
Fixed missing inventory columns on item card [SCI-9469]

lasniscinote:
Jira ticket: [SCI-9469](https://scinote.atlassian.net/browse/SCI-9469)

### What was done
- updated repository_rows jbuilder to include custom columns' cells even when they have no data
- updated en.yml
- updated view with appropriate dividers
- resolved frontend errors with optional chaining where necessary (code robustness)

[SCI-9469]: https://scinote.atlassian.net/browse/SCI-9469?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/javascript/vue/repository_item_sidebar/RepositoryItemSidebar.vue
+++ b/app/javascript/vue/repository_item_sidebar/RepositoryItemSidebar.vue
@@ -43,7 +43,7 @@
                   </span>
                 </div>
 
-                <div id="dashed-divider" class="flex h-[1px] py-0  border-dashed border-[1px] border-sn-light-grey"></div>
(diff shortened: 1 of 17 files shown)


okriuchykhin (approved):

artoscinote (changes requested):
Do not overuse the save navigation operator (`?.`). Change it so it's only used where absolutely necessary, prefer using guard clauses and the `v-if` directive in the template to prevent components from rendering if the data is not ready.

PR #6465, opened 2023-10-17
Add activities for cell editing on the new item card [SCI-9423]

rekonder:
Jira ticket: [SCI-9423](https://scinote.atlassian.net/browse/SCI-9423)

### What was done
Add activities for cell editing on the new item card

[SCI-9423]: https://scinote.atlassian.net/browse/SCI-9423?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/controllers/repository_rows_controller.rb
```diff
@@ -181,6 +183,11 @@ def update_cell
       )
 
     if row_cell_update.succeed?
+      if row_cell_update.record_updated
+        log_activity(:edit_item_field_inventory, @repository_row,
+                     { repository_row: @repository_row.id,
+                       repository_column: update_params['repository_cells'].keys.first })
```
okriuchykhin, line 189: why it is called column?

okriuchykhin: and what if repository_row.name was updated?

PR #6466, opened 2023-10-17
Group of all users from team assigned to a task are not removed when their access is removed [SCI-8587]

G-Chubinidze:
Jira ticket: [SCI-8587](https://scinote.atlassian.net/browse/SCI-8587)

### What was done
_added revoking logic for all team users_

[SCI-8587]: https://scinote.atlassian.net/browse/SCI-8587?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/access_permissions/projects_controller.rb
+++ b/app/controllers/access_permissions/projects_controller.rb
@@ -115,6 +115,7 @@ module AccessPermissions
         @project.last_modified_by = current_user
         if permitted_default_public_user_role_params[:default_public_user_role_id].blank?
           # revoke all team members access
+          remove_access_from_all_team_members
           @project.visibility = :hidden
           previous_user_role_name = @project.default_public_user_role.name
           @project.default_public_user_role_id = nil
@@ -159,6 +160,19 @@ module AccessPermissions
       @project = current_team.projects.includes(user_assignments: %i(user user_role)).find_by(id: params[:id])
 
       render_404 unless @project
+    end
+
+    def remove_access_from_all_team_members
+      users = @project.assigned_users.reject { |user| user == current_user }
+      users.each do |user|
+        user_assignment = @project.user_assignments.find_by(user: user, team: current_team)
+
+        if user_assignment.last_with_permission?(ProjectPermissions::USERS_MANAGE, assigned: :manually)
+          raise ActiveRecord::RecordInvalid
+        end
+
+        propagate_job(user_assignment, destroy: true)
+      end
     end
 
     def propagate_job(user_assignment, destroy: false)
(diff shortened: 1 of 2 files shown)


app/controllers/access_permissions/projects_controller.rb
```diff
@@ -161,6 +162,19 @@ def set_project
...
+      users.each do |user|
+        user_assignment = @project.user_assignments.find_by(user: user, team: current_team)
+
+        if user_assignment.last_with_permission?(ProjectPermissions::USERS_MANAGE, assigned: :manually)
+          raise ActiveRecord::RecordInvalid
+        end
+
+        propagate_job(user_assignment, destroy: true)
+      end
+    end
```
artoscinote, line 176: This logic should go into the Project model, so this happens any time the project visibility is updated.

artoscinote (changes requested):

app/jobs/user_assignments/project_group_un_assignment_job.rb
```diff
@@ -11,6 +11,7 @@ def perform(project)
 
         remove_users_from_my_modules(project, users)
         remove_users_from_experiments(project, users)
+        remove_users_from_projects(project, users)
```
okriuchykhin, line 14: should be remove_users_from_project, as we have here only one project

app/jobs/user_assignments/project_group_un_assignment_job.rb
```diff
@@ -32,5 +33,27 @@ def remove_users_from_my_modules(project, users)
       my_modules.update_all(updated_at: Time.current)
       # rubocop:enable Rails/SkipsModelValidations
     end
+
+    def remove_users_from_projects(project, users)
+      users.each do |user|
+        user_assignment = project.user_assignments.find_by(user: user)
+
+        if user_assignment.last_with_permission?(ProjectPermissions::USERS_MANAGE, assigned: :manually)
```
okriuchykhin, line 41: not only manually assigned, but any

app/jobs/user_assignments/project_group_un_assignment_job.rb
```diff
@@ -32,5 +33,27 @@ def remove_users_from_my_modules(project, users)
...
+          raise ActiveRecord::RecordInvalid
+        end
+
+        propagate_job(project, user_assignment, destroy: true)
+      end
+    end
+
+    def propagate_job(project, user_assignment, destroy: false)
```
okriuchykhin, line 49: why do we need to propagate if we already removed the user from tasks and experiments?

G-Chubinidze: i tried removing assignments the same way it is for experiments/modules but then it just stuck in a loop, im not sure why, but this way works 100%

PR #6561, opened 2023-11-02
Implement Task Due Date Reminder notification [SCI-9562]

G-Chubinidze:
Jira ticket: [SCI-9562](https://scinote.atlassian.net/browse/SCI-9562)

### What was done
_implemented reminder_

[SCI-9562]: https://scinote.atlassian.net/browse/SCI-9562?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/my_modules_controller.rb
+++ b/app/controllers/my_modules_controller.rb
@@ -187,10 +187,6 @@ class MyModulesController < ApplicationController
         log_activity(:rename_task) if name_changed
         log_start_date_change_activity(start_date_changes) if start_date_changes.present?
         log_due_date_change_activity(due_date_changes) if due_date_changes.present?
-
-        if due_date_changes
-          @my_module.update(notification_sent: false)
-        end
       end
     end
     if saved
--- a/app/controllers/my_modules_controller.rb
+++ b/app/controllers/my_modules_controller.rb
@@ -187,6 +187,10 @@ class MyModulesController < ApplicationController
         log_activity(:rename_task) if name_changed
         log_start_date_change_activity(start_date_changes) if start_date_changes.present?
         log_due_date_change_activity(due_date_changes) if due_date_changes.present?
+
+        if due_date_changes
+          @my_module.update(notification_sent: false)
+        end
       end
     end
     if saved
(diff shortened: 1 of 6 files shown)


app/jobs/my_modules/due_date_reminder_job.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class DueDateReminderJob < ApplicationJob
+  def perform
+    my_modules = MyModule.approaching_due_dates
+
+    my_modules.each do |task|
+      TaskDueDateNotification.send_notifications({ my_module_id: task.id })
+      task.update(notification_sent: true)
```
artoscinote, line 9: move this notification_sent logic to the notification class, there should be some after_send hook or something like that, check noticed gem docs

app/controllers/my_modules_controller.rb
```diff
@@ -187,6 +187,10 @@ def update
         log_activity(:rename_task) if name_changed
         log_start_date_change_activity(start_date_changes) if start_date_changes.present?
         log_due_date_change_activity(due_date_changes) if due_date_changes.present?
+
+        if due_date_changes
```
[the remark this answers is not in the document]
artoscinote: Move this logic to the model. I suggest into before_save hook, where you can then set `self.notification_sent = false` before saving

db/migrate/20231029231440_add_task_due_date_reminder_notification.rb
```diff
@@ -0,0 +1,5 @@
+class AddTaskDueDateReminderNotification < ActiveRecord::Migration[7.0]
+  def change
+    add_column :my_modules, :notification_sent, :boolean, default: false
```
artoscinote, line 3: rename this to `:due_date_notification_sent`, as we might have others in the future

app/models/my_module.rb
```diff
@@ -139,6 +139,10 @@ def self.filter_by_teams(teams = [])
     joins(experiment: :project).where(experiment: { projects: { team: teams } })
   end
 
+  def self.approaching_due_dates
+    where(notification_sent: false).select { |task| task.is_one_day_prior? }
```
artoscinote, line 143: Don't use .select, try and implement this using a database query (`due_date_at - NOW() < ?...` something like that)

artoscinote (changes requested):

artoscinote (approved):

PR #6628, opened 2023-11-12
Migrate smart annotation notifications [SCI-9605]

G-Chubinidze:
Jira ticket: [SCI-9605](https://scinote.atlassian.net/browse/SCI-9605)

### What was done
_added notification class_

### DIFF
--- a/app/notifications/recipients/annotated_recipients.rb
+++ b/app/notifications/recipients/annotated_recipients.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class Recipients::AnnotatedRecipients
+  def initialize(params)
+    @params = params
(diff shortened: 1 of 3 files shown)


app/notifications/recipients/annotated_recipients.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class Recipients::AnnotatedRecipients
+  def initialize(params)
+    @params = params
+  end
+
+  def recipients; end
```
artoscinote, line 8: This whole class seems redundant. We either need to properly migrate this to use the class, or just use the target_user from before (or put it in params or something).

PR #6713, opened 2023-11-27
TypeErrors are logged when a Task is moved to In review [SCI-9753]

G-Chubinidze:
Jira ticket: [SCI-9753](https://scinote.atlassian.net/browse/SCI-9753)

### What was done
_type error fix_


artoscinote:
This was more properly fixed by another PR. The issue was that the datepicker for due date was being initialized on locked tasks, which do not include this date field.

PR #6736, opened 2023-11-29
Implement unlinking on inventory unsharing [SCI-9714]

sboursen-scinote:
Jira ticket: [SCI-9714](https://scinote.atlassian.net/browse/SCI-9714)

### What was done
Implement unlinking on inventory unsharing.

[SCI-9714]: https://scinote.atlassian.net/browse/SCI-9714?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -37,6 +37,7 @@ class Repository < RepositoryBase
   after_save :assign_globally_shared_inventories, if: -> { saved_change_to_permission_level? && globally_shared? }
   after_save :unassign_globally_shared_inventories, if: -> { saved_change_to_permission_level? && !globally_shared? }
   after_save :unassign_unshared_items, if: :saved_change_to_permission_level
+  after_save :unlink_unshared_items, if: -> { saved_change_to_permission_level? && !globally_shared? }
 
   validates :name,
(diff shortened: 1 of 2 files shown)


app/models/repository.rb
```diff
@@ -249,6 +250,19 @@ def unassign_unshared_items
                          .destroy_all
   end
 
+  def unlink_unshared_items
+    RepositoryRowConnection.transaction do
```
okriuchykhin, line 254: it should be already inside a transaction

app/models/repository.rb
```diff
@@ -249,6 +250,19 @@ def unassign_unshared_items
...
+      repository_rows_to_unlink = RepositoryRow.joins(:repository)
```
okriuchykhin, line 255: So basically here all other rows get selected? It's better to limit the scope to only needed rows

okriuchykhin (changes requested):
Check comments

okriuchykhin (approved):

PR #6876, opened 2024-01-04
(fix) Errors in system log [SCI-9922]

lasniscinote:
Jira ticket: [SCI-9922](https://scinote.atlassian.net/browse/SCI-9922)

### What was done
Fixed undefined method 'label_templates' for nil:NilClass error

[SCI-9922]: https://scinote.atlassian.net/browse/SCI-9922?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/label_templates_controller.rb
+++ b/app/controllers/label_templates_controller.rb
@@ -174,7 +174,11 @@ class LabelTemplatesController < ApplicationController
   end
 
   def load_label_templates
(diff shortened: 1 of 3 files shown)


lasniscinote:
After not being able to reproduce the error and after a short deliberation with Arto and Darja, it was decided that the error specified in the ticket was too vague and edge-case to warrant this pull request. The solution provided here would only mask the problem.

PR #6894, opened 2024-01-05
Change default value for table metadata [SCI-9932]

rekonder:
Jira ticket: [SCI-9932](https://scinote.atlassian.net/browse/SCI-9932)

### What was done
Change default value for table metadata


okriuchykhin (changes requested):
also update schema.rb

okriuchykhin (approved):

PR #7069, opened 2024-02-07
Implement sorting on all columns of tables revamped in v.1 [SCI-10100]

G-Chubinidze:
Jira ticket: [SCI-101000](https://scinote.atlassian.net/browse/SCI-10100)

### What was done
_added sorting_

#### ToDo:
_Was not able to add sorting for shared state_


app/services/lists/my_modules_service.rb
```diff
@@ -70,7 +76,41 @@ def sort_records
...
+      when 'age_ASC'
+        @records = @records.order(:created_at)
+      when 'age_DESC'
+        @records = @records.order(created_at: :desc)
+      when 'status_ASC'
+        @records = @records.order(:my_module_status_id)
+      when 'status_DESC'
+        @records = @records.order(my_module_status_id: :desc)
+      when 'designated_ASC'
+        @records = @records.left_joins(:user_my_modules)
```
artoscinote, line 88: Check if distinct is needed here to avoid duplicates caused by join.

In this case, there might be duplicates if there are multiple designated users.

artoscinote (changes requested):

app/services/lists/repositories_service.rb
```diff
@@ -19,6 +19,7 @@ def fetch_records
                           .select('teams.name AS team_name')
                           .select('creators.full_name AS created_by_user')
                           .select('archivers.full_name AS archived_by_user')
+                          .select('repositories.id as code')
```
aignatov-bio, line 22: ID is not a Code, why you added this?

aignatov-bio (changes requested):

app/services/lists/repositories_service.rb
```diff
@@ -19,6 +19,7 @@ def fetch_records
                           .select('teams.name AS team_name')
                           .select('creators.full_name AS created_by_user')
                           .select('archivers.full_name AS archived_by_user')
+                          .select('repositories.id')
```
aignatov-bio, line 22: We already have this

aignatov-bio (changes requested):

app/services/lists/repositories_service.rb
G-Chubinidze: @aignatov-bio when i opened this pr it wasn't there, that's why the code conflict.

artoscinote (approved):

PR #7117, opened 2024-02-20
TypeErrors when creating a new task [SCI-10053]

sboursen-scinote:
Jira ticket: [SCI-10053](https://scinote.atlassian.net/browse/SCI-10053)

### What was done
Mount due date date picker without turbolinks in experiment table.

basically in experiments, when we do a row action, we reload all rows from backend, creating thus vue components for each due date cell. We should clean up each time and also there is no need to mount with turbolinks.

[SCI-10053]: https://scinote.atlassian.net/browse/SCI-10053?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/assets/javascripts/experiments/table.js
+++ b/app/assets/javascripts/experiments/table.js
@@ -2,6 +2,7 @@
           initBSTooltips filterDropdown dropdownSelector Sidebar HelperModule notTurbolinksPreview _ */
 
 var ExperimnetTable = {
+  mountedDueDateVueApps: [],
   selectedId: [],
   table: '.experiment-table',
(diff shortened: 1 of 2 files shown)


sboursen-scinote:
I think my fix for this ticket is not necessary anymore (PR is not yet merged)
It is simply because the table now is a Vue component
so loading data doesn't leave orphan due date vue components

I tested it and the issue doesn't happen anymore

PR #7132, opened 2024-02-22
Add missing (you) mark for current user in Grant Access modal [SCI-10273]

sboursen-scinote:
Jira ticket: [SCI-10273](https://scinote.atlassian.net/browse/SCI-10273)

### What was done
Add missing (you) mark for current user in Grant Access modal.


aignatov-bio (changes requested):
i think for this case we need some other solution, this will be generic user serializer. Maybe just add flag - current_user. And add on front-end. Because user assignment serializer more unique

aignatov-bio (approved):

PR #7158, opened 2024-02-28
Fix issue moving experiments [SCI-10292]

wandji20:
Jira ticket: [SCI-10292](https://scinote.atlassian.net/browse/SCI-10292)

### What was done
_Fix issue moving experiments_

#### ToDo:
_N/A_

### Note:
_N/A_

[SCI-10292]: https://scinote.atlassian.net/browse/SCI-10292?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/experiments_controller.rb
+++ b/app/controllers/experiments_controller.rb
@@ -11,6 +11,7 @@ class ExperimentsController < ApplicationController
   before_action :load_project, only: %i(index create archive_group restore_group move)
   before_action :load_experiment, except: %i(create archive_group restore_group
                                              inventory_assigning_experiment_filter actions_toolbar index move)
+  before_action :load_experiments, only: :move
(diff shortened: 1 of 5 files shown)


aignatov-bio (commented):
You can make just final url in action toolbar service. I found this way cleaner in protocols. 
```
def export_action
      return unless @single

      return unless can_read_protocol_in_repository?(@protocols.first)

      {
        name: 'export',
        label: I18n.t('protocols.index.toolbar.export'),
        icon: 'sn-icon sn-icon-export',
        path: export_protocols_path(protocol_ids: @protocols.pluck(:id)),
        type: :emit
      }
    end
```
And why we need redirect to project?

aignatov-bio (approved):

PR #7174, opened 2024-03-01
Change cursor type for inline editing [SCI-10264]

sboursen-scinote:
Jira ticket: [SCI-10264
](https://scinote.atlassian.net/browse/SCI-10264)

### What was done
Change cursor type for inline editing.


aignatov-bio (changes requested):
Must be - `cursor: text`

PR #7333, opened 2024-03-20
Replace state and view dropdowns in experiment canvas page [SCI-10445]

wandji20:
Jira ticket: [SCI-10445](https://scinote.atlassian.net/browse/SCI-10445)

### What was done
_Replace state and view dropdowns in experiment canvas page_

#### ToDo:
_N/A_

### Note:
_N/A_

[SCI-10445]: https://scinote.atlassian.net/browse/SCI-10445?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/javascript/packs/vue/legacy/task_canvas_menu_dropdowns.js
+++ b/app/javascript/packs/vue/legacy/task_canvas_menu_dropdowns.js
@@ -1,0 +1,7 @@
+import { createApp } from 'vue/dist/vue.esm-bundler.js';
+import TaskCanvasMenuDropdowns from '../../../vue/my_modules/state_menu_dropdowns.vue';
+import { mountWithTurbolinks } from '../helpers/turbolinks.js';
+
+const app = createApp({});
+app.component('TaskCanvasMenuDropdowns', TaskCanvasMenuDropdowns);
+mountWithTurbolinks(app, '#taskCanvasDropdowns');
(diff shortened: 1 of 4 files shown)


app/controllers/navigator/base_controller.rb
```diff
@@ -180,7 +180,7 @@ def build_folder_tree(folder, children)
       archived = params[:archived] == 'true'
       tree = fetch_projects(folder.parent_folder, archived).map { |i| project_serializer(i, archived) } +
              fetch_project_folders(folder.parent_folder, archived).map { |i| folder_serializer(i, archived) }
-      tree.find { |i| i[:id] == folder.code }[:children] = children
+      tree.find { |i| i[:id] == folder.code } && tree.find { |i| i[:id] == folder.code }[:children] = children
```
aignatov-bio, line 183: Did you check why this issue appears at all? Theoretically we build tree level based on parent, and folder must exist in this level.

aignatov-bio (changes requested):

app/controllers/navigator/base_controller.rb
wandji20: @aignatov-bio I noticed there is a 500 error caused in this line when viewing tasks in archived state.
This happens when task project folder is active
So query to search project folders will not have the current task project folder in it.

aignatov-bio: but then probably you will lost some folder will be missed in list. So if we inside archived task we still must see whole tree even if this archived

PR #7791, opened 2024-08-08
Prevent repository access to users without repository READ permission on team level [SCI-10972]

artoscinote:
Jira ticket: [SCI-10972](https://scinote.atlassian.net/browse/SCI-10972)

### What was done
fixed repository access
global search fixes
activity fixes

[SCI-10972]: https://scinote.atlassian.net/browse/SCI-10972?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/asset.rb
+++ b/app/models/asset.rb
@@ -63,7 +63,7 @@ class Asset < ApplicationRecord
     options = {}
   )
 
-    teams = options[:teams] || current_team || user.teams.select(:id)
+    teams = options[:teams] || [current_team] || user.teams
 
     assets_in_steps = Asset.joins(:step)
                            .where(steps: { protocol: Protocol.search(user, include_archived, nil, teams) })
@@ -74,7 +74,7 @@ class Asset < ApplicationRecord
                              .pluck(:id)
 
     assets_in_inventories = Asset.joins(repository_cell: { repository_column: :repository })
-                                 .where(repositories: { team: teams })
+                                 .where(repositories: { team: teams.filter { |t| t.permission_granted?(user, RepositoryPermissions::READ) } })
                                  .where.not(repositories: { type: 'RepositorySnapshot' })
                                  .pluck(:id)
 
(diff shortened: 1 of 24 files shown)


app/models/protocol.rb
```diff
@@ -254,11 +254,7 @@ def self.viewable_by_user(user, teams, options = {})
   end
 
   def self.viewable_by_user_my_module_protocols(user, teams)
```
okriuchykhin, line 256: We don't go below tasks with permissions, so it is not possible to assume that such protocols have user assignments. 
And this method is for linked to tasks protocols

app/models/asset.rb
```diff
@@ -63,7 +63,7 @@ def self.search(
     options = {}
   )
 
-    teams = options[:teams] || current_team || user.teams.select(:id)
+    teams = options[:teams] || [current_team] || user.teams
```
aignatov-bio, line 66: `options[:teams]` always return array?

app/models/repository_row.rb
```diff
@@ -128,7 +128,10 @@ def self.search(user,
                   _current_team = nil,
                   options = {})
 
-    teams = options[:teams] || current_team || user.teams.select(:id)
+    teams = (options[:teams] || [current_team] || user.teams.select(:id)).filter do |team|
```
aignatov-bio, line 131: Same here

aignatov-bio (changes requested):
Will we need any changes in addons?

app/models/asset.rb
```diff
@@ -74,7 +74,7 @@ def self.search(
                              .pluck(:id)
 
     assets_in_inventories = Asset.joins(repository_cell: { repository_column: :repository })
-                                 .where(repositories: { team: teams })
+                                 .where(repositories: { team: teams.filter { |t| t.permission_granted?(user, RepositoryPermissions::READ) } })
```
okriuchykhin, line 77: Why we can't use a subquery here?

okriuchykhin: And actually shouldn't we simply check RepositoryPermissions::READ on repositories?

PR #8106, opened 2024-12-17
Add form response models and controllers

artoscinote:
Jira ticket: [SCI-11356](https://scinote.atlassian.net/browse/SCI-11356)

### What was done

[SCI-11356]: https://scinote.atlassian.net/browse/SCI-11356?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/form_response.rb
+++ b/app/models/form_response.rb
@@ -1,0 +1,72 @@
+# frozen_string_literal: true
+
+class InvalidStatusError < StandardError; end
+
+class FormResponse < ApplicationRecord
+  include Discard::Model
+
+  default_scope -> { kept }
+
+  belongs_to :form
+  belongs_to :created_by, class_name: 'User'
+  belongs_to :submitted_by, class_name: 'User', optional: true
+
+  has_one :step_orderable_element, as: :orderable, dependent: :destroy
+
+  enum :status, { pending: 0, submitted: 1, locked: 2 }
+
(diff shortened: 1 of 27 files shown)


config/initializers/extends/permission_extends.rb
```diff
@@ -44,6 +44,15 @@ module FormPermissions
     ).each { |permission| const_set(permission, "form_#{permission.parameterize}") }
   end
 
+  module FormResponsePermissions
```
andrej-scinote, line 47: This permission group has not been added to the OWNER_PERMISSIONS below. 

It is missing migration for old already existing user roles in the database.

app/models/form_response.rb
```diff
@@ -0,0 +1,75 @@
...
+  has_many :form_field_values, dependent: :destroy
+
+  def step
+    step_orderable_element&.step
+  end
+
+  def parent
+    if step_orderable_element
+      protocol = step_orderable_element.step.protocol
+      protocol.my_module || protocol
```
aignatov-bio, line 27: I think parent must be step. Why we check it is my_module or protoco?
In controller when we used word parent it reference to Step

aignatov-bio (changes requested):

aignatov-bio (approved):

PR #8287, opened 2025-03-04
Add field duplicate option [SCI-11345]

aignatov-bio:
Jira ticket: [SCI-11345](https://scinote.atlassian.net/browse/SCI-11345)

### What was done
Add field duplicate option

[SCI-11345]: https://scinote.atlassian.net/browse/SCI-11345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/form_fields_controller.rb
+++ b/app/controllers/form_fields_controller.rb
@@ -3,8 +3,8 @@
 class FormFieldsController < ApplicationController
   before_action :check_forms_enabled
   before_action :load_form
-  before_action :load_form_field, only: %i(update destroy)
-  before_action :check_manage_permissions, only: %i(create update destroy reorder)
+  before_action :load_form_field, only: %i(update destroy duplicate)
+  before_action :check_manage_permissions, only: %i(create update destroy reorder duplicate)
 
   def create
     ActiveRecord::Base.transaction do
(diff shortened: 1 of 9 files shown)


app/controllers/form_fields_controller.rb
```diff
@@ -39,6 +39,18 @@ def update
     end
   end
 
+  def duplicate
+    ActiveRecord::Base.transaction do
```
okriuchykhin, line 43: What's the point in transaction if it won't be reverted in case of error?

app/controllers/form_fields_controller.rb
```diff
@@ -39,6 +39,17 @@ def update
...
+      new_form_field = @form_field.duplicate(current_user)
```
okriuchykhin, line 44: renmae it to .duplicate!

app/controllers/form_fields_controller.rb
```diff
@@ -39,6 +39,17 @@ def update
...
+      log_activity(:form_block_duplicated, block_name: @form_field.name)
+      render json: new_form_field, serializer: FormFieldSerializer, user: current_user
+    rescue ActiveRecord::RecordInvalid
```
okriuchykhin, line 47: what about other types of exceptions?

app/models/form_field.rb
```diff
@@ -11,4 +13,20 @@ class FormField < ApplicationRecord
...
+
+  def parent
+    form
+  end
+
+  def duplicate!(user = nil)
+    new_form_field = dup
+    new_form_field.name = next_clone_name
+    new_form_field.created_by = user || created_by
+    new_form_field.position = form.form_fields.length
```
okriuchykhin, line 25: better use size method

app/models/form_field.rb
```diff
@@ -11,4 +13,20 @@ class FormField < ApplicationRecord
...
+    new_form_field.save!
+
+    new_form_field.insert_at(position + 1)
```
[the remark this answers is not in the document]
aignatov-bio: We have explicit position validation, and we first create record and after move to specific position. As i remember before we had problem with library with duplicated positions.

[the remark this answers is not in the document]
okriuchykhin: In such case make sure that locks are properly set

artoscinote (approved):

PR #8315, opened 2025-03-18
Add admin bypass [SCI-11308]

andrej-scinote:
Jira ticket: [SCI-11308](https://scinote.atlassian.net/browse/SCI-11308)

### What was done
Add admin bypass


app/controllers/users/passwords_controller.rb
```diff
@@ -39,11 +39,15 @@ def update
...
+    log_in_after_password_reset(resource) ? after_sign_in_path_for(resource) : new_session_path(resource_name)
   end
 
-  def two_factor_auth_enabled_for(user)
-    user.two_factor_auth_enabled?
+  def log_in_after_password_reset(user)
+    !user.two_factor_auth_enabled?
+  end
+
+  def check_database_authentication(_)
```
artoscinote, line 49: should be `check_database_authentication?` as it returns a boolean

artoscinote (changes requested):

PR #8409, opened 2025-04-08
Expand form repository row snapshots to cover relational values [SCI-11788]

artoscinote:
Jira ticket: [SCI-11788](https://scinote.atlassian.net/browse/SCI-11788)

### What was done
Expand form repository row snapshots to cover relational values, mock 'through' relation methods, since they can't work in-memory.


app/models/form_repository_rows_field_value.rb
```diff
@@ -62,12 +74,83 @@ def formatted
...
+
+      cell_value.asset = Asset.new(
+        cell_attributes['repository_asset_value']['asset'].except('file')
+      )
+
+      cell_value.asset.file = snapshot_file.blob
+
+      cell_value.asset.readonly!
+    when RepositoryListValue
+      cell_value.define_singleton_method(:repository_list_item) do
```
okriuchykhin, line 125: Why we can not directly assign a new item to the value?

artoscinote: Yes, here it's not necessary since it's a direct association, will fix.

PR #8443, opened 2025-04-18
Only display latest field values in form responses API [SCI-11843]

artoscinote:
Jira ticket: [SCI-11843](https://scinote.atlassian.net/browse/SCI-11843)

### What was done
Only display latest field values in form responses API


app/serializers/api/v2/form_response_serializer.rb
```diff
@@ -13,6 +13,10 @@ class FormResponseSerializer < ActiveModel::Serializer
 
       include TimestampableModel
 
+      def form_field_values
+        object.form_field_values.where(latest: true)
```
okriuchykhin, line 17: you can simply write object.form_field_values.latest
Rails automatically adds scopes for boolean columns

PR #8456, opened 2025-04-24
Add new sorting for projects table [SCI-11835]

aignatov-bio:
Jira ticket: [SCI-11835](https://scinote.atlassian.net/browse/SCI-11835)

### What was done
Add new sorting for projects table

[SCI-11835]: https://scinote.atlassian.net/browse/SCI-11835?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/project_folder.rb
+++ b/app/models/project_folder.rb
@@ -108,6 +108,8 @@ class ProjectFolder < ApplicationRecord
     false
   end
 
+  def description; end
+
   private
 
   def inherit_team_from_parent_folder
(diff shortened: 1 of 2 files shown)


artoscinote (commented):
Add `def description; end` to project_folder, so we can avoid additional ifs in the sorting methods.

artoscinote (changes requested):
Add def description; end to project_folder, so we can avoid additional ifs in the sorting methods.

PR #8510, opened 2025-05-16
Fix redirect for storage locations in activities [SCI-11917]

aignatov-bio:
Jira ticket: [SCI-11917](https://scinote.atlassian.net/browse/SCI-11917)

### What was done
Fix redirect for storage locations in activities


app/controllers/storage_locations_controller.rb
```diff
@@ -217,7 +217,7 @@ def load_storage_location
   end
 
   def check_read_permissions
-    render_403 unless can_read_storage_location?(@storage_location)
+    render_403 unless can_read_storage_location?(@storage_location) && @storage_location.container
```
okriuchykhin, line 220: why is it here?

app/models/storage_location_repository_row.rb
```diff
@@ -19,6 +19,8 @@ class StorageLocationRepositoryRow < ApplicationRecord
     validate :unique_repository_row
   end
 
+  validate :must_belong_to_container
```
okriuchykhin, line 22: if it is always required - just define a scoped belongs_to relation:

belongs_to :container_storage_location,
             -> { where(container: true) },
             class_name: 'StorageLocation',
             foreign_key: 'storage_location_id',
             inverse_of: :storage_location_repository_rows

app/controllers/storage_locations_controller.rb
artoscinote: I think instead, if it's not a container, we should redirect to the index with parent_id set with its id, probably just with a guard clause on top of the show action.

artoscinote (changes requested):

artoscinote (approved):

PR #8562, opened 2025-06-10
Optimize joins in RepositoryDatatableService for better performance on task table [SCI-11998]

okriuchykhin:
Jira ticket: [SCI-11998](https://scinote.atlassian.net/browse/SCI-11998)

### What was done
Optimize joins in RepositoryDatatableService for better performance on task table


app/services/repository_datatable_service.rb
```diff
@@ -50,34 +50,24 @@ def process_query
...
                             .select('SUM(DISTINCT my_module_repository_rows.stock_consumption) AS "consumed_stock"')
         end
       else
         repository_rows = repository_rows
-                          .joins(:repository)
-                          .joins('LEFT OUTER JOIN "my_module_repository_rows" "current_my_module_repository_rows"'\
-                                 'ON "current_my_module_repository_rows"."repository_row_id" = "repository_rows"."id" '\
-                                 'AND "current_my_module_repository_rows"."my_module_id" = ' + @my_module.id.to_s)
-                          .where('current_my_module_repository_rows.id IS NOT NULL '\
-                                 'OR (repository_rows.archived = FALSE AND repositories.archived = FALSE)')
```
artoscinote, line 76: removing this probably means it will show archived items on assign view

okriuchykhin: Archived filtering should happen on line 99

okriuchykhin: Looks like it doesn't, so we need to explicitly load only active rows here

PR #8650, opened 2025-07-08
Permission fixes [SCI-11953]

artoscinote:
Jira ticket: [SCI-11953](https://scinote.atlassian.net/browse/SCI-11953)

### What was done
Permission fixes

[SCI-11953]: https://scinote.atlassian.net/browse/SCI-11953?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/models/concerns/shareable.rb
+++ b/app/models/concerns/shareable.rb
@@ -27,7 +27,7 @@ module Shareable
     end
 
     scope :viewable_by_user, lambda { |user, teams = user.current_team|
-      readable_ids = readable_by_user(user).where(team: teams).pluck(:id)
+      readable_ids = with_granted_permissions(user, "#{permission_class.name}Permissions::READ".constantize, teams).pluck(:id)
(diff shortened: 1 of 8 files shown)


app/models/concerns/shareable.rb
```diff
@@ -27,7 +27,7 @@ module Shareable
     end
 
     scope :viewable_by_user, lambda { |user, teams = user.current_team|
-      readable_ids = readable_by_user(user).where(team: teams).pluck(:id)
+      readable_ids = with_granted_permissions(user, RepositoryPermissions::READ, teams).pluck(:id)
```
okriuchykhin: Probably it is better to use something like "#{name}Permissions::READ".constantize since it can be used in future for different types

PR #8659, opened 2025-07-14
PET UI fixes [SCI-12107]

aignatov-bio:
Jira ticket: [SCI-12107]](https://scinote.atlassian.net/browse/SCI-12107])

### What was done
PET UI fixes


app/javascript/vue/projects/renderers/status.vue
```diff
@@ -1,15 +1,23 @@
...
+      @change="changeStatus"
+      size="xs"
+      :borderless="true"
+      :optionRenderer="optionRenderer"
+      :labelRenderer="optionRenderer"
+    />
+  </div>
+  <div v-else class="flex items-center gap-2 py-0.5">
+    <div class="w-3 h-3 rounded-full"
+        :class="this.statusColor(params.data.status)"></div>
```
artoscinote: This should be done using conditional classes (and is more proper for tailwind, as it technically won't pick up the classes used in that method. Something like:

`:class="{ 'bg-sn-grey-500': params.data.status === 'not_started',  'bg-sn-science-blue': params.data.status === 'in_progress', 'bg-sn-alert-green': params.data.status === 'done }"`

PR #8858, opened 2025-09-09
Update experiment only on forward status change [SCI-12345]

aignatov-bio:
Jira ticket: [SCI-12345](https://scinote.atlassian.net/browse/SCI-12345)

### What was done
Update experiment only on forward status change


andrej-scinote (approved):

app/models/my_module.rb
```diff
@@ -507,6 +507,12 @@ def permission_parent
     experiment
   end
 
+  def status_changed_direction
```
okriuchykhin: there is not much sense to have this method here as it works only with changed and unsaved objects.
Move this check to the observer

okriuchykhin (approved):

PR #8870, opened 2025-09-12
Update tags modal in canvas [SCI-12287]

aignatov-bio:
Jira ticket: [SCI-12287](https://scinote.atlassian.net/browse/SCI-12287)

### What was done
Update tags modal in canvas 

[SCI-12287]: https://scinote.atlassian.net/browse/SCI-12287?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/serializers/my_module_serializer.rb
+++ b/app/serializers/my_module_serializer.rb
@@ -6,7 +6,7 @@ class MyModuleSerializer < ActiveModel::Serializer
   include ApplicationHelper
   include ActionView::Helpers::TextHelper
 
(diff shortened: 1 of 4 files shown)


app/serializers/my_module_serializer.rb
```diff
@@ -115,4 +115,15 @@ def tags
       { id: tag.id, name: tag.name, color: tag.color }
     end
   end
+
+  def tags_html
+    # legacy canvas support
+    return '' unless @instance_options[:controller]
```
artoscinote, line 121: we can just do `return unless`, no need to return empty string

PR #8909, opened 2025-09-30
Add result templates to protocols [SCI-12361]

aignatov-bio:
Jira ticket: [SCI-12361](https://scinote.atlassian.net/browse/SCI-12361)

### What was done
Add result templates to protocols


db/migrate/20250929104755_create_result_templates.rb
```diff
@@ -0,0 +1,20 @@
...
+      t.bigint :last_modified_by_id
+      t.integer :assets_view_mode
+      t.datetime :discarded_at
+
+      t.timestamps
+    end
+
+    add_column :result_orderable_elements, :result_template_id, :bigint
+    add_column :result_tables, :result_template_id, :bigint
+    add_column :result_texts, :result_template_id, :bigint
```
artoscinote: Here we should use `add_reference` instead of a bigint id

PR #9462, opened 2026-05-20
Add permissions for equipment bookings [SCI-13068]

aignatov-bio:
Jira ticket: [SCI-13068](https://scinote.atlassian.net/browse/SCI-13068)

### What was done
Add permissions for equipment bookings

[SCI-13068]: https://scinote.atlassian.net/browse/SCI-13068?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/controllers/calendar_events_controller.rb
+++ b/app/controllers/calendar_events_controller.rb
@@ -2,27 +2,21 @@
 
 class CalendarEventsController < ApplicationController
   before_action :check_calendar_events_enabled
+  before_action :load_parent, only: :index
+  before_action :load_subject, only: :create
   before_action :load_calendar_event, except: %i(index create)
-  before_action :load_repository_row, only: :create
-  before_action :check_read_permission, only: %i(index show)
-  before_action :check_manage_permission, except: %i(index show)
+  before_action :check_read_permission, only: :show
+  before_action :check_manage_permission, except: %i(index show create)
+  before_action :check_create_permission, only: :create
 
   def index
-    calendar_events = current_team.calendar_events.where(event_type: params[:event_type])
+    @calendar_events = current_team.calendar_events.where(event_type: params[:event_type])
 
-    if params.dig(:filters, :sub_types).present?
-      selected_sub_types = params[:filters][:sub_types].select { |_, v| v == 'true' }.keys
-      calendar_events = calendar_events.event_sub_type_filter(selected_sub_types)
-    end
-
-    calendar_events = calendar_events.datetime_filter(params[:start_date], params[:end_date]) if params[:start_date].present? && params[:end_date].present?
-    calendar_events = calendar_events.repository_filter(params[:repository_id]) if params[:repository_id].present?
-    calendar_events = calendar_events.repository_rows_filter(params[:filters][:subject_ids]) if params.dig(:filters, :subject_ids).present?
-    calendar_events = calendar_events.assigned_users_filter(params[:filters][:assigned_user_ids]) if params.dig(:filters, :assigned_user_ids).present?
+    filter_calendar_events!
 
     respond_to do |format|
       format.json do
-        render json: calendar_events,
+        render json: @calendar_events,
                each_serializer: CalendarEventSerializer
       end
     end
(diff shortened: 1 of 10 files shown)


app/permissions/repository.rb
```diff
@@ -140,6 +140,20 @@
       repository.can_manage_shared?(user) ||
       repository.permission_granted?(user, RepositoryPermissions::USERS_MANAGE)
   end
+
+  # Equipment booking: view/create/manage
+
+  can :read_equipment_bookings do |user, repository|
+    Repository.equipment_booking_enabled? && can_read_repository?(user, repository)
```
okriuchykhin: Move the first check out

app/controllers/repositories_controller.rb
```diff
@@ -497,6 +497,13 @@ def user_roles
     render json: { data: user_roles_collection(Repository.new).map(&:reverse) }
   end
 
+  def permissions
```
okriuchykhin, line 500: where are the rest of permissions?

app/controllers/calendar_events_controller.rb
```diff
@@ -98,6 +100,10 @@ def load_calendar_event
     render_404 unless @calendar_event
   end
 
+  def load_repository
```
okriuchykhin: If the repository always needed for permission check, move this check here

app/controllers/repositories_controller.rb
aignatov-bio: For now we need only this. It will be used when we will start repository refactor to VUE.

app/controllers/calendar_events_controller.rb
aignatov-bio: it only used in index action to load list of events, in other places it using directly calendar event as main object.
And it used only once in read permission check.

okriuchykhin: https://github.com/scinote-eln/scinote-web/pull/9462/changes#diff-ff225cf811457b7dcd5e3e0bdf2d5c6b7f01e7f7818a2b92bd788adebba0d9dbR142

https://github.com/scinote-eln/scinote-web/pull/9462/changes#diff-ff225cf811457b7dcd5e3e0bdf2d5c6b7f01e7f7818a2b92bd788adebba0d9dbR150

app/controllers/calendar_events_controller.rb
```diff
@@ -122,9 +128,30 @@ def calendar_event_params
...
+    if @repository_row
+      render_403 unless can_create_equipment_bookings?(@repository_row.repository)
+    else
+      render_403
+    end
+  end
+
+  def check_manage_permission
+    if @calendar_event && @calendar_event.subject_type == 'RepositoryRow'
+      render_403 unless can_manage_repository?(@calendar_event.subject.repository)
```
okriuchykhin: Shouldn't be here can_manage_equipment_bookings?

artoscinote (changes requested):
Change load_repository to load_parent

Ideally, from frontend we would send also subject_type and subject_id, then:
```
def load_parent
  @parent =
    case params[:parent_type]
    when 'Repository'
      @parent = Repository.readable_by_user.find(params[:parent_id]) # this will throw not found, so it already kind of handles read permission check
    else
      raise NotImplementedError
    end
end
```

In permission checks, use the same pattern, case-when by subject (no need to check @calendar_event && ..., it will render_404 anyway), example:
```
def check_read_permission
  case @calendar_event.subject_type
  when 'RepositoryRow'
    render_403 unless can_read_equipment_bookings?(@calendar_event.subject.repository)
  else
    raise NotImplementedError
  end
end
```
In index, change to
```
  def index
    @calendar_events = current_team.calendar_events.where(event_type: params[:event_type])

    filter_calendar_events!
    ...
```
Move the filtering to filter_calendar_events!, for now just re-arrange the filters to group repository-specific filters together.

artoscinote (approved):

PR #9606, opened 2026-07-30
Add websocket for assets preview [SCI-13326]

matej-scinote:
Jira ticket: [SCI-13326](https://scinote.atlassian.net/browse/SCI-13326)

### What was done
Add websocket for assets preview

[SCI-13326]: https://scinote.atlassian.net/browse/SCI-13326?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/javascript/vue/shared/content/attachments/mixins/preview_status.js
+++ b/app/javascript/vue/shared/content/attachments/mixins/preview_status.js
@@ -1,0 +1,80 @@
+import consumer from '../../../../../channels/consumer';
+
+const READY = 'ready';
+const PROCESSING = 'processing';
+const FAILED = 'failed';
+
+export default {
+  data() {
+    return {
+      previewStatus: this.attachment.attributes.preview_status,
+      previewLoaded: false,
+      previewSubscription: null
+    };
+  },
+  computed: {
+    previewProcessing() {
+      return this.previewStatus === PROCESSING;
+    },
+    previewFailed() {
+      return this.previewStatus === FAILED;
+    },
+    previewVisible() {
+      return this.previewLoaded;
+    },
+    previewLoading() {
+      return this.previewProcessing || (this.previewSrc !== null && !this.previewLoaded);
+    },
+    previewSrc() {
+      if (this.previewStatus !== READY) return null;
+
+      return this.attachment.attributes.medium_preview;
+    }
+  },
+  watch: {
+    'attachment.attributes.preview_status'(status) {
+      this.setPreviewStatus(status);
+    }
+  },
+  mounted() {
+    if (this.previewProcessing) this.subscribePreview();
+  },
+  beforeUnmount() {
+    this.unsubscribePreview();
+  },
+  methods: {
+    onPreviewLoad() {
+      this.previewLoaded = true;
+    },
+    onPreviewError() {
+      if (this.previewFailed) return;
+
+      this.setPreviewStatus(PROCESSING);
+    },
+    setPreviewStatus(status) {
+      this.previewStatus = status;
+
+      if (status !== PROCESSING) return;
+
+      this.previewLoaded = false;
+      this.subscribePreview();
+    },
+    subscribePreview() {
+      if (this.previewSubscription) return;
+
+      this.previewSubscription = consumer.subscriptions.create(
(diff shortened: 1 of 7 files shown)


app/jobs/active_storage/preview_job.rb
```diff
@@ -25,14 +27,21 @@ def perform(blob_id)
...
       preview = asset.large_preview.processed
-      Rails.logger.info "Preview for the Blod with id: #{blob.id} - successfully generated.\n" \
+      Rails.logger.info "Preview for the Blob with id: #{blob.id} - successfully generated.\n" \
                         "Transformations applied: #{preview.variation.transformations}"
 
+      # Cleanup successful preview generation on retry
+      if blob.metadata['preview_failed']
+        blob.metadata.delete('preview_failed')
+        blob.save!
+      end
```
artoscinote, line 41: for flags such as this one, we want to avoid going through callbacks, so use _update_column_ instead.

app/javascript/vue/shared/content/attachments/mixins/preview_status.js
```diff
@@ -0,0 +1,79 @@
...
+      return this.previewState === 'failed';
+    },
+    previewVisible() {
+      return this.previewState === 'ready';
+    },
+    previewSrc() {
+      const src = this.attachment.attributes.medium_preview;
+      if (!src || this.previewReloadToken === 0) return src;
+
+      return `${src}${src.includes('?') ? '&' : '?'}preview_reload=${this.previewReloadToken}`;
```
artoscinote: Could we just subscribe to the channel if preview doesn't exist and is not in a failed state? In the component, we then display the loader until src is set, so we don't have to do the GET parameter workaround for reloading the src.

Maybe add preview status to asset serializer (and a preview_status method to the model) to assist with this.

app/controllers/active_storage/representations/redirect_controller.rb
```diff
@@ -24,6 +24,8 @@ def show
           @blob.save!
         end
 
+        AssetPreviewChannel.broadcast_to(@blob.attachments.take.record, status: 'failed') if @blob.attachments.take.record_type == 'Asset'
```
okriuchykhin: I don't think it is needed here as this method directly returns response to the client and is used for inline preview generation(without a delayed job)

app/jobs/active_storage/preview_job.rb
```diff
@@ -25,14 +27,21 @@ def perform(blob_id)
...
+
       asset.update(file_processing: false)
+      AssetPreviewChannel.broadcast_to(asset, status: 'ready') if asset.is_a?(Asset)
```
okriuchykhin: asset is always Asset in this method

app/jobs/active_storage/preview_job.rb
```diff
@@ -8,11 +8,13 @@ class ActiveStorage::PreviewJob < ActiveStorage::BaseJob
...
   discard_on StandardError do |job, error|
     blob = ActiveStorage::Blob.find_by(id: job.arguments.first)
+    asset = blob&.attachments&.take&.record
     ActiveRecord::Base.no_touching do
-      blob&.attachments&.take&.record&.update!(file_processing: false)
+      asset&.update!(file_processing: false)
       blob.metadata['preview_failed'] = true
       blob.save!
     end
+    AssetPreviewChannel.broadcast_to(asset, status: 'failed') if asset.is_a?(Asset)
```
okriuchykhin: asset is always Asset in this method

app/jobs/active_storage/preview_job.rb
matej-scinote: Fixed it and update on other places in this file as well

app/controllers/active_storage/representations/redirect_controller.rb
matej-scinote: Removed it.

app/javascript/vue/shared/content/attachments/mixins/preview_status.js
matej-scinote: added  preview_status and check on fail

artoscinote (approved):

PR #9619, opened 2026-08-05
Fix storage permissions in docker [ST-1249]

matej-scinote:
Jira ticket: [ST-1249](https://scinote.atlassian.net/browse/ST-1249)

### What was done
Fix storage permissions in docker

[ST-1249]: https://scinote.atlassian.net/browse/ST-1249?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,9 +55,9 @@ ENV BUNDLE_PATH=/usr/local/bundle/
 ENV APP_HOME=/usr/src/app
 ENV PATH=$APP_HOME/bin:$PATH
 
-RUN mkdir $APP_HOME
+RUN mkdir $APP_HOME $APP_HOME/storage
 RUN useradd --uid 1000 -m scinote
-RUN chown scinote:scinote $APP_HOME
+RUN chown scinote:scinote $APP_HOME $APP_HOME/storage
 USER scinote
 ENV CHROMIUM_PATH=$APP_HOME/bin/chromium
 WORKDIR $APP_HOME


Dockerfile
```diff
@@ -58,6 +58,7 @@ ENV PATH=$APP_HOME/bin:$PATH
 RUN mkdir $APP_HOME
```
okriuchykhin: RUN mkdir $APP_HOME $APP_HOME/storage

Dockerfile
```diff
@@ -58,6 +58,7 @@ ENV PATH=$APP_HOME/bin:$PATH
...
 RUN useradd --uid 1000 -m scinote
 RUN chown scinote:scinote $APP_HOME
```
okriuchykhin: RUN chown scinote:scinote $APP_HOME $APP_HOME/storage

Dockerfile
```diff
@@ -58,6 +58,7 @@ ENV PATH=$APP_HOME/bin:$PATH
...
+RUN mkdir -p $APP_HOME/storage && chown -R scinote:scinote $APP_HOME/storage
```
okriuchykhin: why do use -p and -R flags here?

Dockerfile.production
```diff
@@ -127,6 +127,7 @@ ENV NODE_PATH=/usr/local/lib/node_modules
 ENV BUNDLE_APP_CONFIG=.bundle
 
 RUN useradd --uid 1000 -m scinote
+RUN mkdir -p $APP_HOME/storage && chown -R scinote:scinote $APP_HOME/storage
```
okriuchykhin: put it after COPY(line 132)

okriuchykhin: and probably you will not need to chown

okriuchykhin (changes requested):
Check the comments

Dockerfile.production
matej-scinote: I will try.

Dockerfile
matej-scinote: The problem is if you build the image locally and you previously already run docker-compose the storage dir is created in your repo, if you then try to build image it will fail because it will copy the storage dir you already have in your repo.

matej-scinote: Recursive flag is unnecessary

okriuchykhin (approved):

PR #9710, opened 2026-09-08
Refactor orderable elements for step [SCI-13494]

matej-scinote:
Jira ticket: [SCI-xxxx](https://scinote.atlassian.net/browse/SCI-xxxx)

### What was done
Refactor orderable elements for step


app/controllers/result_elements/base_controller.rb
```diff
@@ -114,7 +114,12 @@ def create_in_result!(result, new_orderable)
     end
 
     def render_result_orderable_element(orderable)
```
okriuchykhin: I think this method can be removed as there is no much point in having it

app/controllers/step_elements/base_controller.rb
```diff
@@ -109,7 +109,14 @@ def create_in_step!(step, new_orderable)
     end
 
     def render_step_orderable_element(orderable)
```
okriuchykhin: I think this method can be removed as there is no much point in having it

app/serializers/checklist_serializer.rb
```diff
@@ -6,10 +6,19 @@ class ChecklistSerializer < ActiveModel::Serializer
...
+  attributes :id, :name, :urls, :icon, :sa_name, :parent_type, :archived, :archived_by, :archived_on, :locked,
+             :orderable_element, :orderable_type
 
   has_many :checklist_items, serializer: ChecklistItemSerializer
 
+  def orderable_element
+    object.step_orderable_element
+  end
+
+  def orderable_type
```
okriuchykhin: type is included by default in ActiveModelSerializers

app/serializers/checklist_serializer.rb
```diff
@@ -6,10 +6,19 @@ class ChecklistSerializer < ActiveModel::Serializer
   include ApplicationHelper
   include ActionView::Helpers::TextHelper
 
-  attributes :id, :name, :urls, :icon, :sa_name, :parent_type, :archived, :archived_by, :archived_on, :locked
+  attributes :id, :name, :urls, :icon, :sa_name, :parent_type, :archived, :archived_by, :archived_on, :locked,
+             :orderable_element, :orderable_type
 
   has_many :checklist_items, serializer: ChecklistItemSerializer
 
+  def orderable_element
```
okriuchykhin: I think it is better to expose position here directly

app/serializers/checklist_serializer.rb
matej-scinote: I can't find it

okriuchykhin (approved):

